PR #153, opened 2021-01-11
Do not panic when UDP dispatcher fails to write response

maskedeken:
fix [https://github.com/v2fly/v2ray-core/issues/495](https://github.com/v2fly/v2ray-core/issues/495)


rprx:
之前我看到这里也觉得奇怪，感谢 PR

PR #219, opened 2021-01-30
feat: add pprof outbound

yichya:
Make it easier to collect useful information beyond logs.

Usage: quite similar to how to enable Xray APIs

1. Add a top-level configuration called `pprof` with `tag` (outbound tag) as its only configuration
  ```json
  {
      "pprof": {
          "tag": "pprof"
      }
  }
  ```
2. Add a dokodemo door inbound. For example here we use 127.0.0.1:11111
  ```json
  {
      "port": 11111,
      "protocol": "dokodemo-door",
      "tag": "pprof",
      "settings": {
          "address": "127.0.0.1"
      },
      "listen": "127.0.0.1"
  }
  ```
3. Use routing to connect them together
  ```json
  {
      "type": "field",
      "inboundTag": [
          "pprof"
      ],
      "outboundTag": "pprof"
  }
  ```
4. That port will work just like a normal built-in pprof http server. Access http://127.0.0.1:11111/debug/pprof/ or use `go tool pprof` to start profiling or inspect running goroutines.


yichya:
will be included in #365

PR #361, opened 2021-03-10
fix: fix the TrojanServerConfig Fallback issue

### LINKED ISSUE #360: Bug about TrojanServerConfig
The TrojanServerConfig https://github.com/XTLS/Xray-core/blob/924fe1607725f313e4eb053a25291e8202ac7037/infra/conf/trojan.go#L107  in /Xray-core/infra/conf/trojan.go use the `json.RawMessage` as the `FallBack` type.  However, this type leads the filed to become `"null"` after the `json.Unmarshal`, which causes the error in https://github.com/XTLS/Xray-core/blob/924fe1607725f313e4eb053a25291e8202ac7037/infra/conf/trojan.go#L136.

I think we should change the type of `Fallback` to `*TrojanInboundFallback` or change the comparison in https://github.com/XTLS/Xray-core/blob/924fe1607725f313e4eb053a25291e8202ac7037/infra/conf/trojan.go#L136
to `if c.Fallback != "null"`

```
// TrojanServerConfig is Inbound configuration
type TrojanServerConfig struct {
	Clients   []*TrojanUserConfig      `json:"clients"`
	Fallback  json.RawMessage          `json:"fallback"`RPRX, 3 months ago: • v1.0.0
	Fallbacks []*TrojanInboundFallback `json:"fallbacks"`
}
```

```
	if c.Fallback != nil {
		return nil, newError(`Trojan settings: please use "fallbacks":[{}] instead of "fallback":{}`)
	}
```

RManLuo:
Fix #360

### DIFF
--- a/infra/conf/trojan.go
+++ b/infra/conf/trojan.go
@@ -104,7 +104,7 @@ type TrojanUserConfig struct {
 // TrojanServerConfig is Inbound configuration
 type TrojanServerConfig struct {
 	Clients   []*TrojanUserConfig      `json:"clients"`
-	Fallback  json.RawMessage          `json:"fallback"`
+	Fallback  *TrojanInboundFallback   `json:"fallback"`
 	Fallbacks []*TrojanInboundFallback `json:"fallbacks"`
 }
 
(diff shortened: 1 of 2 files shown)


rprx:
感谢 PR

PR #436, opened 2021-03-24
feat: shouldSniffDomain for (FakeDNS) Sniffer

rurirei:
to ensure uses FakeDNSSniffer only (not do pendingSniffers to do domain sniff unnecessarily) when if it succeed.

### DIFF
--- a/common/protocol/http/sniff.go
+++ b/common/protocol/http/sniff.go
@@ -56,7 +56,7 @@ func beginWithHTTPMethod(b []byte) error {
 	return errNotHTTPMethod
 }
 
-func SniffHTTP(b []byte, shouldSniffDomain bool) (*SniffHeader, error) {
+func SniffProtocolHTTP(b []byte) (*SniffHeader, error) {
 	if err := beginWithHTTPMethod(b); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


rprx:
麻烦 @yuhan6665 看一下

rurirei:
@.xiaokangwang may designed v2fly/v2ray-core#401 to quickly return result of FakeDNSSniffer only when metadataOnly true.

this is an opinion to ensure not do domain sniffing twice when metadataOnly false, as for v2fly/v2ray-core#697 @.yuhan6665

common/protocol/http/sniff.go
```diff
@@ -65,6 +65,10 @@ func SniffHTTP(b []byte) (*SniffHeader, error) {
 		version: HTTP1,
 	}
 
+	if !shouldSniffDomain {
```
yuhan6665, line 68: I understand the intention here. However, at this point, the request content is already read in memory. It doesn't seem very expensive to, parse a http header, for example.

rurirei: this do skip the process of parsing host.

while in current process, domain sniffing of FakeDNS makes nothing, it's simply duplicated.

rurirei:
i will do a refactor commit. here keeps for conversation.

PR #476, opened 2021-04-04
fix grpc dial ipv6 address

lucifer9:
Now dial using the proper IPv6 address (without []).

### DIFF
--- a/transport/internet/grpc/dial.go
+++ b/transport/internet/grpc/dial.go
@@ -90,8 +90,14 @@ func getGrpcClient(ctx context.Context, dest net.Destination, tlsConfig *tls.Con
 		dialOption = grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig.GetTLSConfig()))
 	}
 
+	var grpcDestHost string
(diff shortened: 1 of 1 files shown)


transport/internet/grpc/dial.go
```diff
@@ -91,7 +91,7 @@ func getGrpcClient(ctx context.Context, dest net.Destination, tlsConfig *tls.Con
 	}
 
 	conn, err := grpc.Dial(
-		gonet.JoinHostPort(dest.Address.String(), dest.Port.String()),
+		gonet.JoinHostPort(dest.Address.IP().String(), dest.Port.String()),
```
JimhHan, line 94: dest.Address 有可能为 `domainAddress`

JimhHan (changes requested):

transport/internet/grpc/dial.go
lucifer9: You are right.

AkinoKaede (approved):

JimhHan (approved):

AkinoKaede:
Thanks for your work

Issue #554, opened 2021-05-10
使用 XRay 的 Unix domain socket 功能，生成的文件似乎有权限问题。


mclovin-2k:
`    "inbounds": [
        {
            "listen": "/dev/shm/xray_uds_vless_grpc.socket",
            "protocol": "vless",
            ..........
        }
     ]
`

在 inbound 的 listen 属性中，启用 Unix domain socket 之后。生成的对应文件如图。

![QQ截图20210510101846](https://user-images.githubusercontent.com/5440567/117598050-7eb6cb00-b179-11eb-99b9-478136ef68c2.png)

上面4个文件，是 NGINX 的 Unix domain socket 文件，可以看到其权限为所有人可读可写。
后面2个文件是 XRAY 的 Unix domain socket 文件，其权限为所有者可读可写。
这样会导致 NGINX 无法通过 /dev/shm/xray_uds_vless_grpc.socket 和 XRAY 通信。如图：
![QQ截图20210510135015](https://user-images.githubusercontent.com/5440567/117611790-b08a5a80-b196-11eb-90b4-962f712ebf9b.png)

使用 chmod 将 /dev/shm/xray_uds_vless_grpc.socket 权限改为所有人可读可写后，错误消失，通信正常。

rurirei:
you should NOT use root user.

changyp6:
```
 transport/internet/system_listener.go | 8 ++++++++
 1 file changed, 8 insertions(+)

diff --git a/transport/internet/system_listener.go b/transport/internet/system_listener.go
index eb753d6..809f338 100644
--- a/transport/internet/system_listener.go
+++ b/transport/internet/system_listener.go
@@ -4,6 +4,7 @@ import (
        "context"
        "runtime"
        "syscall"
+       "os"
 
        "github.com/pires/go-proxyproto"
        "github.com/xtls/xray-core/common/net"
@@ -45,6 +46,7 @@ func (dl *DefaultListener) Listen(ctx context.Context, addr net.Addr, sockopt *S
        var l net.Listener
        var err error
        var network, address string
+       var ds bool
        switch addr := addr.(type) {
        case *net.TCPAddr:
                network = addr.Network()
@@ -72,6 +74,7 @@ func (dl *DefaultListener) Listen(ctx context.Context, addr net.Addr, sockopt *S
                                return nil, err
                        }
                        ctx = context.WithValue(ctx, address, locker)
+                       ds = true
                }
        }
 
[...]

PR #556, opened 2021-05-11
change socket‘s permission to 0777

lelemka0:
#554 #342

### DIFF
--- a/transport/internet/system_listener.go
+++ b/transport/internet/system_listener.go
@@ -84,8 +84,8 @@ func (dl *DefaultListener) Listen(ctx context.Context, addr net.Addr, sockopt *S
 		l = &proxyproto.Listener{Listener: l, Policy: policyFunc}
 	}
 	if err == nil && ds {
-		if e := os.Chmod(address, 0777); e != nil {
-			newError("failed to set file mode of ", address, "to 0777").Base(e).AtWarning().WriteToLog()
+		if e := os.Chmod(address, 0666); e != nil {
+			newError("failed to set file mode of ", address, "to 0666").Base(e).AtWarning().WriteToLog()
 		}
 	}
 	return l, err
--- a/transport/internet/system_listener.go
+++ b/transport/internet/system_listener.go
@@ -46,7 +46,6 @@ func (dl *DefaultListener) Listen(ctx context.Context, addr net.Addr, sockopt *S
 	var l net.Listener
 	var err error
 	var network, address string
-	var ds bool
 	switch addr := addr.(type) {
 	case *net.TCPAddr:
 		network = addr.Network()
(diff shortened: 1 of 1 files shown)


AkinoKaede:
I don't think it is a good method.

lelemka0:
> I don't think it is a good method.

https://github.com/golang/go/issues/11822#issuecomment-123850227
This seems to be the one of recommended methods.
Another method (syscall.Umask) will affect all files.

ljxfstorm:
As from nginx/src/core/ngx_connection.c:
```
            if (ls[i].sockaddr->sa_family == AF_UNIX) {
                mode_t   mode;
                u_char  *name;

                name = ls[i].addr_text.data + sizeof("unix:") - 1;
                mode = (S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH|S_IWOTH);

                if (chmod((char *) name, mode) == -1) {
                    ngx_log_error(NGX_LOG_EMERG, cycle->log, ngx_errno,
                                  "chmod() \"%s\" failed", name);
                }

                if (ngx_test_config) {
                    if (ngx_delete_file(name) == NGX_FILE_ERROR) {
                        ngx_log_error(NGX_LOG_EMERG, cycle->log, ngx_errno,
                                      ngx_delete_file_n " %s failed", name);
                    }
                }
            }
```
Mode 0666 may be a better choice.

transport/internet/system_listener.go
```diff
@@ -80,6 +83,11 @@ func (dl *DefaultListener) Listen(ctx context.Context, addr net.Addr, sockopt *S
 		policyFunc := func(upstream net.Addr) (proxyproto.Policy, error) { return proxyproto.REQUIRE, nil }
 		l = &proxyproto.Listener{Listener: l, Policy: policyFunc}
 	}
+	if err == nil && ds {
+		if e := os.Chmod(address, 0666); e != nil {
```
rurirei, line 87: move this to the `ds body (if body)` above. there's no need to check nil error and create a temp bool.

lelemka0: ok

rurirei: do not create unnecessary bool `ds = true`, and handle err in the domain_socket body.
```golang
... ...
		} else {
			// normal unix domain socket needs lock
			locker := &FileLocker{
				path: address + ".lock",
			}
			if err := locker.Acquire(); err != nil {
				return nil, err
			}
			if err := os.Chmod(address, 0666); err != nil {
				return nil, err
			}
			ctx = context.WithValue(ctx, address, locker)
		}
```

hmol233:
感谢各位的支持与贡献。
正如 @AkinoKaede 所说，这并不是一个好方法。
稍后我会另外实现一个。

PR #609, opened 2021-06-18
Move `DomainStrategy` & `DialerProxy` to `DialSystem`

hmol233:
fix https://github.com/XTLS/Xray-core/issues/608


badO1a5A90:
感谢PR

PR #669, opened 2021-08-07
Fix cannot disable ReadV

AkinoKaede:


common/buf/readv_reader.go
```diff
@@ -134,17 +133,13 @@ func (r *ReadVReader) ReadMultiBuffer() (MultiBuffer, error) {
 	return mb, nil
 }
 
-var useReadv = true
+var useReadv bool
```
rurirei, line 136: simply `var useReadv = false` could fix the Switch behavior in init().

seems Readv is enabled if there is a applicable environment, you should require @RPRX for a Disable statement.

AkinoKaede: According the release notes of v1.0.0, I think ReadV should be enabled in all environment.

hmol233 (approved):

badO1a5A90:
感谢PR

PR #679, opened 2021-08-20
Update x/sys to support go 1.17

AlekSi:
See: https://github.com/golang/go/issues/46763
More context: https://github.com/Homebrew/homebrew-core/pull/83413


hmol233:
Thanks for your contribution. We have release a new update 🎉

PR #723, opened 2021-09-20
Fix flaky TestVMessDynamicPort

yuhan6665:
I will try fixing some flaky tests, start with https://github.com/XTLS/Xray-core/runs/3605024544?check_suite_focus=true
This test can fail randomly for two reasons:

1. The server core and client core could try to listen on the same port, cause one core fail
2. The server listen for three ports, some might not be available

To fix first problem, I change it so that server will run first. I open another dokodemo inbound and test it is available before creating the client core.
For 2, I added a loop for server core to try pickPort() for 5 times if it failed.
I also fix the tcpConnAvailableAtPort() method and it is running much faster.

### DIFF
--- a/testing/scenarios/common.go
+++ b/testing/scenarios/common.go
@@ -142,6 +142,23 @@ func CloseAllServers(servers []*exec.Cmd) {
 	})
 }
 
+func CloseServer(server *exec.Cmd) {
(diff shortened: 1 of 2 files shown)


teddysun:
looks good to me

badO1a5A90:
非常感谢!

PR #725, opened 2021-09-21
Fix go generate core/format.go

yuhan6665:
Found an issue in format.go. It currently only work with `go run infra/vformat/main.go`
Fix by porting logic from vprotogen.


badO1a5A90:
非常感谢 PR!

PR #733, opened 2021-09-25
Fakedns sniff and override destination for buffer.UDP

yuhan6665:
Re-push https://github.com/XTLS/Xray-core/pull/610
Summary about bug https://github.com/XTLS/Xray-core/issues/563:
This is a unique issue in Xray core when XUDP was added. However, it seems not reproducible in some environment. For example I cannot reproduce any issue when testing with v2rayNG.
I attempted a fix in this PR which seems fix the issue but someone mentioned it may break full cone. Then @hmol233 pushed another fix https://github.com/XTLS/Xray-core/commit/231b157902d3cd8be30f133083b7297749eb48dc
Please take a look @badO1a5A90

### DIFF
--- a/app/proxyman/inbound/worker.go
+++ b/app/proxyman/inbound/worker.go
@@ -6,6 +6,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/xtls/xray-core/app/dispatcher"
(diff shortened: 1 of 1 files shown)


yuhan6665:
New fix in https://github.com/XTLS/Xray-core/pull/1011

PR #739, opened 2021-09-26
Fix some tests to use udp.PickPort()

yuhan6665:
https://github.com/XTLS/Xray-core/pull/731/checks?check_run_id=3713499282
Some tests try to listen udp port, but the port maybe not available.


badO1a5A90:
非常感谢 PR!

PR #744, opened 2021-09-28
Move from deprecated ioutil to os and io packages

kallydev:
XTLS is currently based on Go 1.17, so it is safe to remove deprecated ioutil package, ref v2fly/v2ray-core#1298 and https://golang.org/doc/go1.16#ioutil.


badO1a5A90:
thanks for your contribution!

PR #855, opened 2021-12-15
go 1.17.5

tritu:


kallydev:
How about we just set the Go version to `1.17`?

https://groups.google.com/g/golang-announce/c/hcmEScgc00k

tritu:
While

> How about we just set the Go version to `^1.17`?
> 
> https://groups.google.com/g/golang-announce/c/hcmEScgc00k

Well, we can.
But current build yml is based on https://github.com/actions/setup-go and caching version https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#go is not always up to date normally.

yuhan6665:
Thanks for your contribution!

PR #859, opened 2021-12-16
fix unsupported cipher method: xChaCha20-IETF-Poly1305

xqzr:


yuhan6665:
是我们疏忽了 感谢 pr！

嗯 看来我们需要 1.5.2。。

dyhkwong:
What are you doing here... The NonceSize of `xchacha20-ietf-poly1305` is not 12 but you are using `crypto.GenerateInitialAEADNonce()` to generate a fixed size of AEAD nonce and you are calling `aead.Open`with a 12-byte length nonce size, which will cause panic in shadowsocks validator. This is obviously untested at all.

yuhan6665:
@dyhkwong 已复现 感谢大佬测试 我尝试一下看能不能修

PR #915, opened 2022-01-28
Quic related improvements

yuhan6665:


Hubedge:
Quick questions regarding 79a05d5:
- Can the "wearing-out" effect be reproduced over a safer/stabler network, say, LAN?
- Is it a common practice for HTTP3 implementations/applications to frequently opening new and closing old**est** sessions?
- Finally and most importantly, is this a pattern strong enough to be recognised?

yuhan6665:
@Hubedge 
I don't know what is the root cause of single session wearing-out with Quic in *ray. I intended to do more tests when I got some time. Currently our project is short of dev hand ;)
Note that when to open, when to close the session, fundamentally decided by the traffic pattern of the app and the user. I believe it is more difficult to do these kind of analysis than well known detection methods like individual packet analyze, active probing etc.
As I mentioned, I think it is more valuable to think about Http3 分流和回落 at this time.

PR #918, opened 2022-01-30
Update README.md

Buttires1984:


README.md
```diff
@@ -43,16 +43,19 @@
...
-  - [Qv2ray](https://github.com/Qv2ray/Qv2ray)  (This project had been archived and currently inactive)
-  - [Netch (NetFilter & TUN/TAP)](https://github.com/NetchX/Netch)
 - Android
   - [v2rayNG](https://github.com/2dust/v2rayNG)
   - [AnXray](https://github.com/XTLS/AnXray)
   - [Kitsunebi](https://github.com/rurirei/Kitsunebi/tree/release_xtls)
 - iOS & macOS (with M1 chip)
   - [Shadowrocket](https://apps.apple.com/app/shadowrocket/id932747118)
+  - 
+## GUI Clients that are inactively maintained
```
yuhan6665, line 53: hmm I don't think we need a separate sector for "Clients that are inactively maintained". Have a note after it is good enough.
If some day the software become too obsolete to use, I think we can just remove it.

yuhan6665:
先关了

PR #957, opened 2022-02-21
workflows: add check-latest

tritu:
check-latest can make sure actions will always use the latest minor version of golang to build.
No manually update needed anymore.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,6 +126,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ^1.17.7
+          check-latest: true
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -126,6 +126,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ^1.17.7
```
ebye9, line 128: go-version应该用 `go-version: '1.17'`

yuhan6665: 感谢大佬提醒

tritu: new PR opened

PR #987, opened 2022-03-13
[WIP] Testing with race condition detection

xtls1:


yuhan6665:
看起来这个测试不兼容
```
--- FAIL: TestPeriodicTaskStop (12.31s)
    testing.go:1152: race detected during execution of test
```

xtls1:
@yuhan6665 thanks you for your reply!

The race detection works, but there are several race conditions in the testing that cause the test to fail, and these failed runs have nothing to do with pr, so perhaps we should fix it again after merging this pr?

yuhan6665:
Hmm how many tests need fixing? Is it possible to fix them first? Or our CI will be red for a long time ;)

xtls1:
> Hmm how many tests need fixing? Is it possible to fix them first? Or our CI will be red for a long time ;)

I searched for this [runs](https://github.com/XTLS/Xray-core/runs/5525295334?check_suite_focus=true) using the keyword `race detected`, and it it seems that only `TestPeriodicTaskStop` caused the race condition.

I'll try to fix it as soon as possible.

yuhan6665:
Hmm there might be many other tests that are not compatible with this mode. Maybe you can try run locally to see how many tests may fail

yuhan6665:
先关了 *ray 里面的测试比较强大且复杂 但是有很多长期存在的问题 我有一些修测试的经验 我的建议是从小地方开始 逐步改进
欢迎你继续参与项目！

PR #1007, opened 2022-04-01
fix: expvar error when observatory is not enabled & various typos

yichya:
This commit fixes an error in metrics handler when observatory is not enabled.

At the same time changes some typos found in stats.


yuhan6665:
感谢大佬！ 我之前测试的时候没开 observatory 确实有这个小问题

PR #1040, opened 2022-04-29
fix proto file name conflict by adding package name

pocketW:


yuhan6665:
Thanks for your pr!
BTW @yichya this is the way you can generate proto file with path:
`go generate core/proto.go`

PR #1222, opened 2022-09-17
up yaml.v2 to yaml.v3

luoluodaduan:


yuhan6665:
thanks for your pr! I see github.com/invopop/yaml is still quite new. I would personally wait a bit to see if original repo is really stopped maintaining. Do you see a feature that is important in new repo?

PR #1252, opened 2022-10-13
Added utls to http2 transport

HirbodBehnam:
Hello
I was messing around with xray settings and realized that [utls](https://github.com/refraction-networking/utls) integration only works with raw tcp + tls configuration. So I thought maybe there is a way to bring utls to websocket/gRPC and http2!
Turns out, websocket and gRPC are not really handy to use utls in them. But http is really easy because we simply set DialTLS function and do it ourselves. So I added the ability to choose between `gotls.Client` and `utls.Client` and stablish the TLS with them.

To achieve a clean code, I added an interface to tls package which provides methods needed to initiate the tls connection. Unfortunately, `ConnectionState` method returns two different structs in gotls and utls so I had to create a method called `NegotiatedProtocol` to return the `NegotiatedProtocol` and `NegotiatedProtocolIsMutual`.

At last, I tested it with my own server. I set up my server just like [this](https://github.com/XTLS/Xray-examples/tree/main/VLESS-H2C-Caddy2) example and it is working fine! Next I checked the fingerprints which they also lead to fingerprints used by Firefox and Chrome. I used wireshark to capture the packets and test them.
[Firefox Fingerprint](https://tlsfingerprint.io/id/b250617dca0ac79f)
[Chrome Fingerprint](https://tlsfingerprint.io/id/e47eae8f8c4887b6)

### DIFF
--- a/transport/internet/http/dialer.go
+++ b/transport/internet/http/dialer.go
@@ -39,8 +39,8 @@ func getHTTPClient(ctx context.Context, dest net.Destination, streamSettings *in
 	}
 
 	httpSettings := streamSettings.ProtocolSettings.(*Config)
-	tlsConfig := tls.ConfigFromStreamSettings(streamSettings)
-	if tlsConfig == nil {
+	tlsConfigs := tls.ConfigFromStreamSettings(streamSettings)
+	if tlsConfigs == nil {
 		return nil, newError("TLS must be enabled for http transport.").AtWarning()
 	}
 	sockopt := streamSettings.SocketSettings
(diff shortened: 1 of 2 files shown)


SakuraSakuraSakuraChan:
@HirbodBehnam Thanks, V2ray_simple (a.k.a. Verysimple) support gRPC with uTLS

yuhan6665:
thanks for your contribution!
I heard in recent talks that authoritarian regimes are sharing censorship technology. United, we can stand against them!

PR #1264, opened 2022-10-20
Added uTLS to gRPC

HirbodBehnam:
Hello again
At the very end, I added uTLS to gRPC transport.

What I did to implement this, was that I copy and pasted [this lines of code](https://github.com/grpc/grpc-go/blob/v1.50.1/credentials/tls.go#L33-L146) from gRPC's library into `tls` package of xray. Then, I changed the go's TLS package with uTLS. I also made the `ServerHandshake` method to always panic because it's never called (because this is client's code!) and, well, we don't want uTLS to act as server. I can change this and simply copy and paste the `ServerHandshake` [code](https://github.com/grpc/grpc-go/blob/v1.50.1/credentials/tls.go#L113-L130) from gRPC's library to xray's code; But as I said, I think it's better to keep it this way.
One small change which I also done is this line:
https://github.com/XTLS/Xray-core/pull/1264/files#diff-79f0d50c97be933687e099e60e722cbb9376197d0edde3ba554d23f526a112ebR30
In gRPC's library it uses a map between cipher suites and their string name to get this value. But in this pull request I only use the base 10 representation of the cipher suite which is sent in binary in client hello.
I also re-defined `TLSInfo` as `grpcUtlsInfo` but changed the `State` to be `utls.ConnectionState` instead of `tls.ConnectionState`.

Just like other pull requests I made, if fingerprint is not defined it simply falls back to standard gRPC tls. This ensures maximum backward compatibility.
Also, I tested this pull request with my own server which is behind Cloudflare and Caddy and I can tell you it's working fine!
As well as, I tested the fingerprints with https://tlsfingerprint.io/pcap and wireshark.

### DIFF
--- a/transport/internet/grpc/dial.go
+++ b/transport/internet/grpc/dial.go
@@ -121,7 +121,13 @@ func getGrpcClient(ctx context.Context, dest net.Destination, streamSettings *in
 	}
 
 	if tlsConfig != nil {
-		dialOptions = append(dialOptions, grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig.GetTLSConfig())))
+		var transportCredential credentials.TransportCredentials
+		if fingerprint, exists := tls.Fingerprints[tlsConfig.Fingerprint]; exists {
+			transportCredential = tls.NewGrpcUtls(tlsConfig.GetTLSConfig(), fingerprint)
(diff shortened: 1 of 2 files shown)


yuhan6665:
Many thanks for your work!

PR #1317, opened 2022-11-08
add restrictedIPsPath flag for limit restricted IPs

hossinasaadi:
add `-restrictedIPsPath blockedIPs.txt` OR `-rip blockedIPs.txt`  to Xray for drop connection from limited IPs.
stored IPs should be split by ","  in the text file. 
**EX blockedIPs.txt content** : 
`1.2.3.4,1.1.1.2,5.6.7.8`
it's useful for limiting connected Different IPs to inbound at the same time.
like implemented for x-ui [here](https://github.com/hossinasaadi/x-ui)

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -280,6 +280,9 @@ func initRestrictedIPs(){
 	ticker := time.NewTicker(intvalSecond)
 	quit := make(chan struct{})
 	restrictedIPsPath := getOsArgValue(os.Args,"-restrictedIPsPath","-rip")
+	if restrictedIPsPath == "" {
+		return
+	}
 	go func() {
 		intvalSecond = 30 * time.Second
 		for {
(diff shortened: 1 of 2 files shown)


yuhan6665:
感谢 pr。能否通过路由配置 source 匹配源IP到黑洞达到相同的效果？
https://xtls.github.io/config/routing.html#ruleobject
IP还可以写进文件里面 像 geoip 一样

hossinasaadi:
> Can the source match the source IP to the black hole by routing configuration to achieve the same effect

Hi @yuhan6665 , it's about online devices with different IPs on one Inbound Like Vmess.
when you limit an inbound to 1 IP (in your way at the panel or something), you can get User IPs from the access log and save extra IPs in the passed file to Xray core (blockedIPs.txt)  and it will drop them.
in your way each time you should restart Xray

yuhan6665:
I just remember there is a kinda forgotten draft for this feature. You can take a look https://github.com/XTLS/Xray-core/pull/412
If you can reinstate it we can probably merge. Thanks!

PR #1362, opened 2022-11-22
remove unsafe protocols

JobberRT:
Removed `VMess`/`MTProto`/`Socks`(Why would you use this if there's encrypted protol to choose?)/`WireGuard`


JobberRT:
Sorry for the PR, GitKraken problem

Issue #1372, opened 2022-11-24
The shadowsocks2022 is not implemention of  proxy.UserManager feature?


thank243:
When will this feature be implemented? Thanks

loydcheung:
and ss2022 isn't supported within the “xtls-whatever” template neither as
 I tested.

thank243:
I will pr later.

PR #1401, opened 2022-12-03
update: Implement the proxy.UserManager of ss2022

thank243:
A first implements proxy.UserManager of shadowsocks2022
Considering implements shadowsocks2022 in xray-core may have better performance when crud users.
#1372

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,6 +54,7 @@ require (
 	golang.org/x/time v0.2.0 // indirect
 	golang.org/x/tools v0.3.0 // indirect
 	google.golang.org/genproto v0.0.0-20221118155620-16455021b5e6 // indirect
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
--- a/go.sum
+++ b/go.sum
@@ -330,7 +330,7 @@ google.golang.org/protobuf v1.26.0/go.mod h1:9q0QmTI4eRPtz6boOQmLYwt+qCgq0jsYwAQ
 google.golang.org/protobuf v1.28.1 h1:d0NfwRgPtno5B1Wa6L2DAG+KivqkdutMf1UhdNx175w=
 google.golang.org/protobuf v1.28.1/go.mod h1:HV8QOd/L58Z+nl8r43ehVNZIU/HEI6OcFqwMG9pJV4I=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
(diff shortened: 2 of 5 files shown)


yuhan6665:
Thanks a lot for your work!
How do you test this new feature? Can you give a try for something like
https://github.com/crossfw/Xray-API-documents/blob/main/examples/alterInbound.go#L49

thank243:
> Thanks a lot for your work! How do you test this new feature? Can you give a try for something like https://github.com/crossfw/Xray-API-documents/blob/main/examples/alterInbound.go#L49

Hello @yuhan6665 , I have tested on XrayR to invoke this interface. Its working normally. But when CURD user, it could have some performance problem. The ss2022 on sagernet haven't the remove user interface. Thus the implements just update all users when one user delete or add.

yuhan6665:
I see you are using UpdateUsersWithPasswords(). Currently Xray rely on sing-ss implementation. I think the simplest way is to add the following interface in sing-ss:

- AddUserWithPassword()
- RemoveUser()

This way, update a single user would not require lopp through the whole list and recreate every user again.
CC @nekohasekai 

I'm not familiar with XrayR. Is it in the end call AlterInbound() -> AddUserOperation() ?

thank243:
> I see you are using UpdateUsersWithPasswords(). Currently Xray rely on sing-ss implementation. I think the simplest way is to add the following interface in sing-ss:
> 
> * AddUserWithPassword()
> * RemoveUser()
> 
> This way, update a single user would not require loop through the whole list and recreate every user again. CC @nekohasekai
> 
> I'm not familiar with XrayR. Is it in the end call AlterInbound() -> AddUserOperation() ?

Yes, you are right. Add this interface is a good way by sing-ss. XrayR directly use proxy.UserManager interface to add or remove user. 

https://github.com/XrayR-project/XrayR/blob/212f0ff13591a263753dbc9b5e4962deaa61d1b8/service/controller/control.go#L58-L106

yuhan6665:
Since XrayR can make direct call, another option is to add "AddMultiUsers" and "RemoveMultiUsers" in core. That should also help to some degree.
Anyways, I think this is a great starting point. Hope to see more improvements in this area :) Thanks again!

nekohasekai:
If users need to be added or removed at runtime, synchronization locks must be used, which incurs a performance penalty.

thank243:
> If users need to be added or removed at runtime, synchronization locks must be used, which incurs a performance penalty.

So can we use sync.map to reduce impact on performance？

vangork:
@thank243 [XrayR](https://github.com/XrayR-project/XrayR) seems to implement another approach for online user management which doesn't align with the way of xray ifself. Can you also help to add a PR for `alterinbound` approach?

thank243:
https://github.com/XrayR-project/XrayR/blob/master/service/controller/userbuilder.go#L89-L97

vangork:
@thank243 it dosen't seem to support alterinboundrequest adduseroperation ss2022 server side. it fails with the error msg of `desc = app/proxyman/command: proxy is not a UserManager`

thank243:
Only `type MultiUserInbound struct` implemented userManager. You should confirm the inbound handler is that.

PR #1477, opened 2022-12-31
fix: When add feature will remove old one first

thank243:
Please check it, Maybe it have some data race.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -163,9 +163,9 @@ jobs:
       - name: Prepare to release
         uses: nick-fields/retry@v2
         with:
-          timeout_minutes: 60
-          retry_wait_seconds: 60
-          max_attempts: 60
+          timeout_minutes: 5
+          retry_wait_seconds: 30
+          max_attempts: 10
           command: |
             cp ${GITHUB_WORKSPACE}/README.md ./build_assets/README.md
             cp ${GITHUB_WORKSPACE}/LICENSE ./build_assets/LICENSE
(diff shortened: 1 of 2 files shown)


yuhan6665:
Thanks!
How about keep the AddFeature() as is, but add a new method called ReplaceFeature()? Also, can you post some reference code or document how would you use this new method?

thank243:
IMO, Only one each of `feature.Type` can exist in `instance.features`. Otherwise add same types into `instance.features` the latest `feature` will never be used. https://github.com/XTLS/Xray-core/blob/11ec77bc768a07a727d5f5836aa9fe4c8e20421c/core/xray.go#L37-L44
To ensure Xray instance works properly, the config must contain one Dispatcher, one InboundHandlerManager and one OutboundHandlerManager. Other features are optional. I could totally add optional features after core started up.
For example, I can add dns feature after core started up.
```go
	dnsInstance, err := serial.ToTypedMessage(dnsConfig).GetInstance()
	if err != nil {
		return err
	}
	obj, err := core.CreateObject(c.server, dnsInstance)
	if err != nil {
		return err
	}
	if feature, ok := obj.(features.Feature); ok {
		c.server.AddFeature(feature)
	}
```

thank243:
Hello @yuhan6665 , This PR seems have some problems. I'm confused with this codes. Could you tell the mechanism？
https://github.com/XTLS/Xray-core/blob/11ec77bc768a07a727d5f5836aa9fe4c8e20421c/app/dispatcher/default.go#L103-L107
Where is the ` dc dns.Client` come from? And how can I pass it?

yuhan6665:
I see. I don't think there are bug with the current AddFeature(). The config should have only one feature with the same type and the check here
https://github.com/XTLS/Xray-core/blob/11ec77bc768a07a727d5f5836aa9fe4c8e20421c/core/xray.go#L220-L224
Ensure there will not be more than one feature of the same type.

thank243:
> I see. I don't think there are bug with the current AddFeature(). The config should have only one feature with the same type and the check here
> 
> https://github.com/XTLS/Xray-core/blob/11ec77bc768a07a727d5f5836aa9fe4c8e20421c/core/xray.go#L220-L224
> 
> 
> Ensure there will not be more than one feature of the same type.

You are right. But when 3rd project quote this method could cause some problems. AddFeature is a public method. But it is not very important, so I closed this PR.

PR #1595, opened 2023-02-03
Add Napsternetv free mobile client

amirhakimnejad:
Usable specially on iOS since it is free


RPRX:
> Updated xray (v2ray) core to 1.6.1
> Updated Xray core to 1.6.0

这个项目是在 iOS 上集成了 Xray？不会被系统 kill 吗？希望有更多信息

amirhakimnejad:
After giving it some tests, it does not support xtls, though it supports vless.

amirhakimnejad:
It seems it does crash on iOS. I'm closing this for now

PR #1636, opened 2023-02-10
Improve ReshapeMultiBuffer

H1JK:


yuhan6665 (approved):
感谢大佬的高效实现

RPRX:
Wait

RPRX:
等我加了 REALITY 的代码再 merge，~~不然又要 rebase 了~~

yuhan6665:
> 等我加了 REALITY 的代码再 merge，~不然又要 rebase 了~

好~
~不过没有冲突应该无压力~

RPRX:
这个文件还好，~~前几天已经 rebase 到吐了，不想再来一次~~

proxy/vless/encoding/encoding.go
```diff
@@ -486,11 +486,17 @@ func ReshapeMultiBuffer(ctx context.Context, buffer buf.MultiBuffer) buf.MultiBu
 			if index <= 0 {
 				index = buf.Size / 2
 			}
-			buffer1 := buf.New()
+			var buffer1 *buf.Buffer
 			buffer2 := buf.New()
-			buffer1.Write(b.BytesTo(index))
 			buffer2.Write(b.BytesFrom(index))
-			b.Release()
+			if index >= buf.Size-21 {
```
yuhan6665, line 492: It seems we don't need this check and we can always do `b.Resize(0, index)`?

H1JK: @yuhan6665 What if `tlsApplicationDataStart` appears at the last 21 bytes?

And in the same way, it may also worth to consider the case where the newly created `buffer2` still does not meet the 21-byte free requirement. (I'm not sure)

yuhan6665: Normally a tls record is 23, 3, 3, len_high, len_low, data, tag(16 bytes). So very unlikely it appear at the last 21 bytes.
Note the current way Vision parse and understand the tls record in a simple way, and is not strictly correct. For instance, there could be a 1/16,777,216 chance that 23, 3, 3 appear in the data itself. Someday maybe we can have a real tls parser..
I think reducing unnecessary buffer ops is good enough for this pr.

H1JK: Well, it doesn't matter, I think.

yuhan6665:
Merged. Thanks again!

PR #1638, opened 2023-02-10
fix: closes mux rw as #1417

rurirei:
as per #1417


rurirei:
that's weird to call Interrupt as Close within transport pipe.

cross-hello:
תודה אתה

Mar 3, 2023 19:59:51 rurirei ***@***.***>:

PR #1646, opened 2023-02-11
Vision padding upgrade

yuhan6665:


yuhan6665:
I will update it as discussed tomorrow

PR #1660, opened 2023-02-17
Support SPKI Fingerprint Pinning

yinyue200:
Support SPKI Fingerprint Pinning for TLSObject.

This feature allows users to emulate the risk of man-in-the-middle attacks from implanted CA certificates on devices when non-self-signed certificates is using.

It provides similar functionality to HTTP Public Key Pinning. It allows allows certificates to be renewed normally, compared to the existing functionality.


yinyue200:
Sorry, I accidentally deleted the related branch, so I opened a new one.

PR #1790, opened 2023-03-13
Added user agent to gRPC

HirbodBehnam:
Hello
I'm not sure how this feature will be used but I added an option to set custom user agent header for gRPC client. This can be set in config file in `grpcSettings` as `user_agent`.
For example:
```json
"grpcSettings": {
  "serviceName": "myservice",
  "user_agent": "custom user agent"
}
```
I regenerated the pb file using this command:
```
protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative .\transport\internet\grpc\config.proto
```

This feature MIGHT be used in some CDN firewall rules to block traffic and do other stuff. Or maybe pretend to your CDN or middleware that you are chrome by setting the user agent.


RPRX:
感谢 PR，我写 REALITY 的 SpiderX 时发现有些网站必须设为非 Go 的 User-Agent 才能爬取，所以给 gRPC 加上这一项或许也是有用的（尽管目前浏览器无法直接调用 gRPC，所以 CDN 不应要求 User-Agent 为浏览器）

RPRX:
此外，目前 *ray 的 gRPC transport 只能设置 `serviceName` 并在其后加上 /Tun 作为 path，这一设计是比较糟糕的，为了修补它，我们可以在 `serviceName` 第一个字节为 `/` 时视其为 path，有时间写个 PR 吗？

HirbodBehnam:
As I said, I think changing the user agent of gRPC is kinda useless. I saw a tweet of a guy who blocked all user agents except Chrome's user agent in Cloudflare. Not sure how this would help but I got the idea from there.
About the `serviceName` you said I think it's sufficient to just change [this file](https://github.com/XTLS/Xray-core/blob/f1e35ad9d42ee0c0e95f6d2b49b858775881c7f7/transport/internet/grpc/encoding/customSeviceName.go) and make `Tun` and `TunMulti` strings variables, right?
These two strings must be changed to be dynamic and set from config file. However I didn't get what you ment from `we can treat it as a path when the first byte of serviceName is /`. (I used Google translate :)) Can't someone simply include `/` in their service name to create arbitrary paths? For example if I use `hello/world/this/is/my/service/name` as the service name, doesn't it make my requests go to `www.hostname.com/hello/world/this/is/my/service/name/Tun`?
I can try to sketch a PR for the `Tun` and `TunMulti` dynamic name. I think it's best to open a new issue and move the conversation there right?

RPRX:
> About the `serviceName` you said I think it's sufficient to just change [this file](https://github.com/XTLS/Xray-core/blob/f1e35ad9d42ee0c0e95f6d2b49b858775881c7f7/transport/internet/grpc/encoding/customSeviceName.go) and make `Tun` and `TunMulti` strings variables, right?

√

> However I didn't get what you ment from `we can treat it as a path when the first byte of serviceName is /`.

If `serviceName` starts with `/`, we should use it as the full path directly, without appending `/Tun` or `/TunMulti`.

> Can't someone simply include `/` in their service name to create arbitrary paths? For example if I use `hello/world/this/is/my/service/name` as the service name, doesn't it make my requests go to `www.hostname.com/hello/world/this/is/my/service/name/Tun`?

```go
func (c *Config) getNormalizedName() string {
	return url.PathEscape(c.ServiceName)
}
```

> I think it's best to open a new issue and move the conversation there right?

Of course.

RPRX:
感谢 PR

HSXX:
这个功能存在bug，在服务端看，xray在客户端设置的useragent后面多加了 grpc-go/1.54.0。

PR #1815, opened 2023-03-18
gRPC Custom Path

### LINKED ISSUE #1793: [Enhancement] Allow custom path and stream name for gRPC
We had a talk with @RPRX in #1790 about allowing custom paths in gRPC transport.
There is a recap of our idea.
1. We can assign custom stream names for `Tun` and `TunMulti`. These strings are currently hardcoded [here](https://github.com/XTLS/Xray-core/blob/f1e35ad9d42ee0c0e95f6d2b49b858775881c7f7/transport/internet/grpc/encoding/customSeviceName.go). Changing them shouldn't be a big deal.
2. Allow user to use absolute paths for gRPC. This can be determined from the prefix of `/` in `serviceName`. But this will bring us a problem. How we are going to distinguish traffic with `multiMode` set and not set?

I think implementing 1 is pretty simple and straight-forward. For 2, we can do this:
1. Check if `serviceName` has `/` suffix. If it does not have one, just act like before.
2. If it does have, we don't path escape the `serviceName`. Instead, we will use absolute path defined in the `serviceName` + the custom `Tun` and `TunMulti` stream name. For example if the `serviceName` is set to `/my/path` and our custom `Tun` stream name is `customTun`, we will send our request to `www.hostname.com/my/path/customTun`. On the client side, this looks easy to implement. The implementation should take place in [this file](https://github.com/XTLS/Xray-core/blob/f1e35ad9d42ee0c0e95f6d2b49b858775881c7f7/transport/internet/grpc/encoding/customSeviceName.go) again. However, on the server side we should make sure that the gRPC library does support paths as its service name.

HirbodBehnam:
In this PR, I added the ability to define custom full paths for gRPC as discussed in #1793. I added some tests as well just for the sake of it.
Here are two example configs:
Server:
```json
{
    "log": {
        "loglevel": "debug"
    },
    "inbounds": [{
            "listen": "127.0.0.1",
            "port": 12345,
            "protocol": "trojan",
            "settings": {
                "clients": [{
                        "password": "123456"
                    }
                ]
            },
            "streamSettings": {
                "network": "gun",
                "security": "none",
                "grpcSettings": {
                    "serviceName": "/my/sample/path/customTun|customTunMulti"
                }
            }
        }
    ],
    "outbounds": [{
            "protocol": "freedom"
        }
    ]
}
```
Client Tun:
```json
{
    "log": {
        "loglevel": "debug"
    },
    "inbounds": [{
            "listen": "127.0.0.1",
            "port": "10818",
            "protocol": "socks"
        }
    ],
    "outbounds": [{
            "protocol": "trojan",
            "settings": {
                "servers": [{
                        "address": "127.0.0.1",
                        "port": 12345,
                        "password": "123456"
                    }
                ]
            },
            "streamSettings": {
                "network": "gun",
                "security": "none",
                "grpcSettings": {
                    "serviceName": "/my/sample/path/customTun",
                    "multiMode": false
                }
            }
        }
    ]
}
```
Client Multi Tun:
```json
{
    "log": {
        "loglevel": "debug"
    },
    "inbounds": [{
            "listen": "127.0.0.1",
            "port": "10818",
            "protocol": "socks"
        }
    ],
    "outbounds": [{
            "protocol": "trojan",
            "settings": {
                "servers": [{
                        "address": "127.0.0.1",
                        "port": 12345,
                        "password": "123456"
                    }
                ]
            },
            "streamSettings": {
                "network": "gun",
                "security": "none",
                "grpcSettings": {
                    "serviceName": "/my/sample/path/customTunMulti",
                    "multiMode": true
                }
            }
        }
    ]
}
```

As you can see, we have the limitation that both `Tun` and `Tun Multi` endpoints must have the same sub-route. There is a very small problem with this implementation: Service names like `/` will crash the app. We need to make sure that if the service name starts with `/`, it has at least two `/` in it. We should probably somehow validate the config file.
This PR will close #1793

### DIFF
--- a/transport/internet/grpc/config.go
+++ b/transport/internet/grpc/config.go
@@ -22,7 +22,12 @@ func (c *Config) getServiceName() string {
 		return url.PathEscape(c.ServiceName)
 	}
 	// Otherwise new custom paths
-	return c.ServiceName[1:strings.LastIndex(c.ServiceName, "/")] // trim from first to last '/'
+	rawServiceName := c.ServiceName[1:strings.LastIndex(c.ServiceName, "/")] // trim from first to last '/'
+	serviceNameParts := strings.Split(rawServiceName, "/")
+	for i := range serviceNameParts {
+		serviceNameParts[i] = url.PathEscape(serviceNameParts[i])
+	}
+	return strings.Join(serviceNameParts, "/")
 }
 
 func (c *Config) getTunStreamName() string {
@@ -32,7 +37,7 @@ func (c *Config) getTunStreamName() string {
 	}
 	// Otherwise new custom paths
 	endingPath := c.ServiceName[strings.LastIndex(c.ServiceName, "/")+1:] // from the last '/' to end of string
-	return strings.Split(endingPath, "|")[0]
+	return url.PathEscape(strings.Split(endingPath, "|")[0])
 }
 
 func (c *Config) getMultiTunStreamName() string {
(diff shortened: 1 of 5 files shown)


RPRX:
感谢 PR，麻烦把 `MultiTun` 字样改成 `TunMulti`

RPRX:
再次感谢！

PR #1889, opened 2023-04-03
doc: include a new xray reality installation script

butterfly1924:
This pr adds a new way of install xray reality on your machine


yuhan6665:
thanks!

PR #1948, opened 2023-04-15
Add Libertea to web panel list

VZiChoushaDui:


RPRX:
WSS 可以一眼识别，建议加上 gRPC

VZiChoushaDui:
We currently use specific websocket URLs (and reroute them through haproxy to actual X-Ray process) for individual user management and limiting.

Can you think of any way to achieve the same result on a single port for gRPC?

(Same thing with the new REALITY protocol, we're looking for ways to implement it without breaking user management)

VZiChoushaDui:
Oh, just saw that gRPC configuration also works on a custom endpoint, and so it's probably doable to implement it on the same stack. Will look into it.

But the thing about REALITY remains, can you think of anything regarding this?

RPRX:
Added.

PR #1952, opened 2023-04-16
support memory check for socks inbound

KouYiGuo:
This PR gives socks inbound ability to check memory usage and try GC when out of memory.
It is highly similar to [conntrack killer](https://github.com/SagerNet/sing-box/blob/dev-next/common/dialer/conntrack/killer.go) of [sing-box](https://github.com/SagerNet/sing-box). ~~Maybe this PR will violate its GPL license?~~
Because most clients use tun2socks as the tun of Xray, I think only supporting socks is enough.
Memory usage still needs more improvements. This is the first step.

### DIFF
--- a/common/memory/errors.generated.go
+++ b/common/memory/errors.generated.go
@@ -1,0 +1,9 @@
+package memory
+
+import "github.com/xtls/xray-core/common/errors"
+
+type errPathObjHolder struct{}
+
+func newError(values ...interface{}) *errors.Error {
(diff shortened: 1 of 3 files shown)


RPRX:
如果你是参照它的代码写的并且写得很像的话，不太合适

其实可以简单粗暴地每分钟 GC+FreeOSMemory

KouYiGuo:
明白

PR #2033, opened 2023-05-04
Change X-UI-English to X-UI

sudospaes:
The X-UI-English panel is old and its development has been stopped. The replaced panel is a fork of X-UI-English in which the items are updated and its development continues


fazel32:
Hello, there is a problem in version 3. When we create an account and select the revise option, it takes a long time to revise and create an account.

PR #2042, opened 2023-05-06
feat: app/dns dns over tls

rurirei:
this introduces a simple to build a DoT server as per #1819 

look down and tell if or how to complete this.


RPRX:
感谢 PR，查看 https://github.com/XTLS/Xray-core/issues/1819#issuecomment-1694253636

PR #2087, opened 2023-05-15
Add iOS & macOS arm64 GUI client

Matin-B:
This pull request introduces three new graphical user interface (GUI) clients to execute V2Ray project configurations. The GUI clients provide a user-friendly interface for configuring and managing V2Ray settings efficiently.

Changes Made:

Added GUI client V2Box: [V2box - VPN Internet network proxy , Supports multiple proxy protocols, such as Shadowsocks, V2ray, Vmess, Trojan and Xray]

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,8 @@
   - [X-flutter](https://github.com/XTLS/X-flutter)
 - iOS & macOS arm64
   - [Mango](https://github.com/arror/Mango)
-  - [Wings X](https://apps.apple.com/app/wings-x/id6446119727)
+  - [Wings X](https://apps.apple.com/app/wings-x/id6446119727) (The program appears to have been removed from the App Store)
+  - [V2Box](https://apps.apple.com/us/app/v2box-v2ray-client/id6446814690)
 - macOS arm64 & x64
   - [V2RayXS](https://github.com/tzmax/V2RayXS)
   - [Wings X](https://apps.apple.com/app/wings-x/id6446119727)


RPRX:
感谢 PR，闭源软件需由其开发者提交

Matin-B:
WTF ?
This is an open-source project and I only want to help people to find easily clients for their os. and now must an author of a project suggest an app?

Matin-B:
> > WTF ? This is an open-source project and I only want to help people to find easily clients for their os. and now must an author of a project suggest an app?
> 
> 商店页面没有提到它是个开源程序，并且在GitHub上并没有仓库。你首先需要证明它是开源的。

I understand your concern about the lack of an open source client for Reality configs. However, I believe that there are several reasons why it is still important to introduce this client, even if it is not open source.

First, as you mentioned, there is currently no other client available in the App Store that can be used to connect to Reality configs. This means that if this client is not introduced, users will be unable to use Reality configs on their iOS devices.

Second, open source is not always the best option for every project. In some cases, it may be more important to focus on quality and security, even if that means sacrificing some degree of openness. This is especially true for projects that handle sensitive data, such as Reality configs.

Finally, it is important to remember that open source is not a binary choice. Even if this client is not open source, it can still be made available to the community under a permissive license. This would allow users to inspect the code and make changes if they need to, while still giving the developers control over the project.

In conclusion, I believe that the benefits of introducing this client outweigh the drawbacks. I urge you to reconsider your decision and make this client available to the community.

Issue #2132, opened 2023-05-23
 Question about Source Code Availability and Package Trustworthiness of the Hiddify panel


tavallaie:
I attempted to access the source code for the [hiddify](https://github.com/hiddify/hiddify-config) project on their GitHub repository, but I was unable to locate it. However, during my exploration of their setup, I discovered that they rely on a PyPI package named [hiddifypanel](https://pypi.org/project/hiddifypanel/). This package appears to be a binary, and it is unclear whether it contains any potential backdoors or security vulnerabilities. For reference, you can find the binary at the following [PyPI Inspector link](https://inspector.pypi.io/project/hiddifypanel/7.0.0.dev21/packages/83/86/cede2f5856f45d627cd3421267a4a9e1a397e0d2c0bead3d38a2a833de98/hiddifypanel-7.0.0.dev21-cp310-cp310-manylinux2014_aarch64.whl/hiddifypanel/__main__.py).

It is worth noting that the hiddify team is currently advertising that the XTLS-core team trusts their project. Therefore, I kindly request the repository maintainers to either remove or explicitly tag the project as a partially open source/commercial panel. Alternatively, it would be beneficial to conduct a thorough inspection of the package to ensure its integrity and absence of any potential backdoors in the future.

Thank you for your attention to this matter.

hiddify1:
Dear tavallaie,
Thank you for your comment.
All the api, clients, and also the configuration and connection parts are open source and not only open source but also automatically compile and publish while in that list there are many other close source applications and paid applications.


By the way, what you are talking is about the rebranding out always free UI for non commercial use and unfortunately our license does not allow you to rebrand it. 
If you want to have your name on the panel, you can build a new system with your name and use the open source version of hiddify-config. You just need to follow install.sh, run.sh format and in your panel provide json formated user informations.

yuhan6665:
closing as irrelevant

PR #2028, opened 2023-05-03
Update README.md

hiddify1:
Adding Hiddify Clients for windows and android that are a fork of V2RayN and V2RayNG with more features and adding auto server selector

PR #2135, opened 2023-05-24
Adding `security risks` flag for `Hiddify` in `README.md`

taha-ghadirian:
* 3f0a9b776da1d558d7ca060f79f20517dfb88331 Add security risks tag to [Hiddify](https://github.com/hiddify) projects due to #2132

@tavallaie Found some security risks in Hiddify and asked for change reverts of this Pull request: #2028, which described why in https://github.com/XTLS/Xray-core/pull/2028#issuecomment-1559530044, However, I think it's better not to revert changes of this commit, I'd rather have this security notices for their project until the Hiddify team, describe the reason of having binary, non-secure, packages in their open source and free project. 

Also, meanwhile, Hiddify started advertising their project with the slogan: `Verified by Xray-core`, because their name is in the Xray-Core readme file. So that's why we have become more suspicious of it.

_**This PR can be reverted after the closing of #2132, when Hiddify team made the desired considerations.**_


yuhan6665:
We prefer full open source but also understand author may want to prevent others from copying their work.
If you want to add a label, just say "closed source", don't say security risk. And you need to apply the same standard for all apps in the list. Thanks.

PR #2245, opened 2023-06-22
XRV should work without rawConn

maskedeken:
fix [https://github.com/XTLS/Xray-core/issues/2238](https://github.com/XTLS/Xray-core/issues/2238)


yuhan6665:
感谢佬！

PR #2258, opened 2023-06-26
Support regex matching with attr

yuhan6665:
Thanks to idea from @RPRX

### DIFF
--- a/app/router/condition.go
+++ b/app/router/condition.go
@@ -1,6 +1,7 @@
 package router
 
 import (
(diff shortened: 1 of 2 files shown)


RPRX:
感谢 PR，预处理一下性能会更好，应该是并发安全的，参考
https://github.com/XTLS/Xray-core/blob/07389eca9618ad50ecb9cf524857ceef16ac392a/transport/internet/reality/reality.go#L253

PR #2315, opened 2023-07-10
fix udp encode error when shadowsocks  cipher none

jnxyatmjx:
when shadowsocks cipher is none，panic below
proxy/shadowsocks: dropping invalid UDP packet from: udp:127.0.0.1:43893 > proxy/shadowsocks: len(bs) <= 32


RPRX:
如果直接删掉这个长度检查，可能会导致后续代码的 panic，请测试一下对非 none 的服务端发小包

PR #2335, opened 2023-07-16
#2017 fix: Check if address is nil for IsValid() function.

cty123:
#2017 Panic due to nil dereferencing

While I was investigating #2017, I saw the error log reported by @base510 

```
2023/04/30 20:00:38 [Error] common/net: invalid IP format: []
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x982078]

goroutine 92630 [running]:
github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).getUoTConnection(0xc0001d5ce0, {0x13eeb40, 0xc0008e5290}, {{0x0?, 0x0?}, 0x7f?, 0x0?})
github.com/xtls/xray-core/app/proxyman/outbound/uot.go:14 +0x58
github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).Dial(0xc0001d5ce0, {0x13eeb40?, 0xc0008e5290?}, {{0x0, 0x0}, 0xb, 0x2})
github.com/xtls/xray-core/app/proxyman/outbound/handler.go:214 +0xa26
github.com/xtls/xray-core/proxy/freedom.(*Handler).Process.func1()
github.com/xtls/xray-core/proxy/freedom/freedom.go:140 +0x2c6
github.com/xtls/xray-core/common/retry.(*retryer).On(0xc00125dd08, 0xc00125dd60)
github.com/xtls/xray-core/common/retry/retry.go:27 +0xdb
github.com/xtls/xray-core/proxy/freedom.(*Handler).Process(0xc00021bce0, {0x13eeb40, 0xc0008e5290}, 0xc0008e2460, {0x13e6748, 0xc0001d5ce0})
github.com/xtls/xray-core/proxy/freedom/freedom.go:126 +0x4b0
github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).Dispatch(0xc000c07920?, {0x13eeb40, 0xc0008e5290}, 0xc0008e2460)
github.com/xtls/xray-core/app/proxyman/outbound/handler.go:147 +0x211
github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).routedDispatch(0xc0001c7f80, {0x13eeb40, 0xc0008e5290}, 0xc0008e2460, {{0x0, 0x0}, 0xb, 0x2})
github.com/xtls/xray-core/app/dispatcher/default.go:492 +0xbfe
[...]

PR #2341, opened 2023-07-18
fix: revert previous change on IsValid() function

cty123:
My previous change: https://github.com/XTLS/Xray-core/pull/2335 made a bold assumption that in order for the destination to be valid, the IpAddress must not be nil. But throughout the code base, there might be places where the assumption is false, as the error log reported by @base510,

```
goroutine 47593 [running]:
github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).Dispatch(0xc0002203c0, {0x13bf960, 0xc0002264b0}, {{0x0, 0x0}, 0xb, 0x2})
github.com/xtls/xray-core/app/dispatcher/default.go:287 +0x519
github.com/xtls/xray-core/common/mux.(*Server).Dispatch(0x0?, {0x13bf960?, 0xc0002264b0?}, {{0x0?, 0x0?}, 0x11?, 0x0?})
github.com/xtls/xray-core/common/mux/server.go:41 +0xcf
...
```

For now we should revert the change and add the nil check in the function instead.

### DIFF
--- a/app/proxyman/outbound/uot.go
+++ b/app/proxyman/outbound/uot.go
@@ -11,6 +11,9 @@ import (
 )
 
 func (h *Handler) getUoTConnection(ctx context.Context, dest net.Destination) (stat.Connection, error) {
+	if dest.Address == nil {
+		return nil, newError("nil destination address")
+	}
 	if !dest.Address.Family().IsDomain() {
 		return nil, os.ErrInvalid
 	}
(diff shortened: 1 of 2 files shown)


RPRX:
可以 force-push

@yuhan6665

yuhan6665:
感谢相关pr！
~~暂时不force~~

PR #2428, opened 2023-08-10
Replace `github.com/golang/protobuf` with its successor

hax0r31337:
Since `google.golang.org/grpc` still uses this library, there's won't noticeable change on binary size


yuhan6665:
感谢佬！终于把这个过时依赖干掉了

PR #2462, opened 2023-08-19
fix(proxy): removed the udp payload length check when encryption is disabled

cty123:
Following up on the issue #2310 and the previous PR #2315, I did some investigation and found the length check is not necessary when the encryption is disabled, even though it's not very common use-case to disable encryption for Shadowsocks protocol. 

Given the current way the Shadowsocks header is parsed, the length check on the payload is only necessary when the validator decrypts the payload, in which there might be array out of bound exception. To fix the problem, I moved the length check inside the validator, such that if the payload size is too short, the validator wouldn't even attempt to decrypt the payload. Meanwhile, I added a set of unit tests to cover the change.

### DIFF
--- a/proxy/shadowsocks/protocol.go
+++ b/proxy/shadowsocks/protocol.go
@@ -4,6 +4,7 @@ import (
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/sha256"
+	"errors"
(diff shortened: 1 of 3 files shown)


yuhan6665:
Thanks for your fix and code clean up!

Issue #2460, opened 2023-08-19
Xray stop working if testing the VMESS  config that miss configured.


M450ud:
If you use Xray for vmess config's which the 'Path' parameter is only '/' char the app will stop working and go to panic mode (according to the log). also, the latest version has the bug.
the bug cause problem when checking a list; the user must find the buggy config by checking it one by one.

M450ud:
this is a sample config:
vmess://eyJhZGQiOiI5NC4xNTQuMTU3LjEwMCIsImFpZCI6IjAiLCJob3N0IjoiIiwiaWQiOiI0YTc1YzNiYS1mMmU5LTExZWQtYmYxYi1mMjNjOTEzNjlmMmQiLCJuZXQiOiJ3cyIsInBhdGgiOiIvIiwicG9ydCI6IjgwIiwicHMiOiLwn4es8J+Hp0dCIiwic2N5IjoiYXV0byIsInNuaSI6IiIsInRscyI6IiIsInR5cGUiOiIiLCJ2IjoiMiJ9

and log :

panic: runtime error: slice bounds out of range [1:0]
goroutine 1878 [running]:
github.com/xtls/xray-core/transport/internet/grpc.(*Config).getServiceName(0xc0003d4c40)
	github.com/xtls/xray-core@v1.0.0/transport/internet/grpc/config.go:25 +0x15e
github.com/xtls/xray-core/transport/internet/grpc.dialgRPC({0x11a4d60, 0xc001355050}, {{0x11a61b8?, 0xc0007481e8?}, 0x2?, 0x0?}, 0xd229a7?)
	github.com/xtls/xray-core@v1.0.0/transport/internet/grpc/dial.go:65 +0x4d5
github.com/xtls/xray-core/transport/internet/grpc.Dial({0x11a4d60, 0xc001355050}, {{0x11a61b8, 0xc0007481e8}, 0x823, 0x2}, 0xc001354fc0?)
	github.com/xtls/xray-core@v1.0.0/transport/internet/grpc/dial.go:27 +0x18d
github.com/xtls/xray-core/transport/internet.Dial({0x11a4d60, 0xc001355050}, {{0x11a61b8?, 0xc0007481e8?}, 0x81e8?, 0xc0?}, 0x6?)
	github.com/xtls/xray-core@v1.0.0/transport/internet/dialer.go:58 +0x1ef
github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).Dial
(
0xc000ce0080
[...]

PR #2464, opened 2023-08-20
fix(config): fix grpc cofnig parsing when service name only has one '/' char

cty123:
Following up on #2460, it looks like the service name parsing assumes the service name string in the config file always contains more than 2 '/' characters, which is not always the case. 

```go
lastIndex := strings.LastIndex(c.ServiceName, "/")   // Might return 0
c.ServiceName[1:strings.LastIndex(c.ServiceName, "/")]  // Might become c.ServiceName[1:0] ==> slice bounds out of range [1:0]
```

And when it happens, the entire program would crash due to ```panic: runtime error: slice bounds out of range [1:0]``` error. To reduce the amount of application panic with error messages full of golang jargon, that create confusion to new users, I made it the service name parsing safe and would return empty string if it can't find any.

Meanwhile, since I am not too familiar with GRPC, I am not exactly sure if it's legitimate to have the service name as an empty string. We could add this validation logic later on during config parsing to explicitly block empty service name.

### DIFF
--- a/transport/internet/grpc/config.go
+++ b/transport/internet/grpc/config.go
@@ -21,8 +21,13 @@ func (c *Config) getServiceName() string {
 	if !strings.HasPrefix(c.ServiceName, "/") {
 		return url.PathEscape(c.ServiceName)
 	}
+
 	// Otherwise new custom paths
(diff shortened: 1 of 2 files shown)


yuhan6665:
Thank you!

Issue #1933, opened 2023-04-11
[bug] CIDR mismatch


Loyalsoldier:
You guys should fix this old bug from V2Ray.

Reference: https://github.com/v2fly/v2ray-core/pull/1155
Solution: https://github.com/v2fly/v2ray-core/pull/1157

Note: https://github.com/inetaf/netaddr has been deprecated, use https://pkg.go.dev/go4.org/netipx instead.

kdurov:
感谢帮开 issue。
我还想看看具体什么情况能 reproduce 后开的。

RPRX:
感谢反馈

PR #2489, opened 2023-08-24
fix(app): fixed a bug in geoip matching with refactoring.

cty123:
As raised in #1933, the existing in-house CIDR matching mechanism can produce incorrect result in some cases. The issue was originally brought up from v2ray-core project https://github.com/v2fly/v2ray-core/pull/1155. This PR ported most of the commit from the v2ray-core commit https://github.com/v2fly/v2ray-core/pull/1157 by @Loyalsoldier.

Here's an example for the bug(also as a new test cases I added with the PR),

```go
func TestGeoIPMatcherRegression(t *testing.T) {
	cidrList := []*router.CIDR{
		{Ip: []byte{98, 108, 20, 0}, Prefix: 22},
		{Ip: []byte{98, 108, 20, 0}, Prefix: 23},
	}

	matcher := &router.GeoIPMatcher{}
	common.Must(matcher.Init(cidrList))

	testCases := []struct {
		Input  string
		Output bool
	}{
		{
			Input:  "98.108.22.11",
			Output: true,
		},
		{
			Input:  "98.108.25.0",
			Output: false,
		},
	}

	for _, testCase := range testCases {
		ip := net.ParseAddress(testCase.Input).IP()
		actual := matcher.Match(ip)
		if actual != testCase.Output {
			t.Error("expect input", testCase.Input, "to be", testCase.Output, ", but actually", actual)
		}
	}
}
```

If you copy this over to the ```condition_geoip_test.go``` file on ```main``` branch, the test would fail.
```text
--- FAIL: TestGeoIPMatcherRegression (0.00s)
    /home/ctydw/Projects/Xray-core/app/router/condition_geoip_test.go:154: expect input 98.108.22.11 to be true , but actually false
```
But this is not really expected, because ```98.108.20.0/22``` is from ```98.108.20.0``` to ```98.108.23.255```, so it is in the range. Instead of fixing the algorithm itself, I do think it makes sense to replace the in-house algorithm with the netipx library.


RPRX:
感谢 PR，看起来这两个 test 需修复：TestGeoIPMatcher6US、TestRoutingRule

另外请基于 main 分支 rebase，并升级（若有）新引入的依赖 go4.org/netipx

RPRX:
~~担心的事情还是发生了，麻烦重新 rebase 一下。。。~~

RPRX:
看起来 test 已经没问题了

cty123:
没事，反正我能force push，之前写错一个小地方，debug了半天发现是低级错误。。。

RPRX:
感谢 PR

PR #2505, opened 2023-08-31
fix typo in default.go

dyhkwong:
`strings.HasPrefix(p, protocolString)` is for `strings.HasPrefix("fakedns+others", "fakedns")`.

I tested that "fakedns+others" does not work at all so this fix is in fact meaningless, only making "fakedns+others" identical to "fakedns". Maybe you can remove related code.

related:
https://github.com/v2fly/v2ray-core/commit/1b0e046
https://github.com/v2fly/v2ray-core/compare/9bbc3af...3c0aff7 (incorrect fix)


RPRX:
> Maybe you can remove related code.

可以直接包含在本 PR 内，~~或者 make "fakedns+others" work~~

PR #2582, opened 2023-09-20
set sockopt interface for darwin

hossinasaadi:
sockopt->interface now works in darwin ios/macos
```
      "streamSettings": {
        "sockopt": {
          "interface": "en0"
        }
      },

```


yuhan6665:
thanks for your contribution!

Issue #2617, opened 2023-10-06
IP format checking not correct in local dns client


kyze8439690:
Problem line: https://github.com/XTLS/Xray-core/blob/e244db76fbf871cc61f2f4d99f2676be27c2d9e8/features/dns/localdns/client.go#L36

IPv4 format address is recognized as ipv6 format address.

Code below should fix this issue:

```go
for _, ip := range ips {
	parsed := net.IPAddress(ip)
	if parsed != nil {
		parsedIPs = append(parsedIPs, parsed.IP())
	}
	if ip.To4() != nil {
		ipv4 = append(ipv4, ip)
	}
	if ip.To16() != nil {
		ipv6 = append(ipv6, ip)
	}
}
```

chika0801:
Thanks for the fix, you can send a request at https://github.com/XTLS/Xray-core/pulls.

kyze8439690:
> I don't think your fix will work because `ip.To16()` will always success for IPv4 address.

But it always  check `ip.To4()` first.

us254:
```

for _, ip := range ips {
	parsed := net.ParseIP(ip)
	if parsed != nil {
		switch {
		case parsed.To4() != nil:
			ipv4 = append(ipv4, parsed)
		case parsed.To16() != nil:
			ipv6 = append(ipv6, parsed)
		}
	}
}

```

kyze8439690:
> Although `len(ip)` is not a good check, can you point out how to trigger `net.Lookup` to return non-4-byte IPv4 addresses for a domain in this use case? golang/go@c53390b

Android 13, trigger dns query on a ipv4 only network.

kyze8439690:
> Although `len(ip)` is not a good check, can you point out how to trigger `net.Lookup` to return non-4-byte IPv4 addresses for a domain in this use case? [golang/go@c53390b](https://github.com/golang/go/commit/c53390b)

I think this commit do fix this issue, after I update golang to 1.21.2 and rebuild the core, this issue is gone.

PR #2618, opened 2023-10-06
fix dns client ip format detect error

kyze8439690:
Problem line: https://github.com/XTLS/Xray-core/blob/e244db76fbf871cc61f2f4d99f2676be27c2d9e8/features/dns/localdns/client.go#L36

IPv4 format address is recognized as ipv6 format address.

Code below should fix this issue:

```go
for _, ip := range ips {
	parsed := net.IPAddress(ip)
	if parsed != nil {
		parsedIPs = append(parsedIPs, parsed.IP())
	}
	if ip.To4() != nil {
		ipv4 = append(ipv4, ip)
	}
	if ip.To16() != nil {
		ipv6 = append(ipv6, ip)
	}
}
```

### DIFF
--- a/features/dns/localdns/client.go
+++ b/features/dns/localdns/client.go
@@ -33,10 +33,10 @@ func (*Client) LookupIP(host string, option dns.IPOption) ([]net.IP, error) {
 		if parsed != nil {
 			parsedIPs = append(parsedIPs, parsed.IP())
 		}
-		if len(ip) == net.IPv4len {
+		if ip.To4() != nil {
 			ipv4 = append(ipv4, ip)
 		}
-		if len(ip) == net.IPv6len {
+		if ip.To16() != nil {
 			ipv6 = append(ipv6, ip)
 		}
 	}


us254:
https://github.com/XTLS/Xray-core/issues/2617#issuecomment-1752159171

yuhan6665:
Thanks for the code and comments. @kyze8439690 It seem you need to add a "else" at least, otherwise v4 will be put in ipv6 list as well

us254:
>Thanks for the code and comments. @kyze8439690 It seem you need to add a "else" at least, otherwise v4 will be put in ipv6 list as well
```
for _, ip := range ips {
    parsed := net.ParseIP(ip)
    if parsed != nil {
        parsedIPs = append(parsedIPs, parsed)
    }
    if ip.To4() != nil {
        ipv4 = append(ipv4, ip)
    } else if ip.To16() != nil {
        ipv6 = append(ipv6, ip)
    }
}
```

kyze8439690:
It seems that this issue is gone after I upgrade golang to 1.21.2 and rebuild the core.

Issue #2606, opened 2023-10-02
Bug: MultiFile Multi inbound issue


hiddify-com:
When we have multifile configuration with multi inbound, files with multiple inbounds will be ignored.

### x.json:
``` 
{
        "inbounds": [
       {
            "tag": "x1",
            "listen": "127.0.0.1",
            "port": 1001,
            "protocol": "socks"
        },
       {
            "tag": "x2",
            "listen": "127.0.0.1",
            "port": 1002,
            "protocol": "socks"
        }
        ]
}
```
### y.json:
``` 
{
        "inbounds": [
       {
            "tag": "y1",
            "listen": "127.0.0.1",
            "port": 2001,
            "protocol": "socks"
        },
       {
            "tag": "y2",
            "listen": "127.0.0.1",
            "port": 2002,
            "protocol": "socks"
        }
        ]
}
```

### z.json:
``` 
{
        "inbounds": [
       {
            "tag": "z1",
            "listen": "127.0.0.1",
            "port": 3001,
            "protocol": "socks"
        }
        ]
}
```
### The Problem
Inbounds of y.json override the inbounds in xray so inbounds in x.json will be ignored.

I have tried to do pull request but i could not identify the source of the issue.

### logs:

```
...
2023/10/02 12:40:03 [Info] infra/conf/serial: Reading config: x.json
2023/10/02 12:40:03 [Info] infra/conf/serial: Reading config: y.json
2023/10/02 12:40:03 [Info] infra/conf/serial: Reading config: z.json
xctl> [ z.json ] appended inbound with tag:  z1
2023/10/02 12:40:04 [Info] transport/internet/tcp: listening TCP on 127.0.0.1:2001
2023/10/02 12:40:04 [Info] transport/internet/tcp: listening TCP on 127.0.0.1:2002
2023/10/02 12:40:04 [Info] transport/internet/tcp: listening TCP on 127.0.0.1:3001
2023/10/02 12:40:04 [Warning] core: Xray 1.8.4 started

```

BioniCosmos:
@yuhan6665 I think we should revert changes #2655 and #2659 because the behaviour in the issue is expected.

The logic is [inherited from V2Ray](https://github.com/v2ray/v2ray-core/releases/tag/v4.23.0), so these are breaking changes.

[多文件配置 | Project X](https://xtls.github.io/config/features/multiple.html#%E6%95%B0%E7%BB%84)

yuhan6665:
[...]

PR #2655, opened 2023-10-21
Override inbounds in multi config files

hossinasaadi:
when use multi config file inbounds not appends :
` ./xray -c config.json -c config2.json -c config3.json`
issue mentioned in https://github.com/XTLS/Xray-core/issues/2606


yuhan6665:
thanks for your pr! please take a look at this test `TestConfig_Override`

hossinasaadi:
> thanks for your pr! please take a look at this test `TestConfig_Override`

@yuhan6665  seems test fails cause of wrong test.
 i think with this PR, Inbound Overriding work much better, right?

yuhan6665:
Agree. Let's give it a try :)

PR #2725, opened 2023-11-13
#2605: Add safety check for type casting for QUIC dialer

cty123:
Issue #2605 brought up real problem that QUIC dialer doesn't support sockopt at the moment. Inside `internet.DialSystem(...)` function, one of the branch that involves `redirect(...)` returns `cnc.connection` instance that is currently unhandled by the code logic, and thus caused program panic during runtime.

```go
rawConn, err := internet.DialSystem(ctx, dest, sockopt)
```

```go
// DialSystem calls system dialer to create a network connection.
func DialSystem(ctx context.Context, dest net.Destination, sockopt *SocketConfig) (net.Conn, error) {
	...
	if obm != nil && len(sockopt.DialerProxy) > 0 {
		nc := redirect(ctx, dest, sockopt.DialerProxy)
		if nc != nil {
			return nc, nil
		}
	}
        ...
}
```

It seems the sockopt support for QUIC protocol requires a couple changes including making `cnc.connection` public, such that we can handle in dialer, along with some thorough tests, this commit simply adds safety check to explicity state the fact that QUIC isn't working with sockopt. And the implementation of the feature can be scheduled later on.

### DIFF
--- a/transport/internet/quic/dialer.go
+++ b/transport/internet/quic/dialer.go
@@ -146,10 +146,19 @@ func (s *clientConnections) openConnection(ctx context.Context, destAddr net.Add
 			return qlog.NewConnectionTracer(&QlogWriter{connID: ci}, p, ci)
 		},
 	}
-	udpConn, _ := rawConn.(*net.UDPConn)
-	if udpConn == nil {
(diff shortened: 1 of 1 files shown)


yuhan6665:
Thanks! I will re-push 1.8.5 later to include it.

PR #2734, opened 2023-11-18
Add brutal support for xray-core.

KoriIku:
This PR is only for showcasing the code. Further fixes and improvements are needed. For detailed information, please refer to the issue https://github.com/XTLS/Xray-core/issues/2735.


H1JK:
TCP Brutal needs negotiation to work properly. Only set the user configured rate to one side is not the correct use case of it.

> without the need for client adaptation

This is wrong.

KoriIku:
> TCP Brutal needs negotiation to work properly. Only set the user configured rate to one side is not the correct use case of it.
> 
> > without the need for client adaptation
> 
> This is wrong.

Let's clarify further.

In the configuration of Brutal, there is a "rate" parameter used to represent bandwidth. When setting the rate on the server side, it indicates the server-to-client speed, which, for users, is the download speed.

In most scenarios, download traffic is predominant, so configuring Brutal on the server side can bring significant optimization effects.

The term "negotiation" refers to the client informing the server of its own download speed.

In my case, since my server is only for personal use, and all devices have similar bandwidth sizes, I can directly configure this rate in the server's configuration file. In this situation, negotiation is unnecessary.

Of course, rate negotiation can provide better adaptability. For example, mobile devices and home networks often have different bandwidths, and negotiation can lead to a better user experience. However, this negotiation process is not essential for Brutal itself.

To avoid ambiguity, I propose updating the description in the issue to "No need for client adaptation when the user can determine their required download speed." What do you think?

My idea is to first implement a fixed configuration option on the server to enable Brutal, which is a more convenient and less disruptive solution, allowing users to try Brutal at a smaller cost. While negotiating the rate is beneficial, modifying existing protocols, adapting for all clients, requires more developer involvement, time, and raises the user's learning curve.

@tobyxdd Hello, I apologize for disturbing you. Could you please confirm if my understanding of Brutal is correct?

tobyxdd:
Thanks for the support, but while your understanding of Brutal is correct, I agree with the reply above that this is not the right way to implement it. Clients should send their bandwidth to the server as only they know their own bandwidth. Your approach doesn't give the clients any way to specify this, or even know what rate the server is using. It may work fine for your personal use case, but in this state I don't think it's something Xray will accept.

KoriIku:
Thank you for your response. I believe your perspective is quite accurate. Apparently, I've only considered a very narrow personal use case. For a broader audience of typical users, this falls short and may even pose a risk of increasing packet loss. I've oversimplified the issue; a more comprehensive design is indeed necessary.

PR #2776, opened 2023-12-01
Add link to tutorial for "Xray with WireGuard inbound"

gorilla800:
Add link to tutorial for "Xray with WireGuard inbound" using the new WireGuard inbound feature added in Xray-core v1.8.6.


yuhan6665:
Looks good! Thanks for your documents

PR #2818, opened 2023-12-15
Cleanup sing buffer usage

H1JK:
#2805 fixed compatibiltiy with newer SagerNet/sing versions. However, it is not a complete upgrade and only removed the usage to a removed function, and there are some deprecated functions still in use which should be migrated.

This PR updated sing to v0.3.x beta versions, removed stack buffer related codes, improved some buffer release scenarios and migrated FullReset usage.


yuhan6665:
感谢！

PR #2851, opened 2023-12-24
fix(conf): add Windows support for Unix Domain Socket

yin1999:
## Summary

add Windows support for Unix Domain Socket.

## Supports

[Windows Insider Build 17063 and later](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/) does support Unix Domain Socket, but the UDS check logic has blocked using UDS in windows.

## Tests

config:

```yaml
log:
  loglevel: warning

inbounds:
  - listen: "C:\\Absolute-path\\to\\xray.sock"
    tag: xray
    protocol: http
    sniffing:
      enabled: true
      destOverride: ["http", "tls"]
    settings:
      auth: noauth
      allowTransparent: false

outbounds:
  - tag: freedom
    protocol: freedom

routing:
  domainStrategy: AsIs
  # all pass
  rules:
    - type: field
      inboundTag: xray
      outboundTag: freedom
```

Run `curl` (with [PowerShell Core](https://github.com/PowerShell/PowerShell)) to test the http proxy server:

```bash
curl -I --unix-socket 'C:\\Absolute-path\\to\\xray.sock' -X connect http:/ --next https://www.baidu.com/
```

### DIFF
--- a/infra/conf/xray.go
+++ b/infra/conf/xray.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/xtls/xray-core/app/dispatcher"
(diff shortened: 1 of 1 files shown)


Fangliding:
@yin1999 It would be better add add a support check and return an error log if current version is not supported

yin1999:
> @yin1999 It would be better add add a support check and return an error log if current version is not supported

The original absolute path check method is only applicable to `*nix`, because the absolute path of Windows starts with the drive letter, so `listenDS` is always `false` when configuring Unix Domain socket on Windows.

Log with the build from main branch with the same configuration:

```plain
Xray 1.8.6 (Xray, Penetrates Everything.) Custom (go1.21.4 windows/amd64)
A unified platform for anti-censorship.
Failed to start: main: failed to load config files: [.\config.yaml] > infra/conf: unable to listen on domain address: xxxxx\Xray-core\build_assets\xray.sock
```

PR #2887, opened 2024-01-02
common/singbridge: fix dropped context

alrs:
This picks up a dropped `ctx` variable passed into `XrayOutboundDialer.DialContext()` but then never used.


yuhan6665:
感谢！

Issue #2898, opened 2024-01-05
最近添加的round-robin balancing strategy好像有点问题


yomnxkcs:
测试结果
```bash
=== RUN   TestRoundRobinBalancingStrategy
    round_robin_test.go:24: [ok] expected: a picked: a
    round_robin_test.go:24: [ok] expected: b picked: b
    round_robin_test.go:24: [ok] expected: c picked: c
    round_robin_test.go:22: [fail] expected: b picked: a
    round_robin_test.go:22: [fail] expected: c picked: b
    round_robin_test.go:22: [fail] expected: d picked: c
    round_robin_test.go:22: [fail] expected: e picked: a
--- FAIL: TestRoundRobinBalancingStrategy (0.00s)
FAIL
FAIL    github.com/xtls/xray-core/testing       0.122s
FAIL
```

测试代码
```go
package balancer_test

import (
	"testing"

	"github.com/xtls/xray-core/app/router"
)

func TestRoundRobinBalancingStrategy(t *testing.T) {

	tags1 := []string{"a", "b", "c"}

	// ./xray api rmo "a"
	// ./xray api ado "d" "e"
	tags2 := []string{"b", "c", "d", "e"}

	rr := router.NewRoundRobin(tags1)
	for _, tags := range [][]string{tags1, tags2} {
		for _, exp := range tags {
			tag := rr.PickOutbound(tags)
			if tag != exp {
				t.Error("[fail]", "expected:", exp, "picked:", tag)
			} else {
				t.Log("[ok]", "expected:", exp, "picked:", tag)
			}
		}
	}
}
```

hossinasaadi:
round robin works like mentioned here https://github.com/XTLS/Xray-core/pull/2844#issuecomment-1868881702

yomnxkcs:
> round robin works like mentioned here [#2844 (comment)](https://github.com/XTLS/Xray-core/pull/2844#issuecomment-1868881702)

Parameter of PickOutbound() is not a constant. It could be changed by api call.
e.g.  `./xray api rmo "a"`

hossinasaadi:
> Parameter of PickOutbound() is not a constant. It could be changed by api call. e.g. `./xray api rmo "a"`



you add balancer selector as prefix ("proxy") :
```json
  "balancers" : [{
            "tag": "b1",
            "selector": ["proxy"],
            "strategy": {
                "type":"roundRobin"
            }
          }
        ]

```
so when add new outbound with api you should consider prefix as well.

[...]

PR #2906, opened 2024-01-07
fix roundRobin strategy behavior when using api

hossinasaadi:
issue: https://github.com/XTLS/Xray-core/issues/2898


yuhan6665:
Thanks! I love it

PR #2914, opened 2024-01-08
sort Outbound selector output

hossinasaadi:
as mentioned https://github.com/XTLS/Xray-core/issues/2898#issuecomment-1880261654
outbound selector tags output may not run by order in :
https://github.com/XTLS/Xray-core/blob/3f0bc134298cf512bd29660221a030db91594faf/app/proxyman/outbound/outbound.go#L150
as mentions in golang ref 
See https://golang.org/ref/spec#For_statements

`The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced. If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next. If the map is nil, the number of iterations is 0.
`

### DIFF
--- a/app/proxyman/outbound/outbound.go
+++ b/app/proxyman/outbound/outbound.go
@@ -152,6 +152,7 @@ func (m *Manager) Select(selectors []string) []string {
 		for _, selector := range selectors {
 			if strings.HasPrefix(tag, selector) {
 				tags = append(tags, tag)
+				break
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


yomnxkcs:
If anyone misconfigured like below, each tag will be added twice.
```json
{
  "routing": {
    "balancers": [
      {
        "selector": [
          "pro",
          "proxy"
        ]
      }
    ]
  }
}
```
Maybe we should add a break after append().
Besides that, all looks good.

hossinasaadi:
> break

that was good point. 👍

yuhan6665:
Looks good! Thanks all

PR #2941, opened 2024-01-15
Fix #2654

dyhkwong:
Fix #2654.

P.S. similar code can be found via `go get github.com/sagernet/xray-core@v1.4.6-0.20210926233017-dba63bc71072`


yuhan6665:
感谢佬！

PR #3037, opened 2024-02-20
chore(ci): use go version file to specify the go version

yin1999:
Use [go version file to get go version](https://github.com/actions/setup-go/tree/main?tab=readme-ov-file#getting-go-version-from-the-gomod-file) instead of manually specifying.


yuhan6665:
感谢！

RPRX:
这个改动是合理的

PR #3038, opened 2024-02-21
FakeDNS return TTL=1 and drop HTTPS QType 65 request

yuhan6665:
Co-authored-by: qwerr0


loskiq:
@yuhan6665 hi. I want to ask for what purpose did you change the TTL value to 1? wouldn't that put an extra load on fake dns?

xqzr:
> wouldn't that put an extra load on fake dns?

但减少了，客户端的缓存负担

PR #3084, opened 2024-03-03
New Vision traffic pattern: wide range+

yuhan6665:
An experiment

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -284,14 +284,14 @@ func XtlsPadding(b *buf.Buffer, command byte, userUUID *[]byte, longPadding bool
 	if b != nil {
 		contentLen = b.Len()
 	}
-	if contentLen < 900 && longPadding {
-		l, err := rand.Int(rand.Reader, big.NewInt(500))
+	if contentLen < 600 && longPadding {
+		l, err := rand.Int(rand.Reader, big.NewInt(3400))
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
us254, line 1:
```go
if b != nil {
    contentLen = b.Len()
}

// Step 1: Randomize the padding length range
paddingRange := rand.Intn(1000) + 500

// Step 2: Use a more complex padding length calculation
secretKey := []byte("your-secret-key")
hash := sha256.Sum256(b.Bytes())
paddingFactor := int(binary.BigEndian.Uint32(hash[:4])) % 100

if contentLen < 600 && longPadding {
    paddingLen = int32(paddingRange + paddingFactor)
} else {
    paddingLen = int32(rand.Intn(paddingRange) + paddingFactor)
}

// Step 3: Add random delays between packets
time.Sleep(time.Duration(rand.Intn(100)) * time.Millisecond)

// Step 4: Use a different padding algorithm
paddingData := make([]byte, paddingLen)
_, err := crypto.Rand.Read(paddingData)
if err != nil {
    newError("failed to generate padding").Base(err).WriteToLog(session.ExportIDToError(ctx))
}
// Append the padding data to the content
b.Write(paddingData)
```

yuhan6665:
Move to seed

PR #3115, opened 2024-03-08
Add a new xray wrapper

omid-the-great:
# We are fast to release Xray API
- We use github actions to release the Xray API fully automatically as fast as possible
- We can manage many programming languages, now we support `cpp` and `python`


yuhan6665:
Thanks!

PR #3119, opened 2024-03-10
Tempest isn't only for OpenRC

PoneyClairDeLune:


Fangliding:
My mistake, I even reminded myself to change "for" to "support", but I still forgot

yuhan6665:
感谢

PR #3163, opened 2024-03-20
Chore: Clean up legacy `field` usage

chise0713:
These codes seems useless, because it can be replace by using this example:
```jsonc
{
    "domain": "geosite:cn",
    "outboundTag": "direct"
},
{
    "ip": "geoip:cn",
    "outboundTag": "direct"
}
```
I think we can remove these codes without worries of breaking.


yuhan6665:
看起来是一个过时设计 删点应该问题不大。。感谢

PR #3174, opened 2024-03-22
Improve balancer_info API

hossinasaadi:
fix balancer args


yuhan6665:
Thanks!
If you have some time, it would be great to have an example here or put in documentation :)

hossinasaadi:
> Thanks! If you have some time, it would be great to have an example here or put in documentation :)

Sure ❤️

PR #3232, opened 2024-04-05
Fix IPv6 random IP logic error

ImAubrey:
Optimize the random IP algorithm logic and fix the problem of random IP errors when the prefix is not divisible by 8.

### DIFF
--- a/app/proxyman/outbound/handler.go
+++ b/app/proxyman/outbound/handler.go
@@ -2,12 +2,8 @@ package outbound
 
 import (
 	"context"
+	"crypto/rand"
 	"errors"
-	"io"
-	"math/rand"
-	gonet "net"
-	"os"
-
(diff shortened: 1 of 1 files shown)


ImAubrey:
> Very grateful I did think of the situation of the prefix might not divisible by 8 and put it on the schedule, but I forgot it in the end😂

I came across the update in version 1.8.10 by chance and started using it. Yesterday, I had my own IPv6 Prefix, but my announce was /47. At first, when I filled in /47, I thought it was a problem with my network. It was only after examining the code that I realized this minor issue.

Additionally, in my opinion, converting the IP address and prefix directly into a range, then randomly generating a number within this range, and finally converting this number back to an IP address might be more efficient.

Ha, no worries at all!  Glad I could help!

yuhan6665:
looks great!

PR #3262, opened 2024-04-13
chore: remove the usage of some deprecated tls properties

yin1999:
remove the usage of some deprecated tls properties.

The following tls properties are deprecated:

- [`PreferServerCipherSuites`](https://pkg.go.dev/crypto/tls@go1.22#:~:text=PreferServerCipherSuites%20bool): This property is ignored in Go 1.17: golang/go#45430.
- [`NegotiatedProtocolIsMutual`](https://pkg.go.dev/crypto/tls@go1.22#:~:text=NegotiatedProtocolIsMutual%20bool): This property always returns `true`, see: https://github.com/golang/go/blob/519f6a00e4dabb871eadaefc8ac295c09fd9b56f/src/crypto/tls/conn.go#L1611


yuhan6665:
感谢！

PR #3285, opened 2024-04-18
fix [ log ]: small fix for better readability.

X-Oracle:


Fangliding:
btw, did you found the problem

X-Oracle:
> btw, did you found the problem

fetch Api in Cloudflare worker does not allow changing host name.

I need to have one domain and set host to match that domain.
so, I can't use `Host: "SUBDOMAIN.worker.dev"`

I figured it out with recent log commit.

right now. my workaround is to not set host.

PR #3405, opened 2024-05-29
refactor: separate the logic of `inbound fallback` so that it can be reused

Grvzard:
works fine in deployment


Fangliding:
Can you provide a detailed explanation of the changes?

yuhan6665:
Couple of thought:
1. Write Everything Twice, then de-duplicate if more than three times
2. Trojan is sunsetting, not worth much refactor
So I think current code is ok, thanks for your effort though

RPRX:
如 @yuhan6665 所言，这部分代码其实没必要动，如果觉得代码重复我们可以删掉 Trojan 的 fallbacks 支持，但目前来说没必要动

Grvzard:
>     1. Write Everything Twice, then de-duplicate if more than three times
Totally agree.
>       I think current code is ok
I think xray-core is a tool for creating new protocols, so these changes are made for my convenience.
It's okay if the upstream doesn't need them.

RPRX:
总之还是感谢 PR

PR #3413, opened 2024-06-03
chore: Remove unnecessary `if`

AsenHu:


Fangliding:
How does this lead to leaking? It seems to cause some loops (if the client keeps sending request), but other situations seem to be the same, such as incorrect host

AsenHu:
> How does this lead to leaking?

This was my mistake. Because it's impossible for `if` to be true, it won't lead to leaking.

Fangliding:
So keep it consistent with websocket

PR #3424, opened 2024-06-05
CI: not create zip file when not releasing

AsenHu:
Makefile: remove unneccesary `install` target &

Add build to `.PHONY` target


Fangliding:
If you have more chore to do, please add them all into this PR, I don't want to make too much git history

yuhan6665:
If no other thoughts we can merge and give it a try..

PR #3428, opened 2024-06-06
drain buffer correctly in httpupgrade

mmmray:
it seems the recently added httupgrade testsuite is causing timeouts on master

i have no evidence this is the real issue, but it feels to me that the server could accidentally over-read, and then the encapsulated connection will block forever trying to read data

let's test it in CI a couple of times, i don't have a way to reproduce the issue


yuhan6665:
Looks good to me! The KCP issue is long-standing. I think UDP traffic test has some little chance of failing.

PR #3430, opened 2024-06-06
HTTPUpgrade send headers with specified capitalization

Fangliding:
昨晚上看到就知道怎么搞了 感觉没必要就没动 没想到第二天起来看到还写了个这么复杂的玩意的
个人觉得还是有一点点必要维持可读性的（）

### DIFF
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -208,8 +208,10 @@ func (c *HttpUpgradeConfig) Build() (proto.Message, error) {
 	// Host priority: Host field > headers field > address.
 	if c.Host == "" && c.Headers["host"] != "" {
 		c.Host = c.Headers["host"]
+		delete(c.Headers,"host")
 	} else if c.Host == "" && c.Headers["Host"] != "" {
 		c.Host = c.Headers["Host"]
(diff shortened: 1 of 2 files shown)


Fangliding:
好像revert了不该revert的东西 等一下下（）

Fangliding:
> this version of xray will send two host headers Host and host

It has a pre-check, I just need 2 lines to fix it.

>  i can open a PR to restore the state of commit 097990926fb6ccd3c50c2bd798ef856d9208fb7f

No need. I'm just saying there's no need to worry about the capitalization of the header. If there's a simple way to implement it, let's just keep it

mikeesierrah:
Thanks @Fangliding @mmmray  

also @Fangliding one thing to share : 

IRGFW throttles and blocks all unknown sni's it kinda went MAD in that sense ( Cant understand --> block ) - and you are overestimating abilities of current IRGFW in fingerprinting notls connections 

Still i understand **XRAY is not specific to a single country** its for **everyone** and **you want something that everyone can use** - but its good to be less-fingerprintable with notls

yuhan6665:
Thanks all!

PR #3435, opened 2024-06-07
Chore : change httpupgrade header "upgrade" to "Upgrade"

mikeesierrah:
Since both Chromium and Firefox use `Connection : 'U'pgrade ` in their request header I changed it so it doesn't differ from them and be same as the header in websocket transport 
image
image


Fangliding:
There is a little more work, I will complete it and squash it to #3430 for simplicity

PR #3465, opened 2024-06-21
splithttp: Add support for H2C and http/1.1 ALPN on server

mmmray:
Partially address https://github.com/XTLS/Xray-core/issues/3464

H2C support is required so that fallbacks can work with splithttp at all. **fallback with `path`-based routing still doesn't work.**

PR #3466, opened 2024-06-21
Make http1.1 ALPN work on SplitHTTP client

mmmray:
Users that encounter protocol version issues will likely try to set the ALPN explicitly. In that case we should simply grant their wish, because the intent is obvious.


Fangliding:
I think you can add this change in #3465 instead of opening two pull requests

PR #3468, opened 2024-06-22
Make browserdialer faster

mmmray:
On my machine, browserdialer speedtest gives 10 Mbit/s download. Now it can do 300 Mbit/s.


Fangliding:
looks fine

PR #3469, opened 2024-06-23
More docker tags

mmmray:
push tags for:

* every SHA on master, format `sha-<shortid>`
* every release, format `1.8.16`

right now it is hard to pin a version in my infrastructure, because basically nothing is tagged.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,6 +1,8 @@
 name: Build docker image
 
 on:
+  release:
+    types: [published]
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


yuhan6665:
Thanks we can give it a try

yuhan6665:
> @yuhan6665 可以删除这行吗，如果不使用volume，每次新建容器都会生成一个volume，时间久了会累积很多垃圾文件
> 
> https://github.com/XTLS/Xray-core/blob/8deb953aec3c194300150bb57d858819ed2c9966/.github/docker/Dockerfile#L19

感觉是小问题 手工清下吧

PR #3485, opened 2024-06-30
Fix connection reuse in splithttp HTTP/1.1

mmmray:
Fix two bugs:

* When the connection pool contains more than 5 dead connections, the write fails after 5 tries and is aborted. Instead, pull infinite amount of connections from the pool and when the pool is empty, fail the write on the first fresh connection.

* When the write is retried, the request body is already drained, and the upload request will contain no payload at all. Fix this by stringifying the request eagerly.

Also explicitly set content-length so that transfer-encoding: chunked is avoided. This really doesn't seem to matter to any CDN at all though.

In a future version, there should be an explicit mechanism to read responses asynchronously and retry. This will certainly increase reliability further. ~However, if you actually care about reliability you should use h2~

### DIFF
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -1,6 +1,7 @@
 package splithttp
 
 import (
+	"bytes"
 	"context"
 	gotls "crypto/tls"
 	"io"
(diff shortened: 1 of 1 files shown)


RPRX:
这样的 bug fix 准备好后，你可以直接 Squash and merge

PR #3567, opened 2024-07-20
Fix serverside TLS support of SplitHTTP H1/H2

### LINKED ISSUE #3566: alpn http/1.1 doesn't work !
https://github.com/XTLS/Xray-core/releases/tag/v1.8.20

SplitHTTP 服务端支持 H2C，客户端支持 HTTP/1.1+TLS
SplitHTTP server supports H2C, client supports HTTP/1.1+TLS

But it doesn't work when alpn is http/1.1 !

mmmray:
Fix #3566 

Also update testsuite so that all tests read and write some data. Opening a connection is not enough to trigger connection errors, because the connection is so lazy.

### DIFF
--- a/transport/internet/splithttp/splithttp_test.go
+++ b/transport/internet/splithttp/splithttp_test.go
@@ -2,6 +2,7 @@ package splithttp_test
 
 import (
 	"context"
+	"crypto/rand"
 	gotls "crypto/tls"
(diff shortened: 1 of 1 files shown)


RPRX:
Ready 时你直接合了吧，记得 Squash

Issue #3578, opened 2024-07-22
Please add the possibility of using http/1.1 with fingerprint to splithttp


internetfreedomir:
Thanks to all the people who made splithttp.

This protocol is almost complete, but the only problem is that I cannot use http/1.1 + tls with fingerprint ! 
With the ws protocol, the connection is forced to use http/1.1 and it also works with fingerprint ! 

Is it possible to add this to splithttp as well ?

RPRX:
不建议用 HTTP/1.1，除非服务端不支持 H2，为什么服务端提供了 H2 而你却只想用 HTTP/1.1？

internetfreedomir:
> 不建议用 HTTP/1.1，除非服务端不支持 H2，为什么服务端提供了 H2 而你却只想用 HTTP/1.1？

There are 3 important reasons 

1- Iran applies restrictions based on alpn, for example, sometimes grpc connects well and sometimes ws, and now splithttp with h3 and this shows that there are restrictions based on alpn (of course, these restrictions are not fixed and almost It is changed every 1 to 2 months) 

2- We need a complete protocol that can be connected under any conditions! 

3- Some Iranian CDNs support h2 and http/1.1, but splithttp only connects with http/1.1! (h2 doesn't connect with fingerprint or without fingerprint, but http/1.1 connects without fingerprint)

I don't know the structure of china gfw, but I know very well that Iran's filtering system is not static and is constantly changing.

internetfreedomir:
If we want a protocol that cannot be limited by alpn in any way, http/1.1 must also be connected (with fingerprint).

internetfreedomir:
> * I have serious doubts that you are going to get the performance you want if HTTP/1.1 is actually used, regardless of ALPN.
> * Setting ALPN manually compromises the fingerprint itself. Do you think pretending to be Chrome but only sending `http/1.1` is reasonable? Why do you want to camouflage as websocket connection?
> * Please be specific in what you are seeing. Sometimes this protocol is blocked, sometimes the other one, but there are many ways to distinguish protocols, why do you think it is ALPN?
> * As always, PRs are welcome but features like this are not prioritized for the sake of being "complete."

If we consider what you say, then ws and grpc protocol are completely distinguishable because ws = http/1.1 and grpc = h2, and this makes vpn appear based on alpn ! 
So the first xray protocol that cannot be recognized by alpn is splithttp ?
With this setting :
      "alpn": [
            "h3",
            "h2",
            "http/1.1"
[...]

PR #3583, opened 2024-07-22
SplitHTTP Client: Decouple protocol configuration from TLS ALPN

ll11l1lIllIl1lll:
如果这功能没用或我写得太烂就请直接 Close 吧

因为好像有些人说 TLS ALPN 决定真实使用的协议有点奇怪，并且好像如果出于一些原因想使用 http/1.1 要设置 ALPN 仅 http/1.1 导致可能破坏指纹（虽然说 Chrome 建一个新 Websocket 也是只有 http/1.1 ，但是使 SplitHTTP Client 更像常见 HTTP 请求也没有坏处）。所以我尝试了一下修改。

~~[哦对了， quic-go 不能自定义 ALPN，她会覆盖 ALPN 。](https://github.com/quic-go/quic-go/blob/e1790485266d71fa3efab679c82965428dedb969/http3/roundtrip.go#L291-L292)~~
哦，是前面覆盖的，在 dial 再覆盖一下就好了。现在 quic-go 请求也可以自定义 ALPN 了，然后就顺便保留了用户在使用 h3 时不发送 SNI 的能力。

添加了 splithttpSettings.protocol 配置项，其为一个字符串，可以为 "h2" 或 "h3"，其他字符串则为使用 http/1.1 ，为空或不设置则从 ALPN 推测（和原逻辑相同）。然后我发现好像使用 httpVersion float 接受 1.1, 2, 3 可能更合适


Fangliding:
这是我刚刚想说的(都准备敲字了)
不过还有一件事是监听可能也需要去定义 要不要听h3 是用这个选项还是新建一个字段(yysy ray把请求和监听都用一个streamsetting确实是烂设计) 可能会有两种情况 都听/听tcp/听udp 光一个str好像没法很好表示这俩种情况

internetfreedomir:
This structure is really confusing, but according to what I read in xtls articles, **reality** is always h2, **ws** is always http/1.1, **grpc** is always h2, **h2** protocol is always h2, and **quic** is always h3. 
Because of this, reality can be created with grpc and h2, but not with ws and quic.

So, according to this, in none of these protocols, alpn doesn't have a special function 🤔 

So alpn is either for **tcp** protocol or it is an unrealistic parameter !

Of course, with **splithttp**, the alpn is important and causes changes in the connection type.

These factors make me think that filtering or restricting is possible with alpn !

internetfreedomir:
I thought it might be possible to make splithttp like ws only use http/1.1 with fingerprint.
**Adding a parameter is a wise thing to do !**
This makes it work better with some unpopular CDNs !

RPRX:
> my (really shaky) understanding of the ALPN logic is:
> 
>     1. if it is websocket, the ALPN is always h1
> 
>     2. if it is set via fingerprint, the ALPN is from the fingerprint
> 
>     3. otherwise, it is taken from the alpn setting
> 
> 
> please correct me if this is wrong.

逻辑正确，不过换个表述更好：

Xray 可以手动修改 ALPN，除非使用了浏览器指纹，但若使用了 WebSocket，ALPN 始终只有 `http/1.1`

> and more settings may add more confusion.

赞同，我觉得暂时没有必要增加复杂度

> 因为好像有些人说 TLS ALPN 决定真实使用的协议有点奇怪，并且好像如果出于一些原因想使用 http/1.1 要设置 ALPN 仅 http/1.1 导致可能破坏指纹

~~由 TLS 层决定内层协议是有点怪，虽然这就是 ALPN 的本意，虽然我们对 alpn 选项的用法可能并不是用于协商而是直接决定~~

不会导致破坏指纹，因为 Xray 在使用浏览器指纹时不允许改 Client Hello 的 ALPN，~~此时 alpn 选项刚好可以发挥其它作用~~

RPRX:
@mmmray 即使这种 edgecase 存在，他需要的也是修改浏览器指纹 Client Hello 的 ALPN，与这个 PR 的目标不同

RPRX:
> Xray 可以手动修改 ALPN，除非使用了浏览器指纹，但若使用了 WebSocket，ALPN 始终只有 `http/1.1`

@Fangliding 这句最近重复太多遍了，应该加进文档

internetfreedomir:
> I think what must have happened is, somebody has a CDN that advertises h2 support and negotiates h2 protocol, but the actual h2 implementation is broken in some way and must be avoided. In this case it's not possible to use ALPN to downgrade to h1, because due to `fp`, h2 is negotiated and the CDN may refuse the connection if the client starts sending h1 afterwards.
> 
> But it's such an edgecase, I don't know if it exists, but it's my best reading of #3578 -- should be confirmed

I'm not sure whether these CDNs are broken or not, but it probably applies restrictions that make only splithttp have an unstable connection, although this unstable connection is only for h2 not http/1.1. While normal download and upload links work with h2 without any problem or error. 

And as I said before, splithttp also works with http/1.1 when there is no fingerprint, and my request to xray team was to make http/1.1 exactly like ws so that it works with fingerprint! 

This is not related to the current topic, but I think it is important:
But it is also important that if alpn for ws is only http/1.1 and grpc is only h2, then they can be recognized based on alpn, because I don't think it is only http/1.1 or only h2 in a browser like Google Chrome.

PR #3600, opened 2024-07-26
Remove unnecessary sleep from test

mmmray:
These were copypasted to splithttp tests and then removed there

I can't imagine what they were supposed to do, the tests run fine without


Fangliding:
It appears to be a legacy part, where wstest used gofunc and returned a failure after 5 seconds(timeout)
Later, this mechanism was removed, but this line of code was retained, so it looks confusing

PR #3610, opened 2024-07-29
SplitHTTP: Fix wrong config deserialization

mmmray:
Manual testing was conducted only using explicitly set parameters. However, when the parameters are not explicitly set in JSON config, it seems that `c.MaxUploadSize` will contain `RandRangeConfig { From: 0, To: 0 }` instead of `nil`, which breaks upload entirely.

I now tested all combinations and can confirm that h1, h2, h3 work fine against both new and old servers.

### DIFF
--- a/transport/internet/splithttp/config.go
+++ b/transport/internet/splithttp/config.go
@@ -37,7 +37,7 @@ func (c *Config) GetRequestHeader() http.Header {
 }
 
 func (c *Config) GetNormalizedMaxConcurrentUploads(isServer bool) RandRangeConfig {
-	if c.MaxConcurrentUploads == nil {
+	if c.MaxConcurrentUploads == nil || c.MaxConcurrentUploads.To == 0 {
 		if isServer {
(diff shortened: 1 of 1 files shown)


RPRX:
想把 maxUploadSize 改名为 maxUploadBytes，把服务端默认值也设为 1MB，把服务端 maxConcurrentUploads 默认值也设为 100

RPRX:
改成 maxEachUploadBytes

Fangliding:
Maybe like this?
```
MaxConcurrentUploads *Int32Range        `json:"maxConcurrentUploads"`
```
```
if c.MaxConcurrentUploads != nil {
...
}
```

RPRX:
> @Fangliding you're right, this seems to just be a case of wrong types in the json struct... since this works for now, i will fix it properly later, once the dust settles. thanks

没问题的话我就先发 v1.8.23 了，~~整理代码的工作可以放到后面，我还想把那三个字段的顺序改一下~~

PR #3613, opened 2024-07-29
SplitHTTP client: Add xmux (multiplex controller) for H3 & H2

ll11l1lIllIl1lll:
我不确定我有没有正确理解其工作原理，并且我写完之后才发现上游已经更新了一堆提交，可能我合并的时候也哪里改炸了。直觉告诉我这可能有资源泄露问题，至少好像能用，所以我提交了。
这里假定 http.Client 会对每个请求复用连接，我实际使用的时候也看起来像是对请求复用连接。所以 idk.

这个代码其实挺烂的，我稍后再改，或者还是等另一个人去以更好的方式实现吧，至少肯定不是我。

注：好像 Interval 更倾向于定时任务的间隔，如果是硬限制最小间隔可能 Delay 更符合表达习惯。并且可能那 3 个 sc 参数也应该移到 mux


yuhan6665:
感谢 PR
虽然目前只有 SH 能用 但是考虑一下我们要不要把选项并入 https://xtls.github.io/config/outbound.html#muxobject

Fangliding:
em 毕竟原理不一样 这个是复用client而不是mux.cool 所以放这还算河里吧

RPRX:
Good job！~~这下又不用自己写代码了~~，晚点我 review 一下

dantesieg:
Using xray version including these commits causes crash after a while when using on client, SplitHTTP with h3 alpn. If needed i can send my configs, i used new mux mechanism with h3. Server uses same version taken from Github Actions. I don't know if this helps. I just enabled new mux and tried "prefer_new", "prefer_reuse" and "prefer_existing", didn't change other options.

```
panic: connection already exists

goroutine 400 [running]:
github.com/quic-go/quic-go.(*connMultiplexer).AddConn(0x400007d0e0, {0x75a9ec4f40?, 0x40000702f8?})
	github.com/quic-go/quic-go@v0.45.1/multiplexer.go:59 +0x198
github.com/quic-go/quic-go.(*Transport).init.func1()
	github.com/quic-go/quic-go@v0.45.1/transport.go:266 +0x3b0
sync.(*Once).doSlow(0xb?, 0x1?)
	sync/once.go:74 +0x100
sync.(*Once).Do(...)
	sync/once.go:65
github.com/quic-go/quic-go.(*Transport).init(0x4000d18700, 0x78?)
	github.com/quic-go/quic-go@v0.45.1/transport.go:225 +0x58
github.com/quic-go/quic-go.(*Transport).dial(0x4000d18700, {0x5db44f6450, 0x40021bdea0}, {0x5db44ed5b0, 0x4000d7d5f0}, {0x0, 0x0}, 0x4000d109c0, 0x4000d1ab40, 0x1)
	github.com/quic-go/quic-go@v0.45.1/transport.go:212 +0x70
github.com/quic-go/quic-go.(*Transport).DialEarly(...)
	github.com/quic-go/quic-go@v0.45.1/transport.go:204
github.com/quic-go/quic-go.DialEarly({0x5db44f6450, 0x40021bdea0}, {0x5db44fc670?, 0x40000702f8?}, {0x5db44ed5b0, 0x4000d7d5f0}, 0x4000d109c0, 0x4000d1ab40)
	github.com/quic-go/quic-go@v0.45.1/client.go:95 +0xfc
github.com/xtls/xray-core/transport/internet/splithttp.createHTTPClient.func2({0x5db44f6450, 0x40021bdea0}, {0x5db3d00321?, 0x5bd7?}, 0x4000d109c0, 0x4000d1ab40)
	github.com/xtls/xray-core/transport/internet/splithttp/dialer.go:130 +0x1b4
github.com/quic-go/quic-go/http3.(*RoundTripper).dial(0x4000251e30, {0x5db44f6450, 0x40021bdea0}, {0x4003602db0, 0x10})
	github.com/quic-go/quic-go@v0.45.1/http3/roundtrip.go:312 +0x224
github.com/quic-go/quic-go/http3.(*RoundTripper).getClient.func1()
	github.com/quic-go/quic-go@v0.45.1/http3/roundtrip.go:249 +0x7c
created by github.com/quic-go/quic-go/http3.(*RoundTripper).getClient in goroutine 397
	github.com/quic-go/quic-go@v0.45.1/http3/roundtrip.go:246 +0x258
```

LearZhou:
我这边大致的测试中：prefer_new对h2更友好， prefer_reuse对h3更友好

RPRX:
等 https://github.com/XTLS/Xray-core/pull/3624 和 https://github.com/XTLS/Xray-core/pull/3643 合并后 rebase 一下我再看看吧

RPRX:
@ll11l1lIllIl1lll 可以 rebase 一下了，基于 main

RPRX:
先改了下标题，晚点我比照着 https://github.com/XTLS/Xray-core/issues/3560#issuecomment-2247495778 看下代码

RPRX:
首先 mode 不需要那么多别名，其次基础模式是二选一不是共存，prefer_reuse 就对应 concurrency，prefer_new 就对应另一个

好像没看到限制“一个连接最多被累计复用多少次”的代码

https://github.com/XTLS/Xray-core/issues/3560#issuecomment-2247495778 下面说的 send rate 和 number of bytes sent/rcvd 等先不加，~~只是提一下，毕竟 PR 里也没有~~

RPRX:
> general comment: i think there are already too many things called mux. most are related to the v2ray universe. there is some dangerous overlap with http with "h2mux" in sing-box. this _is_ a mux but the rfcs don't call it that. can it be called "connection pool settings" or something else?

可以叫 XMC：Xray Multiplex Controller，因为这套机制完善后还要加给 Xray 的 Mux、gRPC、H2 等，~~虽然可能会把 H2 删掉~~

RPRX:
@Fangliding xPaddingBytes 的文档

RPRX:
@Fangliding 以及 "Cache-Control: no-store"

mmmray:
it's to awkward to update the document when the release is not out. starting some updates here https://github.com/XTLS/Xray-docs-next/pull/558

Fangliding:
> @Fangliding 以及 "Cache-Control: no-store"

#3562 的改动是默认的 没有config选项((((

RPRX:
To konsclufka：~~出来干活了~~

RPRX:
@mmmray  ~~又要 rebase 了~~

mmmray:
I'll take over this PR. I might add some more things if there is time. I hope to have it ready by sunday.

@APT-ZERO The answer is, yes can be added to grpc and a bunch of others, but it's not planned right now, and it can bypass speed limits like that, but I think if everybody uses it it will become meaningless (especially if you think of Irancell). That's all I'll say here. I suggest to go to github discussions or telegram for general discussion, I think this PR thread is already way too long and hard to navigate ~~but i'm often the first one to complain about such things~~

RPRX:
@mmmray 很高兴你接手了这个 PR，~~我觉得既然伊朗人等着新版~~，可以先发个新版看看 main 分支截至目前积累的 changes 有没有bug，~~而且虽然新版一直在等这个 PR 但这个 PR 也放新版的话这次的东西就太多了~~，所以留给你的时间比较充足

mmmray:
I think this satisfies all requirements from https://github.com/XTLS/Xray-core/issues/3560#issuecomment-2247495778

The mux manager now works at `http.RoundTripper` level, and muxes individual HTTP requests instead of virtual connections.

I think this is better for reusability across transports. It was said somewhere that it should eventually be ported to grpc transport, but it seems that grpc-go actually makes it difficult to use a custom RoundTripper, so it may not happen so soon after all.

The defaults are as follows:

```jsonc
"splithttpSettings": {
	"httpMux": {
        "concurrency": 0, // infinite concurrency per physical connection
        "connections": 0, // do not open any connection "eagerly"
        "connectionLifetimeMs": 0, // permit any connection to survive for "infinity" milliseconds
        "requestsPerConnection": 0  // allow "infinity" requests on one connection before throwing it away
	}
}
```

Those defaults is equal to the current behavior, which is to "mux everything" in a maximally aggressive way. Ironically, when `httpMux` key is removed, it is actually enabled in the most aggressive way, so it works in the opposite way of mux.cool.

I think the last two parameters are self-explanatory. Let's talk about the modes.

The two modes "max connections" and "max concurrency" are not encoded as modes at all. Instead, you set one parameter or the other:

> 同时最多多少子连接（concurrency），先把现有连接复用满再开新的

```jsonc
"splithttpSettings": {
	"httpMux": {
        "concurrency": 8 // each outer connection holds at most 8 concurrent HTTP requests, open more connections as needed
	}
}
```

This should be familar to mux.cool users.

There is a caveat: `concurrency` limits the amount of concurrent HTTP requests, not the amount of concurrent "splithttp virtual connections". SplitHTTP itself has a lot of concurrent requests, so `concurrency=1` actually achieves an "un-muxing" effect and will cause dozens of TCP/UDP connections to spawn. I think this will be interesting to some people who want to cheat QoS. Otherwise, a more conservative value would be much higher and aligned with `scMaxConcurrentPosts`.

> 同时最多多少条 TCP/UDP “连接”，先开新连接直至占满总数再复用

```jsonc
"splithttpSettings": {
	"httpMux": {
        "connections": 8 // open connections until there are 8 of them, and then reuse them
	}
}
```

When there are many open connections, each HTTP request will be sent through a random connection in the pool. (I have not verified this part with any test or wireshark)

Both `connections` and `concurrency` can be set at the same time:

```jsonc
"splithttpSettings": {
	"httpMux": {
        "connections": 8,
        "concurrency": 16
	}
}
```

In this case, the mux manager will open 8 connections first, then open more connections if the concurrency limit has been hit on all connections.

mmmray:
Right now this don't apply to http/1.1 at all. http/1.1 doesn't have mux, but actually it has request pipelining, so maybe these settings should still be applied to that functionality somehow.

RPRX:
@mmmray 先发一个 PR 移除与 `ok` 相关的逻辑？

RPRX:
此外麻烦发个 PR 移除 QUIC 和 DomainSocket 传输层（检测到不存在的传输层就报错）

RPRX:
@Fangliding 那个全局 TransportObject 代码里删了吗？要报错

RPRX:
~~还有啥需要删掉的可以趁着升级 v24 一起删了，要不以后趁着升级大版本正好删东西~~

RPRX:
[更新了 v1.8.24 的 release notes](https://github.com/XTLS/Xray-core/releases/tag/v1.8.24)，补充上了代码退出机制，修改了 Changes 的描述

RPRX:
然后我觉得这个 log 还是有很大问题，比如说日常 log 竟然没有 ctx id，Warning 有但不知道对应哪个，还得统一一下格式

mmmray:
@RPRX about this PR, do you have opinions on:

* behavior of http/1.1 (`httpMux` could control pipelining)
* defaults (for example, we could default to `{"connections": 8}`)

otherwise I'll merge it

I don't think there's any dependency relation between this and the stuff you're talking about.

RPRX:
@mmmray 一般来说上行数据较少，下行数据较多，HTTP/1.1 时下行都是独立的一条条 TCP，所以我觉得不用管 HTTP/1.1

~~这个 PR 我还没看，等我看一下~~

mmmray:
> Then I think there are still big problems with this log. For example, the daily log does not have a ctx id. There is a Warning but I don’t know which one it corresponds to. The format needs to be unified. 

I didn't get it, is it relating to splithttp or the general approach to logging? "daily log", didn't understand that at all :see_no_evil:

RPRX:
> > Then I think there are still big problems with this log. For example, the daily log does not have a ctx id. There is a Warning but I don’t know which one it corresponds to. The format needs to be unified.
> 
> I didn't get it, is it relating to splithttp or the general approach to logging? "daily log", didn't understand that at all 🙈 ~gpt4o doesn't provide a better translation either~

与 SplitHTTP 关系不大，daily log 指的是 Socks 入站的 log

RPRX:
> `concurrency` limits the amount of concurrent HTTP requests

这个不符合预想，我想的是 limits the amount of sub-connections

RPRX:
> sub-connections

一条被代理连接就是一条 sub-connection

mmmray:
> This is not what I expected. What I thought was limits the amount of sub-connections. 

I wanted the multiplexing to happen in such a way that multiple HTTP requests of the same sub-connection can be spread out over multiple physical connection, and mixed up with other subconnections. Therefore, all the mux-ing is based on HTTP requests, not subconnections. The mux layer does not see sub-connections at all.

But now the concurrency parameter is indeed difficult to use. Maybe it should only count GET requests, but not POST requests, what do you think?

Nimanoghravi (approved):

RPRX:
~~这个 APT-ZERO 咋被折叠了，其实我是觉得他已经没有以前那么离谱了~~

> Isn't better to accept range for connectionLifetimeMs?

https://github.com/XTLS/Xray-core/issues/3560#issuecomment-2247495778 已经说过“以上选项均填范围”

> And with connectionLifetimeMs, what will happen to live connections that exist on the old connections, will it force close old connections and cause continuously disconnection of SSH or other alive connection?

没有 force-close，只是不接受新的 sub-connection 了，是 gracefully-close

mmmray:
@APT-ZERO 

> Also it's important to think about effect of this new Mux on GFW and server's hardware resources

Before this PR, everything was maximally muxed. Technically, this PR doesn't introduce mux, but provides options to "un-mux." So this concern of virtual connection limits is not new, and this PR doesn't change anything about that.

And in fact, it's something that has been discussed in the original PR that introduced splithttp. Sub-connections' lifetime are bound by the lifetime of the GET stream. If the GET is closed, the sub-connection is closed. And the entire HTTP stack of CDN/nginx already has limits on the number of open streams, therefore the server is protected.

> And what will happen if Max connections is 4, and there is already 4 connections waiting for a 'graceful end' of their sub-connections?

The connection will not be used for further sub-connections until at least one of them is closed.

@RPRX 

> It wasn't affected by the neighbor, right?

Maybe, I don't know. It's more about some recent anti-censorship ideas. From the start, there was also feedback that the added connection reuse makes perf worse, and removing it entirely or switching to http/1.1 helps with upload performance on some ISP. So I thought, this PR is a good opportunity to fix all of this. 

Anyway, your concern about connection stability is completely valid. I will get back to this. I'll probably do exactly as you say, although it hurts that single-connection performance cannot be improved further. I guess HTTP-level retries won't work.

----

Otherwise, I suggest (to everybody) to review and test the code yourself. Not my description of it, but the actual program. I think feedback like [this](https://github.com/XTLS/Xray-core/pull/3613#issuecomment-2264523468) is good. I feel that a lot of discussions on PRs are abstracted away from the code that is being reviewed, lots of unnecessary discussions are started, ~~and also things land in master that nobody explicitly agreed on~~

RPRX:
> It's more about anti-censorship

这里的目标是增加基本的 multiplex controller，而 anti-censorship 可以是下行也分包并分散到多个 TCP 连接上，对应多个 CDN IP

mmmray:
The PR has been updated, and my big "documentation" comment as well. I think it's good to review now.

The mux manager no longer works at round-tripper level, but actually muxes `DialerClient`, so basically the same design as originally done by @ll11l1lIllIl1lll. The main difference is that the mux manager is more generic and implements "concurrency" correctly.

RPRX:
改个名，既然 `mux` 是给原本没 mux 的东西加 mux，这个可以叫 `demux`，即 @mmmray 所说的 "un-mux"

MaxUses 改名 ConnectionReuseTimes

mmmray:
done, but the "demux" is too strange and different from mux.cool settings.

In my mind "adding mux to config" means "configuring mux explicitly", not "adding mux". What about changing the default value of mux to `concurrency=1` or just some value lower than infinity?

RPRX:
~~遇事不决 X~~，改名 `xmux`，不叫 `HttpDemux` 或 `SplitHTTPMux`，未来这套机制可以 port 给 gRPC，以及 Xray 现有的 mux

mmmray:
bold proposal, mux settings tries to use the "transport-native" mux, then falls back to mux.cool if there is no such thing. prevent people from using mux.cool with h3, which is probably never a good idea (?)

mmmray:
the config parameter is done, some structs are still named specific to splithttp, but it can be fixed in the PR where this is ported to other transports.

RPRX:
SplitHTTPMux->Xmux

Fangliding:
依我看最好别把逻辑搞太复杂 check来check去的 直接在文档警告别启用mux.cool就是了 grpc也有类似的警告 split这个直接叫muxcontroller或者类似的东西（如标题） xmux这个名字留给以后真port到mux

RPRX:
> In this case, the mux manager will open 8 connections first, then open more connections if the concurrency limit has been hit on all connections.

我明天再想一下这个，应该还要重命名

RPRX:
* maxConnections
* maxConcurrency
* cMaxReuseTimes
* cMaxLifeTimeMs

前两个只能二选一，不能出现突破预期的行为，比如 maxConnections 不能被突破，除非受到后两个限制要开一条新连接

RPRX:
前两个只能二选一，均有值则报错，均无值则不限制复用，但后两个限制仍然可以生效

mmmray:
fineee, done. why restrict it though? it feels to me that this combination is fairly intuitive, although useless.

RPRX:
@mmmray json 忘改了，还有字段顺序调整一下

> fineee, done. why restrict it though? it feels to me that this combination is fairly intuitive, although useless.

我看了下之前的代码也没有预连接（不是 minConnections），那么 maxConnections 实际上会变得 meaningless

RPRX:
关于默认值，现在相当于 maxConnections 默认为 1，其它均为 0

或者像 mux 一样改为 maxConcurrency 主导，以及其它两项要不要有默认值，@mmmray ~~你怎么看~~

mmmray:
Actually, the way the current code works, `minConnections` is more correct. It will open connections until the target has been reached. But not "pre": instead everytime a sub-connection is opened, a new connection is opened.

so i think 0 and 1 are mostly logically equivalent.

`concurrency` naming (without min/max) was mostly so that it can be equivalent to mux.cool settings.

I don't really know how one would dominate the other, since they are now mutually exclusive.

mmmray:
Re-reading the comment, I think you're saying that `maxConnections=1` default makes it impossible to set a config like `"mux": {"maxConcurrency": 10}`, but it's not true. You can set this kind of thing fine and it will behave exactly like mux.cool, because maxConnections can still be exceeded through setting maxConcurrency. The only thing I changed in the config is that it's not possible to explicitly set both parameters at the same time.

mmmray:
Then i think the current state is good. maxConnections is broken if maxConcurrency is set to a value greater than 0. The user is not permitted to do that.

The other case is `{"maxConnections": 10, "cMaxReuseTimes": 1}`, then it's possible to exceed the concurrency limit temporarily, but I think that's ok.

RPRX:
终于合了这个 PR，感谢各位的努力！

RPRX:
https://github.com/XTLS/Xray-core/issues/3823#issuecomment-2354497106

av1934413 (approved):

yasinshirashiyani (approved):

PR #3618, opened 2024-07-30
Suggest adding One click easyXray project to README.md

1andrevich:
Suggest easyXray (https://github.com/Team418-git/easyXray) to 'One Click '


Fangliding:
Thanks, but we don't plan to add more things in this area (Actually, there are even some links that should be deleted)

PR #3643, opened 2024-08-05
SplitHTTP: New padding mechanism

mmmray:
1. "ooook" handling on client is removed entirely, since there is no released version of xray-server that actually writes variable-length "ok"

   This partially reverts commit 4b7947cca91cacd63262711210e96c2d444bc7b5 (https://github.com/XTLS/Xray-core/pull/3614) -- servers that are on that commit from main are **incompatible** with clients on this PR

2. X-Padding header is added to every request and response (in fact, all of them have length characteristics if the transported protocol has length characteristics already. it can be reduced once VLESS adds its own padding, but I thought it's best to stay independent of VLESS and also handle other inner layers)

3. The client reads two bytes, and if those are != "ok", forwards them to VLESS/Trojan. This should be acceptable because VLESS/Trojan headers are larger than two bytes.

I think this is the simplest way to deal with this. If headers are used for communication, then browser dialer has to be adjusted as well (may be done in later PR)


ll11l1lIllIl1lll:
Whether adding padding to all traffic exceeds the responsibility of the transport layer is debatable, as that could arguably be handled by the protocol itself. Padding at the transport layer might only be necessary during the handshake phase.

ahmad0489:
I think it is better to add a new parameter called `paddingtype` with `ok` and `header` values. 
This parameter should be set on the server side.
I feel ok padding is better to use with cdn.

ahmad0489:
> @ahmad0489 please take a look at the preceding discussion here: [#3614 (comment)](https://github.com/XTLS/Xray-core/pull/3614#issuecomment-2264385252) -- this decision to phase out `ok` was already made there as it causes various problems regarding compatibility between client and server, and because it does not solve the problem completely. This sentence:
> 
> > I feel ok padding is better to use with cdn
> 
> needs much stronger justification. If you cannot provide this justification publicly please send me an email (see profile). We're not going to add multiple forms of padding without any. Everybody has opinions.

Yes, you're right, I didn't know you were going to phase out `ok`.
Based on the tests I did, I feel that ok padding is a little better.
Maybe this small difference is due to my unstable internet, that's why I said "I feel".

ahmad0489:
> better than this PR or better than no padding at all? the latter I believe immediately, but this PR should be completely equivalent (for as long as you use TLS, but unless specifically targeted, probably without as well)

Ok padding is better than this PR and better than no padding.
Of course, these tests I did aren't enough and more people should test, that's why I requested to add `paddingtype` to be tested in both modes. 
But if you have to phase out "ok" due to incompatibility, it is better to phase out it.

RPRX:
我觉得个别 CDN 早晚会通过简单地禁止 X-Padding header 来禁止 SplitHTTP，~~比如国内某 CDN 禁止过转发 WebSocket~~

> However, I had a specific, useful setup where absolutely no headers are forwarded (`Set-Cookie` neither).

你指的是对 CDN 进行特殊的设置使它不转发任何 header 吗？还是反代软件？总之我觉得应当提醒用户避免这样的设置而不是接受这种情况，所以客户端检测有没有 X-Padding 就行了，没有的话就还是原逻辑，~~甚至兼容 ok padding 那个 commit~~

此外想把 `padding` 改名为 `headerPadding`

RPRX:
`xPaddingHeader`

RPRX:
> `xPaddingHeader`

以及支持设置 -1 以不发送这个 header，~~方便调试 CDN 有没有禁止它~~

mmmray:
> I think some CDNs will sooner or later ban SplitHTTP by simply banning the X-Padding header.

I was thinking it's possible but didn't know how likely it is ~~so I tried to ignore it~~. I will re-design this PR. When using Set-Cookie/Cookie, I was not sure how browser dialer will interact. I should actually finish browser dialer in this PR as well so I can see.

Some ideas:

* Randomized header name (both length and content is random)
* `Set-Cookie: PHPSESSID=<randomstring>` (it's too strange...)

RPRX:
@mmmray 可以先用 X-Padding，等有人报告 CDN 禁止了这个 header 再改，此外这个 PR 内 Browser Dialer 也要支持 X-Padding

PoneyClairDeLune:
Encode via `ETag` perhaps?

ll11l1lIllIl1lll:
> Encode via `ETag` perhaps?

The typical length of an ETag is 32 characters or less. A 1000 characters ETag is an infrequent occurrence and the consequence of blocking it would be insignificant. Nevertheless, it is not uncommon for cookie lengths of 1000 or even 2000 characters to be encountered (for example, on dash.cloudflare.com or youtube.com, based on personal usage). Alternatively, adding superfluous query string (or pathname) parameters may be considered.

If the path, headers and content are padded to a length of 500-character each, the resulting traffic will have a total length of 1500-character padding. (although this seems unnecessary).

RPRX:
Browser Dialer 支持？

RPRX:
> Maybe I don't understand the issue.

我的意思是，CDN 应该会转发 header，除非进行了特殊设置，比如：

> I had a specific, useful setup where absolutely no headers are forwarded

我的意思是应当避免这样的特殊设置，不过不用纠结这个了，我觉得 `ok` 的逻辑不与 `X-Padding` 挂钩也行

PoneyClairDeLune:
> Nevertheless, it is not uncommon for cookie lengths of 1000 or even 2000 characters to be encountered (for example, on dash.cloudflare.com or youtube.com, based on personal usage)

I'm all for using cookies for extensible configurations.

Might have suggested `Authorization` if not for the TSID proposal I submitted. ~~Kinda soft-bricked myself with the SplitHTTP header PR until that's either merged or rejected.~~

RPRX:
@mmmray 我觉得客户端用 querystring `?X-Padding=` 是可行的，服务端还用 `X-Padding` header，字段名改成 `xPaddingBytes`

RPRX:
对了，昨晚在群里看到了 Nginx 默认 1MB 导致测速上行没速度的问题，@Fangliding 进行更多测试？

Fangliding:
我自己几乎不用nginx的 只是怀疑 毕竟卡1MB鬼知道会不会有MiB和MB之类的小问题或者刚好因为什么原因多一点点导致抛出错误 我叫他改2MB也没回我

RPRX:
~~终于又合并了一个 PR~~，我看了下代码，如果未来把 `ok` 删掉应该不会导致客户端卡住，~~不过得花数个版本才能把 `ok` 彻底删掉~~

反正现在客户端一定兼容任何版本的服务端，或许服务端加个逻辑，querystring 为空才发 `ok`？v1.9 把两端冗余的逻辑都删掉

RPRX:
> I think releasing a version with conditional server response will not accelerate this process, it only adds more steps. Maybe I am missing something.

确实有 missing something，现在服务端的逻辑是一定发 `ok`，如果我们在同一个版本比如 v1.9 把两端关于 `ok` 的逻辑都删掉，会造成 v1.9 客户端接收到上个版本服务端一定会发的 `ok`，所以应当现在就给服务端加个 conditional `ok` 的逻辑，v1.9 才能都删掉

PR #3671, opened 2024-08-10
SplitHTTP server: Only "ok" to older clients

mmmray:
https://github.com/XTLS/Xray-core/pull/3643#issuecomment-2282304185

It makes sense. In earlier discussions of breaking changes, we assumed the server updates much faster than the client. But this is easy to do.

Note: `responseFlusher.Flush()` could also be conditional. It would mean that HTTP response headers are only written when the first VLESS response byte is written. For some CDN this is already the case. I am not sure what is preferred from a padding perspective. I guess technically response X-Padding can be removed if `responseFlusher.Flush()` is removed, and VLESS brings its own padding in the future. Anyway, I don't want to change it right now, because I am not sure if it triggers different timeout settings in the CDN. It can be changed later, it's not a breaking change.

### DIFF
--- a/transport/internet/splithttp/hub.go
+++ b/transport/internet/splithttp/hub.go
@@ -192,12 +192,15 @@ func (h *requestHandler) ServeHTTP(writer http.ResponseWriter, request *http.Req
 		h.config.WriteResponseHeader(writer)
 
 		writer.WriteHeader(http.StatusOK)
-		// in earlier versions, this initial body data was used to immediately
-		// start a 200 OK on all CDN. but xray client since 1.8.16 does not
-		// actually require an immediate 200 OK, but now requires these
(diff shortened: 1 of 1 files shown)


RPRX:
VLESS padding 永远无法取代 GET response 的 header padding，因为不能保证经过 middlebox 后它们仍然黏在一起

RPRX:
话说这个 PR 要不要改成检测到没有任何 querystring 才发 `ok`？

RPRX:
> 话说这个 PR 要不要改成检测到没有任何 querystring 才发 `ok`？

~~想了下还是只检测 x_padding 吧~~

Fangliding:
> (by the way, why is `ed` a querystring? i never understood)

Because /ed1024/xxx or xxx/ed1024 doesn't seem like a good choice
As for why it's not made into a wsSetting, because this way the GUI client doesn't need any changes

PR #3677, opened 2024-08-12
new fragment method: Fragment segmentation

radioactiveAHM:
Okay, this new method is as weird as the previous one. It sends each fragment into separate TCP segments. Imagine you have a fragment with a size of 200 bytes. If you configure it to use 4 segments for each fragment, then this fragment will be sent in 50-byte chunks. I code it in a dynamic way.

Configuration:
```
"fragment": {
	"packets": "tlshello",
	"length": "200/4",
	"interval": "10-20"
}
```
or
```
"fragment": {
	"packets": "tlshello",
	"length": "100-200/3",
	"interval": "10-20"
}
```
the number after `/` determines the number of segments for each fragment.

Result:
Working pretty fine in Iran on both MCI and MTN operators.
Screenshot 2024-08-12 222304
I used `"length": "200/4"`.

The change in `config.pb.go` file is because of `go run ./infra/vprotogen/main.go` command.


Fangliding:
I don't think it's necessary to make it more complicated. If the firewall can reassemble client hello, then no amount of extra work will be of any help, unless there is evidence that it can bring substantial improvement

ImMohammad20000:
> I don't think it's necessary to make it more complicated. If the firewall can reassemble client hello, then no amount of extra work will be of any help, unless there is evidence that it can bring substantial improvement

some ISP in iran can block the fragment method xray-core use right now, this method can bypass this blocking in those ISP

ssmetall:
> @ImMohammad20000 fangliding was referring to #3660 which is not released yet -- does this new PR outperform this previous PR or help with a specific ISP? it does seem slightly more complicated. if not, don't play your cards all at once. and fang is right, if the previous one gets blocked most likely this one will be blocked as well.

this method work perfectly , iam using now , please released xray core

radioactiveAHM:
@mmmray @Fangliding
This method is neither worse nor better than the previous one, but it works fine. The cost of reassembly in this method is significantly higher for GFW. It’s better to have two different methods instead of relying on just one. Users are reporting that this method is working fine.

ssmetall:
> @mmmray @Fangliding This method is neither worse nor better than the previous one, but it works fine. The cost of reassembly in this method is significantly higher for GFW. It’s better to have two different methods instead of relying on just one. Users are reporting that this method is working fine.

yes / method is testing now and work perfectly , 
What is the idea of @RPRX ?

ssmetall:
> > The cost of reassembly in this method is significantly higher for GFW.
> 
> I don't think this is the case at all. TLS record fragment was already implemented through `tlshello`, each TLS record in its own TCP packet. Now the PR in main adds a way to have all TLS records in one TCP packet. This PR adds a middleground to have some TLS records per packet. That's fine but where is the difference in cost for GFW? Either it has TCP reassembly or it doesn't, either it has proper handling of TLS records or it doesn't. If it has both, then both of your PRs will be blocked.
> 
> It would have been better to keep this PR private for a while and release it once the previous PR gets blocked (if it still works), to extend the survival time of these tricks. Releasing a bunch of things like this at once just accelerates the cat-and-mouse game and allows the GFW to block a bunch of things at once (with constant effort).
> 
> @ssmetall You have already said it, once is enough :( we understand that this is working better than the _released_ fragment, but it's not clear how the additional settings help on top of @radioactiveAHM's _unreleased_ PR.

At the current time, the fragment is working in the form of tcp segments without any problems, the more tricks and methods are available for the fragment, the more confused gfw becomes.

radioactiveAHM:
@mmmray @ssmetall You both are right. Most users want this to be merged, but having this method as a backup (private draft) is also a great idea. I’ve done what I could to help, I’ll leave it up to the main developers to decide.

RPRX:
知道为什么 Xray 现有的那些分片方法开始失效了吗，因为 Xray 加了，会引发大量的使用和伊朗 GFW 的关注，以前就 warn 过了

所以我认同 @mmmray 的看法，这些 small tricks 不要一下子都加进 Xray，把存活时间拉长些，所以感谢 PR，但是现在不合适

ssmetall:
> Why are you Chinese so broad and selfish?
> 
> You said you want to build a 1.9.0 kernel, but you haven't yet
> 
> What is the benefit of adding this to the firewall? It's all bad Why you want to continue with your stupid idea " do not spreading the method" there is always a way that we can defeat gfw

It's better not to be demanding. Right now, we are connected and owe it to RPRX. However, I agree with one of your statements: the more connection methods there are, the more GFW gets confused.

ssmetall:
> Look brother I am not saying that it should add special methods to the fragment I don't care if he does or not because Mahsa team will add them later
> 
> My problem is that he shows us Iranians as selfish as themselves and says don't expose your ways, He is wasting all the hard work of our engineers, such as: Segaru Joseph mark Mahsa team and...

Segaro, Joseph and Mahsa are scams, I have information from them that shows that they are collecting data from mahsang , so it is better not to mention their names.

mikeesierrah:
> Why are you Chinese so broad and selfish?
> 
> You said you want to build a 1.9.0 kernel, but you haven't yet
> 
> What is the benefit of adding this to the firewall? It's all bad Why you want to continue with your stupid idea " do not spreading the method" there is always a way that we can defeat gfw

DO NOT MAKE IT PERSONAL OR RACIAL 

they have a point

ssmetall:
> you make me laugh mahsa and mark are scammers??😂 ok fine do what ever you want but if you again insult iranian people,beware of the consequences.

By the way, because I'm Iranian, I'm saying that they're cheating, I'll let them see their screen chats, what kind of dirty people are they, and what are they looking for in Mahsang?

Fangliding:
This place is for discussing develop matters, please go elsewhere for unrelated discussions

PR #3691, opened 2024-08-16
SplitHTTP: Do not produce too large upload

mmmray:
Since https://github.com/XTLS/Xray-core/pull/3603 and https://github.com/XTLS/Xray-core/pull/3611, iperf and speedtest are triggering "too large upload" by the server. This is because v2ray's MultiBuffer pipe can actually return data larger than the configured size limit.

I'm surprised nobody noticed it so far. In principle, any heavy upload could disrupt the entire connection. In its infinite wisdom, speedtest.net hides such errors and only shows low upload instead. I only noticed this issue myself when inspecting server logs.

### DIFF
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -228,7 +228,10 @@ func Dial(ctx context.Context, dest net.Destination, streamSettings *internet.Me
 	httpClient := getHTTPClient(ctx, dest, streamSettings)
 
 	maxUploadSize := scMaxEachPostBytes.roll()
-	uploadPipeReader, uploadPipeWriter := pipe.New(pipe.WithSizeLimit(maxUploadSize))
+	// WithSizeLimit(0) will still allow single bytes to pass, and a lot of
+	// code relies on this behavior. Subtract 1 so that together with
+	// uploadWriter wrapper, exact size limits can be enforced
+	uploadPipeReader, uploadPipeWriter := pipe.New(pipe.WithSizeLimit(maxUploadSize - 1))
 
 	go func() {
 		requestsLimiter := semaphore.New(int(scMaxConcurrentPosts.roll()))
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -227,7 +227,8 @@ func Dial(ctx context.Context, dest net.Destination, streamSettings *internet.Me
 
 	httpClient := getHTTPClient(ctx, dest, streamSettings)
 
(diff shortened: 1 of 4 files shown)


mmmray:
@APT-ZERO the test is just meant to demonstrate that querystrings don't break the transport entirely. it serves no purpose, but the test was incomplete.

Fangliding:
That's why some users report that increasing the maximum upload bytes can improve upload speed significantly
I used to suspect that this was some HTTP issue so client send bigger packet, but I didn't expect the reason to be like this

mmmray:
by the way, i'm struggling with MultiBuffer pipe. it has a strange API due to performance concerns, and a lot of code seems to rely on the behavior of `WithSizeLimits(0)`. It would be more appropriate to use a different buffered pipe but it would mean more dependencies. Maybe I will write another one someday.

mmmray:
also I suppose this is why @MHSanaei added different defaults to splithttp: https://github.com/MHSanaei/3x-ui/commit/d319476eb6e7b9426b2441069d619284e1ab6ac2#diff-b5ef5f7bdd4bc564ca919c23529956b3402c3065543fbe2f773088b6893cb467R559

(most of those options have no real effect on the server, but the bump of MaxEachPostBytes works around this bug)

ahmad0489:
> also I suppose this is why @MHSanaei added different defaults to splithttp: [MHSanaei/3x-ui@d319476#diff-b5ef5f7bdd4bc564ca919c23529956b3402c3065543fbe2f773088b6893cb467R559](https://github.com/MHSanaei/3x-ui/commit/d319476eb6e7b9426b2441069d619284e1ab6ac2#diff-b5ef5f7bdd4bc564ca919c23529956b3402c3065543fbe2f773088b6893cb467R559)
> 
> (most of those options have no real effect on the server, but the bump of MaxEachPostBytes works around this bug)

Because the 3x-ui panel has two types of export. 

1- Normal export, for example, to vmess or vless.

2- Export as json sublink! 

These parameters that are set in the server (panel) are placed in json and given to the user (as a sublink).

RPRX:
~~其实我早就测出了测速上行偶尔有速度偶尔没速度，本来还以为是网络问题，所以在群里看到这个 https://github.com/XTLS/Xray-core/pull/3643#issuecomment-2279928488 挺关注的~~

RPRX:
~~其实我早就测出了测速上行偶尔有速度偶尔没速度，本来还以为是网络问题，所以在群里看到这个 https://github.com/XTLS/Xray-core/pull/3643#issuecomment-2279928488 时挺关注的~~

PR #3698, opened 2024-08-19
Fix: No route when bind_interface

chise0713:


proxy/wireguard/tun_linux.go
```diff
@@ -199,6 +199,7 @@ func createKernelTun(localAddresses []netip.Addr, mtu int, handler promiscuousMo
 
 		r := netlink.NewRule()
 		r.Table, r.Family, r.Src = ipv6TableIndex, unix.AF_INET6, addr.IPNet
+		r.Table, r.Family, r.OifName = ipv6TableIndex, unix.AF_INET6, n
```
mmmray, line 202:
```suggestion
		r.OifName = n
```

those other things are already assigned

proxy/wireguard/tun_linux.go
```diff
@@ -199,6 +199,8 @@ func createKernelTun(localAddresses []netip.Addr, mtu int, handler promiscuousMo
 
 		r := netlink.NewRule()
 		r.Table, r.Family, r.Src = ipv6TableIndex, unix.AF_INET6, addr.IPNet
```
mmmray, line 201: what happens to this rule? I think it is simply overwritten

chise0713: I forget to `append()`, my fault.

chise0713: But my local repository didn't, so maybe yesterday I was too sleepy so I forgot to Ctrl + S 🥲

Issue #3699, opened 2024-08-19
panic with this error


kabo00os:
### Integrity requirements

- [X] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [X] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [X] I searched issues and did not find any similar issues.

### Version

1.8.23

### Description

After some time this error will be displayed

### Reproduction Method

splithttp method

### Client config

<details><pre><code>
-
</code></pre></details>

### Server config

<details><pre><code>
-
</code></pre></details>

### Client log

<details><pre><code>
panic: close of closed channel

goroutine 2667 [running]:
github.com/quic-go/quic-go/http3.(*hijackableBody).requestDone(...)
	github.com/quic-go/quic-go@v0.45.1/http3/body.go:124
github.com/quic-go/quic-go/http3.(*hijackableBody).Close(0x1400018ef30)
	github.com/quic-go/quic-go@v0.45.1/http3/body.go:130 +0x38
github.com/quic-go/quic-go/http3.(*gzipReader).Close(0x140008817f8?)
	github.com/quic-go/quic-go@v0.45.1/http3/gzip_reader.go:38 +0x28
github.com/xtls/xray-core/transport/internet/splithttp.(*LazyReader).Close(0x140007edc80)
	github.com/xtls/xray-core/transport/internet/splithttp/lazy_reader.go:53 +0xc8
github.com/xtls/xray-core/transport/internet/splithttp.(*LazyReader).Close(0x140007edcb0)
	github.com/xtls/xray-core/transport/internet/splithttp/lazy_reader.go:53 +0xc8
github.com/xtls/xray-core/transport/internet/splithttp.(*splitConn).Close(0x1400028bcc0)
	github.com/xtls/xray-core/transport/internet/splithttp/connection.go:26 +0x50
github.com/xtls/xray-core/proxy/vless/outbound.(*Handler).Process(0x140003858c0, {0x10537bd58, 0x140003b1b90}, 0x1400030ecc0, {0x105379008, 0x1400038c300})
	github.com/xtls/xray-core/proxy/vless/outbound/outbound.go:295 +0x1158
github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).Dispatch(0x1400038c300, {0x10537bd58, 0x140003b1b90}, 0x1400030ecc0)
[...]

PR #3710, opened 2024-08-20
SplitHTTP: Fix connection leaks and crashes

mmmray:
There are a few bugs that are fixed here:

1. (all http versions) When the client opens a connection and closes it without any traffic, the Close would lock forever and not appear on the wire at all
2. (all http versions) When the client terminates the connection correctly, the server would not notice and keep the upload queue and related datastructures around until the proxied connection times out.
3. (quic only) closing the connection might crash the client as well, Fix https://github.com/XTLS/Xray-core/issues/3699

Are fixed in the following way:

* Problem 1 is addressed by removing the use of LazyReader. It turns out that while `ok` is being peeked, `Close` is entirely ignored
* Problem 2 is addressed on the server by using the request's context to close the connection
* Problem 3 is addressed by using the request's context to close the connection instead of `Body.Close`.

### DIFF
--- a/transport/internet/splithttp/client.go
+++ b/transport/internet/splithttp/client.go
@@ -49,6 +49,8 @@ func (c *DefaultDialerClient) OpenDownload(ctx context.Context, baseURL string) 
 	var downResponse io.ReadCloser
 	gotDownResponse := done.New()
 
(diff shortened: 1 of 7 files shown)


kabo00os:
This problem occurs with burstObservatory
When several configs are added, sometimes this problem occurs

PR #3715, opened 2024-08-21
Pin protobuf packages

mmmray:
It happened in the past that I ran with the wrong protobuf version installed locally, and apparently there is even still some file wrong in splithttp. Fix this issue once and for all.


yuhan6665:
I think we should treat it like other dependencies and keep promote latest.
And if you are like me, who is lazy to update all files, you can do
`go run ./infra/vprotogen/main.go -pwd .`

mmmray:
I bumped the version to latest and regenerated all the files, but I still think the version should be pinned. Currently the experience for new contributors is broken because the latest protoc-gen-grpc version introduced some breaking changes.

yuhan6665:
Thanks! Let's merge it for now. I may need to change to reference master later..

Issue #116, opened 2020-12-30
panic: Calling IP() on a DomainAddress


maskedeken:
在往buf.Buffer的UDP属性赋值的时候，创建
&net.UDPAddr{
    IP:   addr.IP(),
    Port: int(port.Value()),
}

但这里的addr是从addrParser.ReadAddressPort()而来，这里的addr有可能是域名，如果直接调用addr.IP()，会直接panic

func (domainAddress) IP() net.IP {
	panic("Calling IP() on a DomainAddress.")
}

rprx:
两天前的提交已经修复 https://github.com/XTLS/Xray-core/commit/13ad3fddf6f7ff01fe9541b6c9f2dca6f7f87e4b

（不过最新的 commit 有问题，等下会 force

maskedeken:
怎么解决循环引用的问题？

rprx:
https://github.com/XTLS/Xray-core/commit/6bcac6cb1016295a0722015ec853ece135766cc3

rprx:
> 
> 
> 怎么解决循环引用的问题？

https://github.com/XTLS/Xray-core/commit/6bcac6cb1016295a0722015ec853ece135766cc3

rprx:
如果发现存在其它问题，也请直接反馈

比如目前发现任意门 TPROXY 那里不能缓存 conns，只能发一个包就 close，否则行为会异常，等下打算看看把 dial 改为 listen 的表现

z719893361:
问题修复了吗，我用最新版碰到了相同的问题

somebodywashere:
Also had this a few times on one server. Can you by any chance have any outbound with ForceIP?

I didn't checked the code but I didn't see this error anymore when I removed plain ForceIP from outbound (ForceIPv4v6 & ForceIPv6v4 is ok for me)

somebodywashere:
> i posted my full config above, no I don't use any variant of ForceIP.

Oh, sorry, didn't pay enough attention. Now I see. That's strange than.

sssagsag:
Hi @RPRX  , on core 1.8.23, this log appears and gets disconnected.(go1.22.5 linux/amd64)

XRAY: panic: Calling IP() on a DomainAddress.
XRAY: 0x15626b0?, {0x156c6d0?, 0xc009f7b9c8?})

mmmray:
[...]

PR #3718, opened 2024-08-22
Fix data leak between mux.cool connections

mmmray:
Fix https://github.com/XTLS/Xray-core/issues/116

This issue does not appear at all if mux.cool is disabled. Maybe the issue can be avoided if routing is disabled too, but it's possible that the same issue appears elsewhere.

I have debugged this issue in prod with a lot of print statements, and can confirm that this patch fixes it. ~~Tests should be added. ***The mux server has no tests at all!***~~

Step-by-step:

1. The mux server is calling handleStatusNew with the same context multiple times. That function then calls w.dispatcher.Dispatch

2. Dispatch assigns some values specific to the connection onto the outbound here: https://github.com/XTLS/Xray-core/blob/83eef6bc1f554be84aeb799417688a070cd32ab8/app/dispatcher/default.go#L241-L242

3. Because the outbounds array on the connection object is the same across multiple connections (it's the same context), the value of outbound.Target flips back and forth between values such as "1.1.1.1" or "chat.facebook.com" (IPAddress vs DomainAddress)

4. Many layers deeper in routing, as part of this Dispatch call, this function is called: https://github.com/XTLS/Xray-core/blob/83eef6bc1f554be84aeb799417688a070cd32ab8/features/routing/session/context.go#L53-L55

   It does two things, in order:

   1. Checks if Target is an IPAddress (IsIP)
   2. If it's an IP, call IP()

   Because Target keeps mutating back and forth between multiple kinds
   of address, it sometimes is an IP for Step 1, and a DomainAddress for
   Step 2

5. Calling IP() on a DomainAddress panics:

        panic: Calling IP() on a DomainAddress.

        goroutine 123018 [running]:
        github.com/xtls/xray-core/common/net.domainAddress.IP(...)
            github.com/xtls/xray-core/common/net/address.go:172
        github.com/xtls/xray-core/features/routing/session.(*Context).GetTargetIPs(0xc004175428)
            github.com/xtls/xray-core/features/routing/session/context.go:54 +0x5f
        github.com/xtls/xray-core/app/router.(*MultiGeoIPMatcher).Apply(0xc0003a63e0, {0x156c770?, 0xc004175428?})
            github.com/xtls/xray-core/app/router/condition.go:143 +0x3e
        github.com/xtls/xray-core/app/router.(*ConditionChan).Apply(0x495c9d?, {0x156c770, 0xc004175428})
            github.com/xtls/xray-core/app/router/condition.go:32 +0x5c
        github.com/xtls/xray-core/app/router.(*Rule).Apply(...)
            github.com/xtls/xray-core/app/router/config.go:30
        github.com/xtls/xray-core/app/router.(*Router).pickRouteInternal(0xc000367a40, {0x156c770, 0xc004175428})
            github.com/xtls/xray-core/app/router/router.go:196 +0x176
        github.com/xtls/xray-core/app/router.(*Router).PickRoute(0x1562810?, {0x156c770?, 0xc004175428?})
            github.com/xtls/xray-core/app/router/router.go:84 +0x25
        github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).routedDispatch(0xc000398fc0, {0x1562810, 0xc0050cd350}, 0xc00418dea0, {{0x15625a8, 0xc005974c10}, 0x1bb, 0x2})
            github.com/xtls/xray-core/app/dispatcher/default.go:420 +0x307
        created by github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).Dispatch in goroutine 123015
            github.com/xtls/xray-core/app/dispatcher/default.go:252 +0x56c

### DIFF
--- a/common/mux/server.go
+++ b/common/mux/server.go
@@ -118,6 +118,9 @@ func (w *ServerWorker) handleStatusKeepAlive(meta *FrameMetadata, reader *buf.Bu
 }
 
 func (w *ServerWorker) handleStatusNew(ctx context.Context, meta *FrameMetadata, reader *buf.BufferedReader) error {
+	// deep-clone outbounds because it is going to be mutated concurrently
+	// (Target and OriginalTarget)
+	ctx = session.ContextCloneOutbounds(ctx)
 	errors.LogInfo(ctx, "received request for ", meta.Target)
 	{
 		msg := &log.AccessMessage{
@@ -286,8 +289,7 @@ func (w *ServerWorker) handleFrame(ctx context.Context, reader *buf.BufferedRead
 	case SessionStatusEnd:
 		err = w.handleStatusEnd(&meta, reader)
 	case SessionStatusNew:
-		// clone outbounds because it is going to be mutated concurrently (Target and OriginalTarget)
-		err = w.handleStatusNew(session.ContextCloneOutbounds(ctx), &meta, reader)
+		err = w.handleStatusNew(ctx, &meta, reader)
 	case SessionStatusKeep:
 		err = w.handleStatusKeep(&meta, reader)
 	default:
--- a/common/mux/server.go
+++ b/common/mux/server.go
@@ -170,7 +170,7 @@ func (w *ServerWorker) handleStatusNew(ctx context.Context, meta *FrameMetadata,
 				b.Release()
 				mb = nil
 			}
-			errors.LogInfoInner(ctx, err,"XUDP hit ", meta.GlobalID)
+			errors.LogInfoInner(ctx, err, "XUDP hit ", meta.GlobalID)
 		}
 		if mb != nil {
 			ctx = session.ContextWithTimeoutOnly(ctx, true)
@@ -286,7 +286,8 @@ func (w *ServerWorker) handleFrame(ctx context.Context, reader *buf.BufferedRead
 	case SessionStatusEnd:
 		err = w.handleStatusEnd(&meta, reader)
 	case SessionStatusNew:
-		err = w.handleStatusNew(ctx, &meta, reader)
+		// clone outbounds because it is going to be mutated concurrently (Target and OriginalTarget)
+		err = w.handleStatusNew(session.ContextCloneOutbounds(ctx), &meta, reader)
 	case SessionStatusKeep:
 		err = w.handleStatusKeep(&meta, reader)
 	default:
--- a/common/mux/server_test.go
+++ b/common/mux/server_test.go
@@ -1,0 +1,124 @@
+package mux_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/xtls/xray-core/common"
+	"github.com/xtls/xray-core/common/buf"
+	"github.com/xtls/xray-core/common/mux"
+	"github.com/xtls/xray-core/common/net"
+	"github.com/xtls/xray-core/common/session"
+	"github.com/xtls/xray-core/features/routing"
+	"github.com/xtls/xray-core/transport"
+	"github.com/xtls/xray-core/transport/pipe"
+)
+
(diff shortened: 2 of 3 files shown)


mmmray:
Without mux, the outbounds array is reset entirely: https://github.com/mmmray/Xray-core/blob/7a80df07b2aa83685d41d72885d201ab47ef9c05/app/proxyman/inbound/worker.go#L83

maybe it's a more appropriate fix?

mmmray:
@yuhan6665 or @RPRX please review, if you have ideas on how to test, tell me as well.

yuhan6665:
Thanks for your great investigation @mmmray!
I have another thought:
As far as I know, we will only have three different target addresses (at most): original from protocol, sniffed, and for routing. Why do we need to mutate? We can refactor so that these addresses should never change?
Maybe, add a convenient function like check if original is fake IP and return sniffed instead..
What do you think @RPRX?

mmmray:
Going even further, I don't know why the context is used to carry Target at all. I suppose that somewhere, it was too annoying to pass as function argument, but I did not notice it. But i'm also new to the code.

yuhan6665:
In general, context should contain (as much as) session level information. I think inbound, outbounds so on and so forth info is ok. XTLS also use it pass raw conn object. A bit strange but I think it is also ok.

mmmray:
I think some refactoring in this area is really needed. The other one is around casting to PacketConnWrapper (see #3711 #3559). Anyway, maybe for another time.

sssagsag:
> I think some refactoring in this area is really needed. The other one is around casting to PacketConnWrapper (see #3711 #3559). Anyway, maybe for another time.

thanks for investigating and fixing , please build new version of core soon as possible
It has been almost a month that the core with new merged has not been released

mmmray:
Indeed I think a new version should be released soon. There are a few bugfixes, and this new fragment PR sitting around in master doesn't help with its lifetime. I was hoping to wait for https://github.com/XTLS/Xray-core/pull/3711 though. Let's see.

sssagsag:
> Indeed I think a new version should be released soon. There are a few bugfixes, and this new fragment PR sitting around in master doesn't help with its lifetime. I was hoping to wait for #3711 though. Let's see.

The new fragment has no problem with any operator in Iran, please don't make a cover for it 😆

sssagsag:
and splithttp almost blocked in iran , beacuse quic blocked in many isp

mikeesierrah:
@sssagsag nope , thats not true

splithttp isnt based on quic , udp mode - quic - h3 is optional

sssagsag:
> @sssagsag nope , thats not true
> 
> splithttp isnt based on quic , udp mode - quic - h3 is optional

yes iknow , but i cant used normally with many isp

sssagsag:
> Indeed I think a new version should be released soon. There are a few bugfixes, and this new fragment PR sitting around in master doesn't help with its lifetime. I was hoping to wait for #3711 though. Let's see.
> 
> I think the next version was planned to be related to splithttp mux settings, but... it's ready when it's ready IMO and shouldn't hold up critical bugfixes.

maybe this helped #3677

mmmray:
To me the problem is the racing write in Dispatch, not the racing read in GetTargetIPs. If I only fix GetTargetIPs as you said, the following sequence is still possible. (Sub-)Connection A and B are part of the same physical mux.cool connection:

* Dispatch from connection A is writing
* Dispatch from connection B is writing
* GetTargetIPs from connection A is reading (it reads the result from Dispatch B)
* GetTargetIPs from connection B i reading (it reads the result from Dispatch B)

So now, xray may route connection A based on information found in connection B! And beside GetTargetIPs, there are many other reads for multiple routing rules. Really, the most problematic part of this bug is not the crash, but the potential mis-routing as a result of the race condition.

PR #3756, opened 2024-09-03
Transport: Remove KCP

mmmray:
This one I am not sure about. There are some members in TG who use this protocol successfully, many more than QUIC. Maybe it can be considered in a later version when Xray has more options for UDP direct connection. Anyway, let's discuss.


Fangliding:
If my memory is fine, removing KCP doesn't seem to be part of the plan

mmmray:
yeah it was mentioned in some thread a few weeks/months back, but not in the most recent messages by rprx.

iambabyninja:
Actually, the number of people using this mode of transport is negligible.  
The percentage of those who use it consciously is even smaller.  
At least in Russia, that's for sure.

PR #3757, opened 2024-09-03
Transport: Remove DomainSocket

mmmray:
As requested. If you need this, why not use the corresponding sockopt?

reapply https://github.com/XTLS/Xray-core/pull/3755


RPRX:
https://github.com/XTLS/Xray-core/pull/3755#issuecomment-2325533003

PR #3793, opened 2024-09-10
Log: Improve feature deprecation/removal logs

KobeArthurScofield:
The default `logLevel` is `warning` so normally there are not many people know the features that are deprecating.

This pr does these:

- Changing the feature deprecation warning lo level `warning`: so normally most people can see the warning, and we can have a good use of it.
- Add feature removal warning: This indicate the feature has been removed, and can be use as a "REMOVE BEFORE FLIGHT" tag so we can remove this warning after a long enough period of time.

这个改动是像原来设计/现在这样放在 `features` 里面，还是转移到 `errors` 比较好？


yuhan6665:
感谢！

PR #3794, opened 2024-09-10
UDP noise: Base64 and multi-packet support

dragonbreath2000:
This was per request of maintainers(mostly RPRX)
usage:
``` json
      {
        "protocol": "freedom",
        "settings": {
          "noise":[
          {
            "packet":"base64:7nQBAAABAAAAAAAABnQtcmluZwZtc2VkZ2UDbmV0AAABAAE=",
            "delay":"5-10"
          },
          {
            "packet":"rand:10-20",
            "delay":"10-20"
          }
        ]
        },
        "tag": "direct"
      }
```
the use of array is optional(to avoid a breaking change)
some cleanup is happened: used int32Range for delay instead of string,moved the config parsing part of noise to another function
now u can send as many noise packets as u want,the code will loop through and send all of them
u can add a base64 encoded string ,copy it straight from wireshark and boom
Note:if u want to use this with h3,do not send a quic client hello as noise,it will confuse the server and respond to that packet instead of intended one


yuhan6665:
@RPRX

RPRX:
看起来换成了 SplitN

按照惯例，noise 改名 noises 以支持数组，noise 留下报错（参考 fallback->fallbacks）

Fangliding:
> 看起来换成了 SplitN
> 
> 按照惯例，noise 改名 noises 以支持数组，noise 留下报错（参考 fallback->fallbacks）

本来就是故意同时支持数组和非数组避免break change的 这样不是白搞了

Fangliding:
我觉得现在这样就挺好的 单个noise就一个结构体 不用写数组 更简洁不用套一层中括号 也兼容上版本的写法
多个noise就上数组

mmmray:
I also like this form more, but if the convention in xray is to have separate pluralized keys, I think it is more important to follow it. ~~I think xray suffers a lot from multiple ways of writing config, depending on who wrote a feature.~~

dragonbreath2000:
> 看起来换成了 SplitN
> 
> 按照惯例，noise 改名 noises 以支持数组，noise 留下报错（参考 fallback->fallbacks）
>
Ok I will change it to noises
Do I keep the optional array type or remove it?~personally I like to keep it since if I only want to have one noise why write a array but if that is the convention here then fine~

mmmray:
If `noise` and `noises` exists, then they can each be object and array, and a dynamic Listable type is not necessary

infra/conf/freedom.go
```diff
@@ -168,12 +152,15 @@ func (c *FreedomConfig) Build() (proto.Message, error) {
 		}
 	}
 	if c.Noise != nil {
-		for _, n := range c.Noise {
+		return nil, errors.New(`Freedom settings: please use "noises":[{}] instead of "noise":{}`)
```
mmmray, line 155: In order to maintain backwards compat, I suggest to append to `config.Noises` instead. The behavior when both are defined can be an error, I think.

infra/conf/freedom.go
```diff
@@ -168,12 +152,15 @@ func (c *FreedomConfig) Build() (proto.Message, error) {
...
+	}
+	if c.Noises != nil {
+		for _, n := range c.Noises {
 			NConfig, err := ParseNoise(n)
 			if err != nil {
-				return nil, err
+				return nil, errors.New(err)
```
mmmray, line 161: why is this wrapping necessary?

dragonbreath2000:
> For direct connections(no cdn & no webserver), it is possible to hardcode fake packets in server and make it ignore the known fake packets, in this way it is possible to make noise to work with TCP too, when client sends known fake packets, must ignore it and continue the connections, fake packets + random data of vmess/ss+tcp for sure will bypass firewall

If we want to this to work for any tcp based protocol,it needs to be implemented in transport layer,not in freedom which is way outside the scope of this pr

RPRX:
不喜欢搞那么多语法糖，~~大道至简~~，这功能刚加的该 break 就 break，不然单个 inbound 和 outbound 就不该移除了

干脆彻底不兼容，参考 https://github.com/XTLS/Xray-core/pull/3812#issuecomment-2351327709 ，分成 type 和 value，别 SplitN 了，只要 noises，配置 noise 就报错

RPRX:
https://github.com/XTLS/Xray-core/pull/3804#issuecomment-2351369556

Fangliding:
确实该分 写文档的时候就觉得这写法老奇怪了

Fangliding:
我稍微改一下吧 这应该很简单

Fangliding:
> > 分成 type 和 value，别 SplitN 了，只要 noises，配置 noise 就报错
> 
> @mmmray 有空改一下吗，不过能不能叫 type 我不确定

小改了一下 这样应该就可以了
可以叫type 之前我写的自定义sockopt结构其实和这个很类似
image

dragonbreath2000:
> > > Split into type and value, don't use SplitN, just use noises, and you will get an error if you configure noise
> > 
> > 
> > @mmmrayDo you have time to change it? But I'm not sure if it can be called type.
> 
> After a little modification, it should work. It can be called type. The custom sockopt structure I wrote before is actually very similar to this one. image

sorry i was not here for some time,I got sick badly,but I was working on it now
I was thinking we could go like this:
``` json
"noises":[
   {
          "rand":"10-20",
          "delay":"5-10"
   },
   {
         "base64":"encoded string"
   }
]
```
Instead of having a Packet and delay we can have each of them separately,I just have to add more stuff to noise struct,the code will be cleaner too, no spliting,no colon nothing,and we can use int32Range for rand too
what do do u think?

RPRX:
@Fangliding https://github.com/XTLS/Xray-core/pull/3804#issuecomment-2351369556

RPRX:
@Fangliding 把 type 移到最上面

RPRX:
@mmmray 可以帮忙改一下吗，~~感觉风扇好像没看到上两条评论~~

mmmray:
not sure I understood, here you go (I feel that providing an example of the new syntax is nice-to-have in the error msg)

mmmray (approved):

Fangliding:
> @mmmray 可以帮忙改一下吗，~感觉风扇好像没看到上两条评论~

~~确实才看到 这不是github网页点一下小铅笔改就行了 换个行而已~~

RPRX:
@Fangliding Freedom 标题改为 noises

Fangliding:
刚好有一些小修改 顺便加进去就好了

PR #3797, opened 2024-09-11
SplitHTTP: Read and validate HTTP/1.1 responses

4nd3r5on:
Changes made to read responses from a server
(and check the response codes) before making a request if HTTPv1 is used


transport/internet/splithttp/connection.go
```diff
@@ -1,11 +1,112 @@
...
 	"io"
 	"net"
+	"strconv"
+	"strings"
 	"time"
 )
 
+var (
+	ErrBadRespCode = errors.New("bad response code")
+	BadCodes       = map[int]struct{}{502: {}, 503: {}, 505: {}}
```
mmmray, line 16: everything but 200 OK is bad, can simplify this (also inline the ErrBadRespCode, it carries no type information anyway)

transport/internet/splithttp/connection.go
```diff
@@ -1,11 +1,112 @@
...
+		}
+	}
+
+	return codes, err
+}
+
+// ConnHolder implements the net.Conn interface
+// adds logic of reading the responses before writing the next request
+// Used as a bugfix for HTTP1.1
+type ConnHolder struct {
```
mmmray, line 65: since this thing is speciifc to h1 i suggest moving it into yet another file `h1_connection` (and renaming the constructor + type)

transport/internet/splithttp/connection.go
```diff
@@ -1,11 +1,112 @@
...
+	ResponsesToRead int
+	Conn            net.Conn
```
mmmray, line 67:
```suggestion
	net.Conn
```

if you do this then you don't need to forward all the methods

transport/internet/splithttp/connection.go
```diff
@@ -1,11 +1,112 @@
...
 )
 
+var (
+	ErrBadRespCode = errors.New("bad response code")
+	BadCodes       = map[int]struct{}{502: {}, 503: {}, 505: {}}
+)
+
+// Optimised to read only response codes
+// Reads response codes until getting EOF or error
+func ConnHttpReadRespCodes(conn net.Conn) (codes []int, err error) {
```
mmmray, line 21: on second look, i think this will actually hang if the connection has not yet produced new responses, which is bad since it tries to read infinite amount of responses. I suggest to read one http response with `net/http` everytime there is a new request, and that's it.

transport/internet/splithttp/connection.go
4nd3r5on: Is it possible to have some "OK-like" codes there, like 100, 201, 202, and etc codes that may not require outputing the error? I wasn't sure about this so decided to make "bad codes" list on which we definitely need to return the error

mmmray: the server is xray, so we can define that 200 is the only valid status code here (you can check hub.go to see that this is the only successful status code in practice)

RPRX:
这个 PR 等 @mmmray 的 approve

mmmray:
@RPRX pairing with him on tg, I'll approve it when ready

RPRX:
> @RPRX pairing with him on tg, I'll approve it when ready

~~希望我们明天能发新版~~

mmmray:
I think the deadline for tomorrow cannot be met with this and the UDP noise PR #3794  ~~(I thought the idea of this new versioning was to get rid of deadlines and too much planning...)~~

Fangliding:
Relax, it can be delayed by a day or two. We don't have a deadline

transport/internet/splithttp/h1_conn.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"bufio"
+	"net"
+)
+
+type H1Conn struct {
+	UnreadedResponsesCount int
+	// To reuse response reader, so we won't lose data
+	// If some EOF will accure
```
mmmray, line 11: the reason you want to reuse the buf reader is because it sometimes reads more data than it returns, and returns the over-read data on the next `Read` call. so it's not related to EOF

mmmray (approved):
I think this PR can be merged, except for one nitpick about some code comment

RPRX:
@mmmray 这个 pr 改个合适的标题

RPRX:
~~好想给这个 pr 重命名为 SplitHTTP client，强迫症发作了~~

StateException:
Sorry for digging up an old PR, but this never actually took effect.

`UnreadedResponsesCount` is declared in `h1_conn.go` and read in `client.go`, but nothing increments it anywhere, so the `http.ReadResponse` branch is dead and the responses are still not read. `git log -S UnreadedResponsesCount` gives just this one commit, so it has been like that since the merge.

It does show up from the outside. Set `scMaxEachPostBytes` to 4096 on the server, leave the client at the default, and `hub.go` starts answering 413. Same server, same binary, only `alpn` differs:

```
http/1.1   the upload hangs, nothing gets through, and the client log stays empty
h2         the app gets a reset in 0.1 s, client logs "bad status code:413 Request Entity Too Large"
```

On h2 `c.client.Do` checks the status; on 1.1 the check sits in the dead branch, so a rejected upload just fails silently.

The responses also pile up in the socket receive buffer since nobody reads them — with a small `scMaxEachPostBytes` I saw around 200 KB per client.

I assume the increment was meant to come later, together with the config value from the TODO. Is that still the plan? (v26.3.27)

Issue #3823, opened 2024-09-16
xmux for grpc and h2


mmmray:
(trying to "plan" this thing so that the designing doesn't happen on the PR itself)

xmux is a setting in splithttp. however, the principle can be applied to h2 and grpc transports. there are a few ways to port it:

## Option 1: Just copy it to `grpcSettings`

```
"grpcSettings": {
    "xmux": ...
}
```

## Option 2: Make it global

`xmux` replaces the `mux` key. For transports that support `xmux`, it is used. For transports that do not support `xmux`, mux.cool is used:

```
"network": "websocket",
"mux": { // mux.cool
    "concurrency": 8
}
```

```
"network": "grpc",
"mux": { // h2mux/xmux
    "concurrency": 8
}
```

can add a `"type"` setting like sing-box, but IMO it can be chosen automatically (for each transport, there is an "obvious best" one)

thoughts?

mmmray:
What does `minConnections` do?

"xmux" just means that it uses the "native mux of HTTP" (H2 mux or QUIC). H1 cannot be done, as HTTP/1.1 has no mux capability.

RPRX:
H2 准备删掉，用 streaming request 代替，gRPC 可以有 xmux，先不改 mux，此贴终结

RPRX:
https://github.com/XTLS/Xray-core/pull/3819#issuecomment-2354548675

RPRX:
以及对于 grpc，Xray 已经改了太多了，我想的是重新设计一下命名为 xgrpc，然后把 grpc ~~这个名称滥用~~给移除

RPRX:
**大概就是建议迁移至 xhttp 和 xgrpc，2025 移除 h2 和 grpc**

mmmray:
Personal opinion, create xhttp, xgrpc, ... first, *long* before deprecating the other stuff. In hindsight, the removal of QUIC would've probably been better if xhttp3 (that is, streaming upload, streaming download, over QUIC) was already launched, because there would've been clear empirical evidence that one is working better (or just as good) as the other. However, QUIC transport itself has few users, so it's fine. I feel differently about gRPC and H2.

For example, there is the assumption that xhttp is a killer for h2 because it has the ability to pass nginx. But from my understanding, xhttp will inherently consume more resources due to the global sessions map, and if we assume this protocol will be fully embraced by users, then we have to assume that these additional anti-censorship properties are so useful in practice that users are happy to make this tradeoff.

[...]

PR #3830, opened 2024-09-19
SplitHTTP server: Add global CORS headers for browser dialer

PoneyClairDeLune:
Spotted a potential oversight sifting through the browser dialer section of SplitHTTP. If there are complications regarding CORS, this PR should fix them.


mmmray:
I typically add them with nginx or cloudflare rules and never connect to the core directly. However, I recognize that this should be supported. I think splithttp in particular has already too many response header settings (`noSSEheader`), is it time for a generic responseHeaders setting?

Or actually, the headers can be added unconditionally if the path is validated, so no option is needed.

I also see you only set them for the GET branch, I think it's better to have those headers in `h.config.WriteResponseHeaders` so they get set for POST as well.

Fangliding:
I think set it default should be enough, no need to add a new field

PoneyClairDeLune:
@mmmray @Fangliding What about merging both into `noDefaultHeaders`?

mmmray:
the reason one would want to disable SSE header is not related to this, so it feels awkward to couple them. I am now convinced that it's best to send them unconditionally.

mmmray (approved):

mmmray:
I think this will work. Thanks!

RPRX:
@mmmray 话说 xmux 好像支持不了 browser dialer，因为 chrome 自动控制复用且没提供控制它的 API？

mmmray:
Yes, I think that's correct. I can't control h2mux behavior or even the HTTP version at all. I guess one could redesign browser dialer so that xray is the one launching and controlling chromium (and then, it can launch many processes or tweak some about:config to force un-muxing), or maybe un-muxing can be done by opening many incognito windows, but it all feels very strange.

RPRX:
@mmmray 写一下 https://github.com/XTLS/Xray-core/issues/3823#issuecomment-2362658554 ，这个月就它了

mmmray:
I've seen the post, can't promise anything

RPRX:
> I've seen the post, can't promise anything by end of month

~~写起来绝对很简单，而且正好周末~~

PR #3835, opened 2024-09-19
Xmux: Value of 0 is treated the same as no value

mmmray:
It turns out that some panels like to set `"xmux": {"maxConnections": 0, "maxConcurrency": 0}`, and of course that fails now. To make their job easier, let's treat `0` the same as not setting the parameter.

Again, I don't like that xray's defaults are hardcoded all over the place. I would have liked a different default value for Xmux in a future version, but it actually can't be done in practice because everybody just copypastes the defaults from the docs into their own sourcecode (and sometimes changes them silently to their own idea of a good default)

### DIFF
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -263,7 +263,7 @@ func (c *SplitHTTPConfig) Build() (proto.Message, error) {
 		c.Host = c.Headers["Host"]
 	}
 
-	if c.Xmux.MaxConnections != nil && c.Xmux.MaxConcurrency != nil {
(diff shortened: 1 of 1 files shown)


Fangliding:
I don't think we should be take responsible for the incorrect behavior of panels

yuhan6665:
I think it is related to protobuf config.
In vanilla proto, you cannot distinguish between an int 0 vs not defined (null)

RPRX:
~~看了下 pr 应该不用发个新版，主要是懒，反正这个月还有新版~~

PR #3799, opened 2024-09-12
Routing: Add regexp syntax support to UserMatcher

vnxme:
This PR implements regexp syntax support to UserMatcher, i.e. allows to match user email in routing rules with regular expressions using `regexp:{pattern}` syntax. It has also been requested in https://github.com/XTLS/Xray-core/discussions/2536.

PR #3900, opened 2024-10-11
Routing: Fix regexp syntax support in UserMatcher

vnxme:
Last month the regexp syntax support in UserMatcher was added by https://github.com/XTLS/Xray-core/pull/3799. I've just noticed I made a typo which prevents regular expression matching at all. Sorry for that. This PR makes it work fine.


yuhan6665:
Thanks!

PR #3949, opened 2024-10-28
Update README.md

ahmad0489:
If you check out [this](https://github.com/MHSanaei/3x-ui?tab=readme-ov-file#ssl-certificate), you'll see that you can add SSL right after installing panel without even going into the panel.

Also it's a good thing that the panel can be opened with HTTP because people can use a reverse proxy like Nginx or Cloudflare (which adds SSL automatically).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
   - [ghcr.io/xtls/xray-core](https://ghcr.io/xtls/xray-core) (**Official**)
   - [teddysun/xray](https://hub.docker.com/r/teddysun/xray)
   - [wulabing/xray_docker](https://github.com/wulabing/xray_docker)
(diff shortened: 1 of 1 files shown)


ahmad0489:
And Please refrain from making false accusations.

RPRX:
1. YouTubers are teaching people to use plain HTTP directly
2. You don't need to listen to 0.0.0.0 to use Nginx
3. It's better to use HTTPS between Cloudflare and your VPS

ahmad0489:
> 1. YouTubers are teaching people to use plain HTTP directly
> 2. You don't need to listen on 0.0.0.0 to use Nginx
> 3. It's better to use HTTPS between Cloudflare and your VPS

YouTubers are responsible for what they teach. A YouTuber might say, 'Give me your bank card information,' and if the YouTuber uses this account, the bank is not guilty because it was the user (cardholder) who did not follow security principles and trusted the YouTuber.

Security principles of the panel and according to the warning inside the panel, it is said to use TLS, just like a bank that tells cardholders not to share their card information with anyone!

So I'm saying again, you're making a false accusation, and that's ugly! I'm opening a new request and changing the text of your accusation to a warning not to use these panels without TLS.

RPRX:
> YouTubers are responsible for what they teach.

YouTubers 是有责任，但那些面板不删默认的明文 HTTP 才是 YouTubers 会教 http://ip 的根本原因

PR #3951, opened 2024-10-29
Update README.md

ahmad0489:
Based on what I said in [this](https://github.com/XTLS/Xray-core/pull/3949#issuecomment-2444064615) message, instead of accusing the panel creator, try to encourage panel users to add TLS.

@RPRX

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@
   - [ghcr.io/xtls/xray-core](https://ghcr.io/xtls/xray-core) (**Official**)
   - [teddysun/xray](https://hub.docker.com/r/teddysun/xray)
   - [wulabing/xray_docker](https://github.com/wulabing/xray_docker)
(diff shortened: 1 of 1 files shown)


Fangliding:
个人意见
```
The following panels have been verified to provide only secure access methods.
We STRONGLY discourage accessing panels using plain HTTP. Because it will cause security problem If you are using panels not listed here, please also configure HTTPS and random path.
```

RPRX:
我写那个 WARNING 在一定程度上是继续敦促它们禁用明文 HTTP，不会改，任何人再发这种 PR 我直接 block

ahmad0489:
> 我写那个 WARNING 在一定程度上是继续敦促它们禁用明文 HTTP，不会改，任何人再发这种 PR 我直接 block

Is your problem only with install.sh?
Does it mean that if the user is asked during installation whether they want TLS and is warned that the absence of TLS can compromise data, the problem will be solved?

Or do you think they should force TLS to be added? Do the creators of web servers like Nginx and Apache force users of these programs to use TLS? Doesn't the absence of TLS in Nginx or Apache lead to data leaks?

@RPRX 
@Fangliding

PR #3968, opened 2024-11-04
Test: Fix incorrect output format

tcpdumppy:
Fix prompt from go vet

`dns/dnscommon_test.go:91:14: non-constant format string in call to (*testing.common).Errorf`


yuhan6665:
Thanks!

PR #4005, opened 2024-11-12
Use a more direct and less error-prone return value

pinglanlu:
Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone. 

Just like  https://github.com/XTLS/Xray-core/blob/44b1dd0e674347691c62a13d227edf0d36b7e290/proxy/proxy.go#L538

### DIFF
--- a/proxy/vmess/validator.go
+++ b/proxy/vmess/validator.go
@@ -81,7 +81,7 @@ func (v *TimedUserValidator) GetAEAD(userHash []byte) (*protocol.MemoryUser, boo
 	if err != nil {
 		return nil, false, err
 	}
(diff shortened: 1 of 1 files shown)


Fangliding:
Thank you for your attention. But I think this change is too small and not worth opening a PR. If you can collect and correct more of these problems, we would be glad to merge, or I can fix this in the next change

pinglanlu:
> Thank you for your attention. But I think this change is too small and not worth opening a PR. If you can collect and correct more of these problems, we would be glad to merge, or I can fix this in the next change

Thank you for your reply. Please give me some time to find and modify all the cases and resubmit them.

PR #4008, opened 2024-11-12
Chore: Use a more direct and less error-prone return value

pinglanlu:
The new pr for https://github.com/XTLS/Xray-core/pull/4005.

I wrote a tool to detect this, but it has a high false positive rate. 

I checked it one by one and I am sure there are only two such examples in the current code...


yuhan6665:
Thanks for your effort!

PR #4010, opened 2024-11-12
Update Dockerfile to Use Multiple Config Files

lxsq:
Docker volume /etc/xray only contain s config.json in default condition. Changing to confdir could provide support to both single file mode and multi-file mode.


yuhan6665:
Thanks!

Fangliding:
难道不应该用更烂大街的conf.d吗

PR #4012, opened 2024-11-13
Chore: Drop dead code in test

zonescape:
As far as I can see the code in the `init()` function in `infra/conf/dns_test.go` does some preparation but the file created is never used. So I removed the code.


yuhan6665:
Thanks

PR #4017, opened 2024-11-14
Chore: Refactor infra/conf.TestToCidrList()

zonescape:
Improvements:
- drop `init()` functions, use more straightforward code
- don't write `geoip.dat` into the project directory*
- don't write `geoip.dat` twice

\*`geoip.dat` is written into the project directory if you have that file in `./resources` only:
- `init()` function in `dns_test.go` sets environment variable `xray.location.asset` to `os.Getwd()` i.e. to `infra/conf`
- `init()` function in `router_test.go` writes `geoip.dat` to `platform.GetAssetLocation()` i.e. to `infra/conf`

### DIFF
--- a/infra/conf/dns_test.go
+++ b/infra/conf/dns_test.go
@@ -2,35 +2,15 @@ package conf_test
 
 import (
 	"encoding/json"
-	"os"
-	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


yuhan6665:
Thanks! Looks good to me

PR #4045, opened 2024-11-21
Test: Remove temporary directory afterwards

zonescape:
The temporary directory created by `testing/scenarios.genTestBinaryPath()` is currently not deleted. Note that `TestMain()` is a [special function that is executed before any tests in a package](https://pkg.go.dev/testing#hdr-Main).


yuhan6665:
Thanks!

PR #4204, opened 2024-12-24
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


yuhan6665:
Thanks!

PR #4217, opened 2024-12-27
Bring quic back

GFW-knocker:
supporting quic with multiple flavors is important for us.
so i decide to bring the removed quic back


GFW-knocker:
oh sorry.
i want to PR myself
not you.

PR #4233, opened 2024-12-31
Freedom: Fix noises delay

GFW-knocker:
fix a typo in noise delay parsing
that always set DelayMin to 0

and thus udp noise delay never executed
https://github.com/XTLS/Xray-core/blob/main/proxy/freedom/freedom.go#L422


RPRX:
~~可以给修改权限吗~~

RPRX:
刚想起来 Int32Range 已确保 From 比 To 小，所以这里不用比较了

Issue #4237, opened 2025-01-01
UDP Noise Bug


titancomputer:
### Integrity requirements

- [X] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [X] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [X] I searched issues and did not find any similar issues.
- [X] The problem can be successfully reproduced in the latest Release

### Description

Hi
i think the method of noise creating has problem.

when i use base64 noise, it converts my base64 to string when it's going to send noise.

it should send real base64 or convert them to hex not string.

or may you can add hex noise in addition to str base64 and rand.

please take a look at these debug logs

2025/01/01 23:12:26 [Debug] [3607294496] proxy/freedom: NOISE[delay_min:10  delay_max:16  **str_noise**:"**\xee\x00\x00\x00\x01\x08\xaa\xaa**" delay_min:10  delay_max:16

### Reproduction Method

run this json with core 25.1.1 and see debug logs
[warp_mmm_noise.json](https://github.com/user-attachments/files/18286474/warp_mmm_noise.json)

### Client config

N/A

### Server config

N/A

### Client log

```
2025/01/01 23:12:26 [Debug] peer(bmXO…fgyo) - Sending keepalive packet
2025/01/01 23:12:26 [Info] [3607294496] proxy/freedom: connection opened to udp:188.114.98.55:988, local endpoint [::]:63497, remote endpoint 188.114.98.55:988
2025/01/01 23:12:26 [Debug] [3607294496] proxy/freedom: NOISE[delay_min:10  delay_max:16  str_noise:"\xee\x00\x00\x00\x01\x08\xaa\xaa" delay_min:10  delay_max:16  str_noise:"\xee\x00\x00\x00\x01\x08\xaa\xaa" delay_min:10  delay_max:16  str_noise:"\xee\x00\x00\x00\x01\x08\xaa\xaa" delay_min:10  delay_max:16  str_noise:"\xee\x00\x00\x00\x01\x08\xaa\xaa" delay_min:10  delay_max:16  str_noise:"\xee\x00\x00\x00\x01\x08\xaa\xaa"]
2025/01/01 23:12:26 [Info] [332116113] proxy/socks: TCP Connect request to tcp:www.gstatic.com:80
2025/01/01 23:12:26 [Info] [332116113] app/dispatcher: sniffed domain: www.gstatic.com
[...]

PR #4239, opened 2025-01-01
Freedom noises: Support "hex" as `type` & `packet`

GFW-knocker:
add option for hex noise.

we need to send noise packet in hex
for example :
QUIC header "ee0000000108aaaaaaaaaaaaaaaa000044d0" in hex 
can't be represented in str or base64 with printable chars.

to see difference please check packets in wireshark with [this config](https://github.com/mahsanet/MahsaFreeConfig/blob/main/udp_noise/warp_mmm_noise.json)

this pr solve this issue: https://github.com/XTLS/Xray-core/issues/4237


Fangliding:
Any binary data can be encoded as base64 (in your example `7gAAAAEIqqqqqqqqqqqqqqqoAAETQ`), the only problem is that the output from the log is incorrect, and you are even continuing to use the wrong variable name

GFW-knocker:
@Fangliding 
i'm not talking about output log
i'm talking about actual packets sent in network
please see packets in wireshark

when you send str or base64 , 
it converted to utf-8 before sent but we need raw hex

example:
str "ee00" --> byte[] {0x65,0x65,0x30,0x30}
hex "ee00" --> byte[] {0xee,0x00}

(also `7gAAAAEIqqqqqqqqqqqqqqqoAAETQ` is not a valid base64 ! but even if so, it sent in utf-8)

Fangliding:
配置文件中可以指定是直接发送str(utf8)抑或是将输入的base64解码为二进制数据并发送
base64除了编码文本还被用于编码字节数据用于在纯文本环境传递(实际上它编码的一直都是字节数据) 你觉得这是无效的base64 因为你只知道它可以编码文本
名字起个 GFW-knocker 结果只会摆弄那几个wireshark输出结果连这么基本的编码知识都不知道

RPRX:
base64 被发明出来就是用于编码 bytes 非可打印字符的，否则它和 str 有什么区别，并且 base64 比 hex 更紧凑（占用空间更少）

不过 hex 的好处是可读性更高，加一个 hex 也不是不可以，但这并不意味着你的说法是对的

GFW-knocker:
@Fangliding 

yes you are right. we can convert hex to base64
i made a mistake because some online tools convert it incorrectly

even your base64 conversion is invalid
`7gAAAAEIqqqqqqqqqqqqqqqoAAETQ`

i put in the config:
`{
              "type":"base64",
              "packet":"7gAAAAEIqqqqqqqqqqqqqqqoAAETQ",
              "delay":"10-16"
},`
i get:
infra/conf: Invalid base64 string

the question is why we don't accept hex ?
wouldn't be easier to have hex option too?

RPRX:
@GFW-knocker rebase 一下

RPRX:
@GFW-knocker 能开放你的仓库的修改权限吗

RPRX:
@GFW-knocker 或者你把我设为 collaborator

Fangliding:
> @Fangliding
> 
> yes you are right. we can convert hex to base64 i made a mistake because some online tools convert it incorrectly
> 
> even your base64 conversion is invalid `7gAAAAEIqqqqqqqqqqqqqqqoAAETQ`
> 
> i put in the config: `{ "type":"base64", "packet":"7gAAAAEIqqqqqqqqqqqqqqqoAAETQ", "delay":"10-16" },` i get: infra/conf: Invalid base64 string
> 
> the question is why we don't accept hex ? wouldn't be easier to have hex option too?

I am not against this feature. The reason why I closed PR is that it cannot fix the issue and conflicts with the real fix

GFW-knocker:
> @GFW-knocker 或者你把我设为 collaborator

i invite you as collaborator
thank you

Issue #4302, opened 2025-01-18
"localhost" dns always returns both ipv4 and ipv6 regardless of queryStrategy!


patterniha:
### Integrity requirements

- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

test version: v25.1.1
test os: android, windows
according to https://xtls.github.io/config/dns.html#dnsobject
after xray-core v1.8.6 built-in dns server support "global" and "per server" queryStrategy.
so according to the documentation,
for example if "global queryStrategy" is UseIP and "dns server 1 queryStrategy" is UseIPv4,
dns server 1 should only return ipv4 and final result should only contain ipv4(if for example
we have only server 1 dns server)
but for example if we only have "localhost" dns with UseIPv4 queryStrategy and run "host cloudflare.com" command in terminal
result incorrectly  contains both ipv4 and ipv6!!!

*** instead of "localhost" if we use other types of dns (udp,tcp,doh,...) it works correctly and dns query result contains only ipv4.

example config with "localhost" dns which incorrectly returns both IPv4 and ipv6: https://raw.githubusercontent.com/patterniha/xray_configs/refs/heads/main/buggy_localhost_dns_example.json

example config with "udp" dns which correctly only returns ipv4: https://raw.githubusercontent.com/patterniha/xray_configs/refs/heads/main/bug_free_udp_dns_example.json

### Client config

buggy config and bug-free config link is in Description

### Server config

my example config is serverless config.

### Client log

no need, just run "host cloudflare.com" in terminal to see the result incorrectly contains both ipv4 and ipv6.

### Server log

no need

Fangliding:
Looks like localhost dns doesn't implement this feature, I will add it later

PR #3728, opened 2024-08-25
Support DNS queryStrategy config for UDP NameServer

xiaorouji:
https://github.com/XTLS/Xray-core/pull/2564 
It seems that UDP is not added, add it now.

PR #4303, opened 2025-01-18
DNS: Implement queryStrategy for "localhost"

Fangliding:
#4302
我也不知道为什么当初没实现 或许是有原因的 可能需要考古

### DIFF
--- a/app/dns/nameserver.go
+++ b/app/dns/nameserver.go
@@ -43,7 +43,7 @@ func NewServer(ctx context.Context, dest net.Destination, dispatcher routing.Dis
 		}
 		switch {
 		case strings.EqualFold(u.String(), "localhost"):
-			return NewLocalNameServer(), nil
+			return NewLocalNameServer(queryStrategy), nil
 		case strings.EqualFold(u.Scheme, "https"): // DOH Remote mode
 			return NewDoHNameServer(u, dispatcher, queryStrategy)
 		case strings.EqualFold(u.Scheme, "https+local"): // DOH Local mode
--- a/app/dns/nameserver_local.go
+++ b/app/dns/nameserver_local.go
@@ -34,23 +34,6 @@ func (s *LocalNameServer) QueryIP(ctx context.Context, domain string, _ net.IP, 
 		err = dns.ErrEmptyResponse
 	}
 
-	// If ipv4 or ipv6 is disabled, filter out the unwanted IPs
-	if err != nil && (!option.IPv4Enable || !option.IPv6Enable) {
-		var FilteredIPs []net.IP
-		for _, ip := range ips {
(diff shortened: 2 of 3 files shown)


patterniha:
> [#4302 ](https://github.com/XTLS/Xray-core/issues/4302) I don’t know why it wasn’t implemented in the first place. Maybe there’s a reason. Archaeology may be needed.

@cty123 added per server queryStrategy on Sep 18, 2023 #2564 
@xiaorouji realize that "udp" was missing on Aug 25, 2024  #3728 , but didn't realized that "localhost" was missing!!! 

and now I accidentally realize that "localhost" was missing  :)

cty123:
When I implemented the feature for localhost, the ci failed on windows for some reasons. I thought it was windows specific setup problem. Since I hadn't been using a windows machine for a long time, I wasn't able to reproduce the problem locally.

patterniha:
> #4302 我也不知道为什么当初没实现 或许是有原因的 可能需要考古

your solution for filtering ipv6 and ipv4 is different from @xiaorouji and @cty123 solutions.

Why do we first query for both IPv6 and IPv4 and then filter? we couldn't just query for the IPv4/ipv6 from the beginning?

Fangliding:
> When I implemented the feature for localhost, the ci failed on windows for some reasons. I thought it was windows specific setup problem. Since I hadn't been using a windows machine for a long time, I wasn't able to reproduce the problem locally.

以前test写的有问题就是动不动炸 不一定是代码问题 一般三个平台有一个跑通了就没啥事 全挂了就是代码有问题 现在已经修的七七八八了 只有macos有时候还是莫名其妙炸掉

我再翻了一下 很久以前这个逻辑本来已经在核心里了 把option丢进去重新应用下就好了 根本没必要过滤

patterniha:
> > When I implemented the feature for localhost, the ci failed on windows for some reasons. I thought it was windows specific setup problem. Since I hadn't been using a windows machine for a long time, I wasn't able to reproduce the problem locally.
> 
> 以前test写的有问题就是动不动炸 不一定是代码问题 一般三个平台有一个跑通了就没啥事 全挂了就是代码有问题 现在已经修的七七八八了 只有macos有时候还是莫名其妙炸掉
> 
> 我再翻了一下 很久以前这个逻辑本来已经在核心里了 把option丢进去重新应用下就好了 根本没必要过滤

now there is another problem for windows:
in windows if my wifi only support ipv4, "localhost" dns only return "ipv4" even if queryStrategy is UseIP.(if UseIPv6 return null)

but in linux it doesn't matter your wifi support ipv6 or not and correctly return ip version base on queryStrategy.
so in linux even if your wifi only support ipv4, "localhost" dns return both ipv4 and ipv6 if queryStrategy is UseIP.

Do I need to open a separate issue for this?

PR #4349, opened 2025-02-04
Outbound: Add outbound sendThrough origin behavior 

lastrise:
In case when there are few IP-addresses on the server, it would be good for outbound to have behavior when it resolves origin IP address and then sends packets through it. Before the choice was delegated to OS.

This behavior can be used with explicit specification in the outbound config:

`{"outbounds":[
    {
      "protocol":"freedom",
      "tag":"DIRECT",
      "sendThrough": "origin"
    }
}`

### DIFF
--- a/app/proxyman/outbound/handler.go
+++ b/app/proxyman/outbound/handler.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	gonet "net"
 	"os"
+	"strings"
 
 	"github.com/xtls/xray-core/app/proxyman"
 	"github.com/xtls/xray-core/common"
--- a/app/proxyman/outbound/handler.go
+++ b/app/proxyman/outbound/handler.go
@@ -8,7 +8,6 @@ import (
 	"math/big"
 	gonet "net"
 	"os"
-	"strings"
 
 	"github.com/xtls/xray-core/app/proxyman"
 	"github.com/xtls/xray-core/common"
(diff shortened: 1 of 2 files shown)


Fangliding:
它被用于什么场景
让我猜猜你有一台有数个IP的服务器分发给多个用户希望他们都能简洁地走自己的IP出去 倒也无可厚非

lastrise:
> 它被用于什么场景? 让我猜猜你有一台有数个IP的服务器分发给多个用户希望他们都能简洁地走自己的IP出去 倒也无可厚非

I would like to utilize all IP addresses of the server. Now whatever address client uses to connect inbound with listen address '0.0.0.0', OS will use only the first IP address attached to interface.

Fangliding:
> OS will use only the first IP address attached to interface.

这倒是很自然的行为 如果应用程序不手动bind的话任何操作系统都是这样 除非自行配置路由表(

lastrise:
> > OS will use only the first IP address attached to interface.
> 
> 这倒是很自然的行为 如果应用程序不手动bind的话任何操作系统都是这样 除非自行配置路由表(

This way makes config unified without host-specific parameters, so we can redeploy one config on many nodes. Are there other ways to do that? Shouldn't attribute sendThrough control this behavior?

Fangliding:
> > > OS will use only the first IP address attached to interface.
> > 
> > 
> > 这倒是很自然的行为 如果应用程序不手动bind的话任何操作系统都是这样 除非自行配置路由表(
> 
> This way makes config unified without host-specific parameters, so we can redeploy one config on many nodes. Are there other ways to do that? Shouldn't attribute sendThrough control this behavior?

It is obvious, setting many sendthrough manually
I'm not against this change, I'm just curious

lastrise:
> > > > OS will use only the first IP address attached to interface.
> > > 
> > > 
> > > 这倒是很自然的行为 如果应用程序不手动bind的话任何操作系统都是这样 除非自行配置路由表(
> > 
> > 
> > This way makes config unified without host-specific parameters, so we can redeploy one config on many nodes. Are there other ways to do that? Shouldn't attribute sendThrough control this behavior?
> 
> It is obvious, setting many sendthrough manually I'm not against this change, I'm just curious(

Sorry ) We have platform that deploys one config for many xray nodes, and it would be very inconvenient to manage routes statically for every node

lastrise:
Hello.
Could you check, please?

@yuhan6665

lastrise:
@RPRX 
@Fangliding 

Hey, could somebody else review it, please?)

RPRX:
我可以 review 这个 PR，但需要等到新版本之后

yuhan6665:
Looks good to me, thanks!

RPRX:
看起来不会导致 panic，虽然后面那个 inbound conn localAddr 可能是 unix，不过这是用户自己配置的问题

我现在给 commit message 的选项值加 ""，比如 "origin"，选项本身加 ``，比如 https://github.com/XTLS/Xray-core/commit/4b01eb439829c9adc090f18467376680f2db7f28 应该是 `listen`，~~我忘改了~~

我感觉这里配置是 sendThrough 但代码里是 via 很不直观，就像配置是 *Settings 但代码里是 *Config，以后我想都给改成一样的

PR #4362, opened 2025-02-06
Chore: Make some Maps into real Sets

arturmelanchyk:


RPRX:
麻烦基于 https://github.com/XTLS/Xray-core/commit/07c35ed52ad0e57e9c47a25c7d029b9d4b09ec52 rebase 一下

arturmelanchyk:
Hi @RPRX 
the branch has been rebased

RPRX:
这是在解决什么问题？Golang 的新行为？

RPRX:
~~虽然 struct{} 在 Golang 中地位确实特殊~~

有相关解释发来看一下吗

RPRX:
你的意思是反正存储 true 或 false 都用不到，不如用空值 `struct{}`？

arturmelanchyk:
@RPRX 
in this specific case `map[string]bool` is used as a `Set`, to keep unique values as keys of the map. The `bool` is redundant and not needed. By having `struct{}` instead of irrelevant `bool` values [we avoid unneeded memory allocation](https://stackoverflow.com/questions/37320287/maptstruct-and-maptbool-in-golang) as well as we avoid confusion for developers

PR #4451, opened 2025-03-01
Chore: Use native h2c support

yin1999:
Go 1.24 has add the native unsecured http2 support: https://redirect.github.com/golang/go/issues/67816. And the [x/net/http2/h2c package is going to be deprecated](https://redirect.github.com/golang/go/issues/72039). We may be able to replace the x/net/http2/h2c package and use native h2c support instead.


RPRX:
改了下标题

Issue #4531, opened 2025-03-22
certificate files relative path problem


patterniha:
[windows v25.3.6]

suppose the file "xray.exe" is in folder "F1",
and you run command ".../F1/xray.exe" from folder "F2".

geoip.dat and geosite.dat should be in "F1" folder, **which shows that the relative path for geofiles begins from the folder where "xray.exe" is there.**

but certificate files should be in "F2" folder, **which shows that the relative path for certificate files begins from the folder where the "command" is executed.**

```
"certificateFile": "mycert.crt",
"keyFile": "mycert.key"
```

This issue has caused the mitm-config to not work in v2rayN and v2rayNG, because the The folder where the command is executed is unspecified.

///

as a result, like geofiles, the relative path for certificate files should begins from the folder where "xray" is there.

Fangliding:
性质不一样 geofile 是 assets 是去固定文件夹下面找，证书是指定文件路径，你有见过 geofile 要你写路径吗？这个 assets 文件夹是通过命令行参数传入的 默认的 xray 二进制所在的文件夹而已
你要写证书 最好直接写在配置文件 而且你也可以写绝对路径

patterniha:
> Different in nature. Geofile is an asset and is found in a fixed folder, while a certificate is a specified file path. Have you ever seen a geofile ask you to write a path? This assets folder is the default folder where the xray binary is located, which is passed in through the command line parameter. If you want to write a certificate, it is best to write it directly in the configuration file and you can also write an absolute path.

@Fangliding 

absolute path is different for each user, the users should edit the "serverless_with_mitm_for_Iran.jsonc" before use,
this is not convenient for all users.

if you run "xray" from "C:\Users\username"(default powershell execute folder) the certificate should be in "C:\Users\username" folder instead of xray-folder!!! why???

///

also, what is the "xray" absolute path in v2rayNG???

using absolute path is impossible.

 @RPRX

This issue creates a serious problem for using MitM-config, Most users are unable to use mitm-config because of this issue.

patterniha:
> Different in nature. Geofile is an asset and is found in a fixed folder, while a certificate is a specified file path. Have you ever seen a geofile ask you to write a path? This assets folder is the default folder where the xray binary is located, which is passed in through the command line parameter. If you want to write a certificate, it is best to write it directly in the configuration file and you can also write an absolute path.

Your argument for geofiles is correct, but you didn't give an argument for certificate as to why its relative path should begin from command path.

[...]

PR #4536, opened 2025-03-23
Add "certificate" environment variable

patterniha:
Wrongly, the source of relative path for certificate files is "command-run" folder: https://github.com/XTLS/Xray-core/issues/4531

also, it is not elegant that we put all certificate files in exec folder, so we need environment-variable for certificate files.

therefore, gui/panel can create a folder for certificates files and set certificate-environment-variable to that folder.

also, v2rayNG only has one folder for asset files, so in v2rayNG we can set
geo-environment-variable=certificate-environment-variable and then we can add certificate files.

///

logic:

```
if cert/key path is absolute path:
    read from that path
else:
    if we have certificate-environment-variable:
        read from certificate-environment-variable folder
    else:
        read from exec folder
```

### DIFF
--- a/common/platform/filesystem/file.go
+++ b/common/platform/filesystem/file.go
@@ -3,6 +3,7 @@ package filesystem
 import (
 	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/xtls/xray-core/common/buf"
 	"github.com/xtls/xray-core/common/platform"
@@ -26,6 +27,13 @@ func ReadFile(path string) ([]byte, error) {
 
 func ReadAsset(file string) ([]byte, error) {
 	return ReadFile(platform.GetAssetLocation(file))
+}
+
+func ReadCertificate(file string) ([]byte, error) {
+	if filepath.IsAbs(file) {
+		return ReadFile(file)
+	}
+	return ReadFile(platform.GetCertificateLocation(file))
 }
 
 func CopyFile(dst string, src string) error {
--- a/common/platform/others.go
+++ b/common/platform/others.go
@@ -42,3 +42,8 @@ func GetAssetLocation(file string) string {
 	// asset not found, let the caller throw out the error
 	return defPath
 }
+
+func GetCertificateLocation(file string) string {
+	certificatePath := NewEnvFlag(CertificateLocation).GetValue(getExecutableDir)
+	return filepath.Join(certificatePath, file)
+}
--- a/common/platform/platform.go
+++ b/common/platform/platform.go
@@ -8,11 +8,12 @@ import (
 )
 
 const (
-	PluginLocation  = "xray.location.plugin"
-	ConfigLocation  = "xray.location.config"
-	ConfdirLocation = "xray.location.confdir"
-	ToolLocation    = "xray.location.tool"
-	AssetLocation   = "xray.location.asset"
+	PluginLocation      = "xray.location.plugin"
+	ConfigLocation      = "xray.location.config"
+	ConfdirLocation     = "xray.location.confdir"
+	ToolLocation        = "xray.location.tool"
+	AssetLocation       = "xray.location.asset"
+	CertificateLocation = "xray.location.certificate"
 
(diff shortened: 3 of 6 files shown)


common/platform/platform.go
```diff
@@ -8,11 +8,12 @@ import (
...
-	ConfigLocation  = "xray.location.config"
-	ConfdirLocation = "xray.location.confdir"
-	ToolLocation    = "xray.location.tool"
-	AssetLocation   = "xray.location.asset"
+	PluginLocation      = "xray.location.plugin"
+	ConfigLocation      = "xray.location.config"
+	ConfdirLocation     = "xray.location.confdir"
+	ToolLocation        = "xray.location.tool"
+	AssetLocation       = "xray.location.asset"
+	CertificateLocation = "xray.location.certificate"
```
RPRX, line 16: CertLocation = "xray.location.cert"

common/platform/others.go
```diff
@@ -42,3 +42,8 @@ func GetAssetLocation(file string) string {
 	// asset not found, let the caller throw out the error
 	return defPath
 }
+
+func GetCertificateLocation(file string) string {
```
RPRX, line 46: GetCertLocation

RPRX:
@patterniha

patterniha:
done.

common/platform/filesystem/file.go
```diff
@@ -28,6 +29,13 @@ func ReadAsset(file string) ([]byte, error) {
 	return ReadFile(platform.GetAssetLocation(file))
 }
 
+func ReadCertificate(file string) ([]byte, error) {
```
RPRX, line 32: ReadCert

RPRX:
@patterniha

common/platform/windows.go
```diff
@@ -24,3 +24,8 @@ func GetAssetLocation(file string) string {
 	assetPath := NewEnvFlag(AssetLocation).GetValue(getExecutableDir)
 	return filepath.Join(assetPath, file)
 }
+
+func GetCertLocation(file string) string {
+	certificatePath := NewEnvFlag(CertLocation).GetValue(getExecutableDir)
```
RPRX, line 29: certPath

common/platform/others.go
```diff
@@ -42,3 +42,8 @@ func GetAssetLocation(file string) string {
 	// asset not found, let the caller throw out the error
 	return defPath
 }
+
+func GetCertLocation(file string) string {
+	certificatePath := NewEnvFlag(CertLocation).GetValue(getExecutableDir)
```
RPRX, line 47: certPath

RPRX, line 47: comment

patterniha:
done, all you had to do was say rename all "certificate" to "cert" :)

patterniha:
> Just change the readdfile method to read it once in the exec folder without fallback to the asset directory (default binary location) and read it once. In the future, similar things can be reused by calling this method directly. This is not a separate one.

we can add "readFileFromExec" function:
```
func readFileFromExec(file string) ([]byte, error) {
	if filepath.IsAbs(file) {
		return ReadFile(file)
	}
	return ReadFile(filepath.Join(getExecutableDir(), file))
}
```

But it is not useful for our needs, because we should first check if the file-path is absolute then check the environment-variable and then check the exec folder.

Or maybe I didn't understand what you meant, anyway, you can make any changes you want.

RPRX:
我主要是觉得加个 `XRAY_LOCATION_CERT` 是可以有的，毕竟 `XRAY_LOCATION_ASSET` 都有

反正我加这些东西都是从 ~~为了全人类~~ 以及完善 Xray 基础功能的角度切入

yuhan6665:
https://github.com/XTLS/Xray-core/pull/2911

RPRX:
> #2911

@Fangliding 文档需要更正 `masterKeyLog`，以及 `expectIPs` 更名 `expectedIPs`，以及新增 `XRAY_LOCATION_CERT`

PR #4558, opened 2025-03-27
DNS: change error loglevel for DOH for NXDOMAIN

patterniha:
dns-query-answer for some domains contains no ip, NXDOMAIN
```
nslookup jkasssahdjashdjh.com

-> server can't find jkasssahdjashdjh.com: NXDOMAIN
```

in Xray-core when dns is DOH and and dns-query answer has no ip (NXDOMAIN), an error-log is created and printed:

`2025/03/27 17:09:46.372277 [Error] app/dns: failed to retrieve response for jkasssahdjashdjh.com. > Post "https://free.shecan.ir/dns-query": context canceled`

**this log is only shown for DOH-dns, and we have no log for other types of dns.**

///

some apps in android constantly asking for IP of an invalid domain(NXDOMAIN), this causes thousands of unnecessary logs in v2rayNG and slows down the app.

///

this is only a temporary solution.(and only for log)

### DIFF
--- a/app/dns/nameserver_doh.go
+++ b/app/dns/nameserver_doh.go
@@ -260,7 +260,7 @@ func (s *DoHNameServer) sendQuery(ctx context.Context, domain string, clientIP n
 			}
 			resp, err := s.dohHTTPSContext(dnsCtx, b.Bytes())
 			if err != nil {
-				errors.LogErrorInner(ctx, err, "failed to retrieve response for ", domain)
+				errors.LogInfoInner(ctx, err, "failed to retrieve response for ", domain)
 				return
 			}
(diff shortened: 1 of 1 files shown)


patterniha:
Also, I think NXDOMAIN should be cache for 2 minutes to avoid consecutive requests which, in addition to the log, also has a negative impact on the network.

patterniha:
@Fangliding 

My main request is about **log**, but you talk about requests and cache.

If NXDOMAIN error log should be displayed, it should be displayed for all types of dns, And if it shouldn't be displayed, it shouldn't be displayed for all.

But it is only displayed for DOH.

///

Requests and cache are a separate discussion.

PR #4560, opened 2025-03-27
DNS: Use cache for NXDOMAIN (rcode 3)

patterniha:
First, I have to thank @Fangliding for his guidance on https://github.com/XTLS/Xray-core/pull/4558#issuecomment-2759007356.

I mistakenly think that the problem is related to log, but in fact log is just log and has nothing to do with performance.

in https://github.com/XTLS/Xray-core/pull/3784 @Fangliding change the original code `err != errRecordNotFound` to `err == nil || err == dns_feature.ErrEmptyResponse` To be more accurate.

but in android some apps send thousands of requests in a matter of seconds, that return `NXDOMAIN`.

This causes performance to drop drastically, and many people have complained about this.

So I also added `NXDOMAIN` to the modes that should use `cache`.

photo_2025-03-28_02-10-14

### DIFF
--- a/app/dns/nameserver_doh.go
+++ b/app/dns/nameserver_doh.go
@@ -357,7 +357,7 @@ func (s *DoHNameServer) QueryIP(ctx context.Context, domain string, clientIP net
 		errors.LogDebug(ctx, "DNS cache is disabled. Querying IP for ", domain, " at ", s.name)
 	} else {
 		ips, ttl, err := s.findIPsForDomain(fqdn, option)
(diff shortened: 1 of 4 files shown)


xqzr:
相关链接：
https://t.me/projectXray/4241610
https://t.me/projectXray/4241636

Meo597:
> 话说我之前改 DNS 时也注意到了，ray 每多支持一种 DNS 查询方式就是 copy 整份代码，这挺劣质的，有空把它们改为复用吗？

#5248

PR #4626, opened 2025-04-18
modify udp-listen-address and dialer-connect-timeout

patterniha:
During implementing "happy eyeballs", I encountered a strange thing:

in https://github.com/XTLS/Xray-core/blob/0995fa41fe692e332412670665bd934e4c734caa/transport/internet/system_dialer.go#L56-L60

the "0.0.0.0" set for udp listen-IP, even when the dest address is IPv6 !

This affects the `ListenPacket` function.

although, it seems golang ignore type of "0.0.0.0" when network type is udp(= udp4 & udp6)

anyway, i think it is better to set ":0" for listen address.

///

also, chrome tcp connect timeout is 21 seconds, so it is better set dialer timeout to 21 seconds like chrome.

### DIFF
--- a/transport/internet/system_dialer.go
+++ b/transport/internet/system_dialer.go
@@ -53,10 +53,10 @@ func (d *DefaultSystemDialer) Dial(ctx context.Context, src net.Address, dest ne
 
 	if dest.Network == net.Network_UDP && !hasBindAddr(sockopt) {
 		srcAddr := resolveSrcAddr(net.Network_UDP, src)
-		listenHostPort := ":0"
+		listenAddress := ":0"
 		listenNetwork := "udp"
 		if srcAddr != nil {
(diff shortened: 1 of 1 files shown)


RPRX:
> although, it seems golang ignore type of "0.0.0.0" when network type is udp(= udp4 & udp6)

现有代码没问题的话就继续让它 work，改成 :0 反而在 IPv4 时有歧义，~~就是说反正怎么改都有歧义~~

> also, chrome tcp connect timeout is 21 seconds, so it is better set dialer timeout to 21 seconds like chrome.

可以再开个 PR

patterniha:
@RPRX 

Please take some time to check "DNS: Reviewed" and "happy eyeballs" PR.

I checked everything, If there is any ambiguity, I can explain what I did.

Issue #4473, opened 2025-03-06
feature: happy eyeballs (RFC 8305) for UseIP-domainStrategy


patterniha:
(test on: windows-v25.3.3)

in freedom-outbound and a tcp-connection, or in proxy-outbound and a tcp/udp(quic)-connection,
suppose that our connect-address is a "domain" and [ip1, ip2, ip3, ip4] is the list of resolved-IPs.

**Current behavior**:
- it tries to connect to ip1 for 30 seconds, and if it fails, it tries to connect to ip2 for 30 seconds, and ...
- so if, for example, ip1 and ip2 are unavailable for any reason, It takes 60 seconds to connect to the ip3 and as a result, it takes **60 seconds** to connect to the destination !!!

**Correct behavior (happy eyeballs RFC 8305)**:
- First, we sort the IP-list in such a way that v4 and v6 IPs appear one after the other, for example if ip1 and ip2 are version-4, and ip3 and ip4 are version-6 after sorting we have [ip1, ip3, ip2, ip4].
- then try to connect to ip1 in thread1, after for example 300ms try to connect to ip3 in thread2, ...
- **As a result several attempts may be in flight at the same time, and whenever one of the attempts succeed, all other connections are cancelled, and the winning connection is used. This means a much shorter wait before one of the IP addresses connect successfully.**

**First application**:
- Happy Eyeballs is particularly important for dual-stack clients, when some hosts may have resolvable IPv6 addresses that are somehow unreachable.

**Second application**:
- For any reason, some IPs may not be available, as a result, with happy eyeballs, we can find the IP that can be connected as soon as possible.

**Application in iran**:
- In Iran some instagram IPs are blocked and some are not, in resolved-ip-list some IPs are blocked and some are not, as a result, with happy eyeballs, we can find the IP that can be connected as soon as possible.

**tips**:
- For Happy eyballs, it doesn't matter whether the IP is version 4 or 6, but it is better to sort the resolved-ip-list so that v4 and v6 IPs appear one after the other.([RFC 8305](https://datatracker.ietf.org/doc/html/rfc8305.html))
- happy eyeballs is used in freedom-outbound when address is a domain and connection is tcp,
- or is used in proxy-outbounds when address is domain and connection is tcp or udp(quic).

///
in python stock asyncio supports Happy Eyeballs, so we just need to set the time to wait for a connection attempt to complete, before starting the next attempt in parallel.

so there should be a ready library for happy eyeballs for tcp-connections in Go too.

[...]

PR #4635, opened 2025-04-21
New feature: Happy Eyeballs (RFC 8305)

patterniha:
when using "AsIs" mode, golang controls how to connect to a "domain" address.

currently, golang use RFC-6555 to connect to the domains, although it is obsolete, but it is better than nothing. it solve the problem when one IP-type is unreachable. but regardless of the IP type, some IPs may not be available, thus RFC 8305 was created https://github.com/XTLS/Xray-core/issues/4473#issuecomment-2708057545

RFC 6555 is obsoleted by RFC 8305, and python use RFC-8305 from v3.8.

///

anyway, It doesn't matter what RFC is implemented for the "AsIs" mode, because we should use built-in-dns to bypass GFW DNS spoofing, 
and for "useIP/forceIP" mode, we don't even have RFC-6555, and only one random IP is selected.

**even Xray-core cannot detect user network and for IPv4-only network, an IPv6 can be chosen!**(we can use forceIPv4v6 for this problem but this makes IPv4 always a priority)

and also Iran-GFW blocked some range-IPs of meta(facebook, instagram,...) and some not, so with happy-eyeballs iranian-users  can access instagram without using any server.(in 99% cases)

as a result, I implemented Happy Eyeballs RFC-8305 for "forceIP/useIP" mode.

///

because happy eyeballs only applies when sockopt-domainStrategy is "forceIP/useIP" and this options are in "sockopt" settings, so i put "happyEyeballs" settings in "sockopt" settings.

///

we can use happy eyeballs for all type of proxy: freedom, vless, ...

///

```
"streamSettings": {
  "sockopt": {
    "domainStrategy": "forceIP",
      "happyEyeballs": { 
              "enabled": true,
              "prioritizeIPv6": true, 
              "maxConcurrentTry": 4,
              "tryDelayMs": 250,
              "interleave": 1
       }
     }
   } 
```
///

`enabled` (bool): enable/disable happy eyeballs, default is false (but i think it should be enabled by default later)

`prioritizeIPv6` ( bool): indicate "First Address Family" in RFC-8305, default is false

`interleave` (uint32): indicate "First Address Family count" in RFC-8305, default is 1

`maxConcurrentTry` (uint32):   maximum concurrent attempt (this is only maximum and in most cases our concurrent attempts is less, unless all connection fail to connect) also we can always have a maximum of concurrent-attempt as many IPs as we have, and this option is useful when the number of IPs is too high, and we want to control the number of concurrent-attempts, default is 4.

`tryDelayMs`: delay time between each attempt, RFC-8305 recommend 250ms, so the default is 250ms

///

for example suppose our IP-list is [ip4-1, ip4-2, ip4-3, ip4-4, ip6-1, ip6-2, ip6-3, ip6-4]

when `interleave ` is 1 and `prioritizeIPv6 `is false, the sorted-ip-list is:
[ip4-1, ip6-1, ip4-2, ip6-2, ip4-3, ip6-3, ip4-4, ip6-4]

and when for example `interleave `is 2 and `prioritizeIPv6 ` is true:
[ip6-1, ip6-2, ip4-1, ip4-2, ip6-3, ip6-4, ip4-3, ip4-4]

then delay 250ms for each attempt until first connection is established.

the first-stablished-connection is winner connection and selected for sending/receiving data.


xqzr:
它是否适用于 QUIC？

patterniha:
> Does it work with QUIC?

No, it is implemented in dialer layer, and only TCP can be use it.
Because quic is udp and udp is connection less.

For quic we should implement it at higher level.(and just for non-freedom outbounds)

Anyway, Happy eyeballs has two important algorithms, **sort-IP algorithm** and **Race-algorithm** both of them can be reused for QUIC.

patterniha:
@RPRX 

Happy eyeballs is just two simple algorithms sort-IP algorithm and race algorithm. I don't understand why it takes so long to merge.

Also, You forgot to return error for `ForceIP` mode when no ip found, and wrongly, `ForceIP` had no difference with `useIP`.(for sockopt-domainStrategy)

RPRX:
> You

~~印象中你提到的这两处代码应该不是我写的~~，如果可以的话，能否拆分 PR

> That's all I did in this PR. I don't know why it took so long to merge.

只是最近时间少，合理的 PR 最终都会 merge

patterniha:
> If possible, can you split the PR?

@RPRX 

They are dependent on each other and cannot be separated because the first PR must be based on the second PR, not on XTLS:main.

patterniha:
@RPRX 

Ok, I open this PR after "sockopt-dialerProxy" and "sockopt-forceIP" bugs are fixed in https://github.com/XTLS/Xray-core/pull/4661

PR #4666, opened 2025-04-29
DNS New Features: `disableCache`, `finalQuery`, `unexpectedIPs`, `"*"`, `UseSystem-queryStrategy`, `useSystemHosts`

patterniha:
### New features:

**`disableCache `** for each DNS-Server-Object:

currently we have only one global disableCache option that affects all DNS-servers, but we may want to disable the cache only for a specific DNS-Server.

///

**`finalQuery`** for DNS-Server-Object:**

Suppose you want to use DNS-Server-A for "youtube.com", but use DNS-Server-B for other google sites and use DNS-Server-C for others, so you should set:

```
"dns":{
    "servers": [
      {
       "address": "server-A",
       "domains": ["youtube.com"]
      },
      {
       "address": "server-B",
       "domains": ["geosite:google"]
      },
      "server-C"
    ]
}

```

But for whatever reason, server-A may be unavailable for a while(for example, the network may be unreachable for a while) so it uses server-B for "youtube.com", but we don't want this to happen. 

Currently, there is no mechanism to prevent using server-B for "youtube.com", this is due to strange behavior of `skipFallback` (except creating custom-geosite where "youtube.com" is removed from "google" list, but this is not possible for all users)

but now we can set `finalQuery= true`  for server-A, so any result from server-A return as a final-result and no other DNS-server will be performed.

```
"dns":{
    "servers": [
      {
       "address": "server-A",
       "domains": ["youtube.com"],
       "finalQuery": true
      },
      {
       "address": "server-B",
       "domains": ["geosite:google"]
      },
      "server-C"
    ]
}

```
 
///

**`unexpectedIPs`** for DNS-Server-Object:**

Suppose we want no IP to be in an IP-range-A, and if all IPs in IP-range-A, the next-dns-fallback performed.
for example for Serverless-for-Iran anti-sanction-version, i want to use a anti-sanction DNS, but goverment-run-anti-sanction-DNS only bypass sanctions and not filter.
IRGFW return 10.10.34.0/24, 2001:4188:2:600:10:10:34:0/120 range for blocked domain, so if the return-IPs is in these ranges, the fallback-DNS should be performed.

one way to achieve this goal is creating custom-geosite and then using **`!`** sign, but this is not possible for all users.
another way is to calculate reverse-CIDR-list, for example using [online-tools](https://krasovs.ky/2021/07/04/wireguard-allowed-ips.html) to calculate reverse-CIDR-list, but the reverse-CIDR-list is long and it causes the configuration to be messy.

as a result i add `unexpectedIPs` option, and an IP is matched if and only if does not match any of the IP-ranges in the `unexpectedIPs `list, in other words: 
`expectedIPs = [0.0.0.0/0, ::/0] - unexpectedIPs.`

also, we may need all IPs to be in range-A, and no IP to be in range B, so we need both `expectedIPS `and `unexpectedIPs`: `expectedIPS=[range-A], unexpectedIPs=[range-B]`

```
"dns":{
    "servers": [
      {
       "address": "anti-sanction-dns",
       "unexpectedIPs": ["10.10.34.0/24", "2001:4188:2:600:10:10:34:0/120"]
      },
      "fallback-dns"
    ]
}
```

///

**Add to Documentation:**

```
(dns-server settings)

disableCache: bool

* The cache is disabled, if disableCache = true in DNS-settings or this DNS-Server-settings.

///

finalQuery：bool

* when true, the result is returned in any case(even when IP-list is empty) and no other fallback will be performed.

///

expectedIPs: [string]

...

if you add "*" in this list, the original-IPs still returned if no ip matched.


///
unexpectedIPs: [string]

* reverse form of expectedIPs, an IP is matched if and only if does not match any of the IP-ranges in the list, in other words: 
`expectedIPs = [0.0.0.0/0, ::/0] - unexpectedIPs.`

if you add "*" in this list, the original-IPs still returned if no ip matched.

///
queryStrategy: "UseIP" | "UseIPv4" | "UseIPv6" | "UseSystem"

when queryStrategy is UseSystem, every time dns-Query call, it check the system-network to see if it supports IPv6(and IPv4) or not, if it support IPv6(or IPv4), the IPv6(or IPv4) is also returned, otherwise not returned.

///
(dns settings)

useSystemHosts: bool

if true, system-hosts appends to config-hosts at start, default is false.
```


j2rong4cn:
@RPRX 我写了个只返回rcode的DNS服务，要不要合并到这里？
```json
{
  "domains": [
    "geosite:category-ads-all",
    "domain:lan",
    "regexp:^[^\\.]*$"
  ],
  "skipFallback": true,
  "finalQuery": true,
  "address": "rcode://3"
}
```

https://github.com/j2rong4cn/Xray-core/commit/d08f64833eecc6c843a6831f2593f0207bf5a646

patterniha:
> @RPRX 我写了个只返回rcode的DNS服务，要不要合并到这里？
> 
> ```json
> {
>   "domains": [
>     "geosite:category-ads-all",
>     "domain:lan",
>     "regexp:^[^\\.]*$"
>   ],
>   "skipFallback": true,
>   "finalQuery": true,
>   "address": "rcode://3"
> }
> ```
> 
> [j2rong4cn@d08f648](https://github.com/j2rong4cn/Xray-core/commit/d08f64833eecc6c843a6831f2593f0207bf5a646)

although, it is possible to add rcode-only-dns-server but it is better to add this feature to the `DNS-hosts` instead of adding rcode-only-dns-server.

i change the `hosts` code to achieve this goal, but currently we only have rcode=0:
https://github.com/XTLS/Xray-core/pull/4673

i think there is no difference between rcode=0, and rcode=3 or others in practice, because we don't return any IP.

anyway, if rcode-number is important for your job, you can extend `hosts` code to return desired rcode.

Anyway, @RPRX  has to make the final decision and I just said my opinion.

j2rong4cn:
@patterniha
I need to return rcode=3. I used nslookup to test and found that nslookup will fallback when rcode=0.
I see StaticHosts.Lookup does not return an error, so I wrote rcode-only-dns-server

patterniha:
> @patterniha I need to return rcode=3. I used nslookup to test and found that nslookup will fallback when rcode=0. I see StaticHosts.Lookup does not return an error, so I wrote rcode-only-dns-server

so i think it is better to extend `hosts` code, for example:
`"geosite:category-ads-all": "#3"`

This is definitely a better and more direct way, anyway @RPRX gives the final opinion.

j2rong4cn:
> > @patterniha I need to return rcode=3. I used nslookup to test and found that nslookup will fallback when rcode=0. I see StaticHosts.Lookup does not return an error, so I wrote rcode-only-dns-server
> 
> so i think it is better to extend `hosts` code, for example:
> 
> `"geosite:category-ads-all": "#3"`
> 
> This is definitely a better and more direct way, anyway @RPRX gives the final opinion.

Good, let's do as you say.

RPRX:
简单看了一下，直接把 `geoip` 改为 `expected_geoip` 这种 breaking change 是不行的，会严重损坏现有配置

patterniha:
> After a quick look, it is not feasible to `geoip`change directly to this breaking change, which will seriously damage the existing configuration.`expected_geoip`

@RPRX 

No, it is 100% compatible with current configuration.
This renames only affects the code, and not json.

So it is not change the current logic and configuration.

RPRX:
我觉得用 `priorIPs` 取代 `allowUnexpectedIPs` 是合理的，不过 geo 是支持反选的比如 `"geoip:!cn"`，所以两个 un 似乎没必要

RPRX:
不过现在反选只支持 geo，如果有需要的话再加一个反选 IP/CIDR 的语法就差不多了

RPRX:
还有就是先 `expectedIPs` 再 `priorIPs`，看名称的话顺序不太对，应当把 `priorIPs` 改名为 `secondExpectedIPs`

patterniha:
> I think it is reasonable to `priorIPs`replace with `allowUnexpectedIPs`, but geo supports inverse selection `"geoip:!cn"`, so two uns seem unnecessary.

@RPRX 

Yes, support inverse but:
1. You cannot `and` two inverse, for example you cannot have `!geoip:cloudflare && !geoip:cloudfront`
2. You can inverse only geoip, and you cannot inverse Non-geoip cird,  for example i need inverse of {10.10.34.0/24 + 2001:4188:2:600:10:10:34:0/120} for serverless-for-Iran-anti-saction but we can't even have inverse of {10.10.34.0/24}, let alone the inverse of both.

patterniha:
> Also, first , `expectedIPs`then `priorIPs`, the order is not right if you look at the names, you should `priorIPs`change to`secondExpectedIPs`

First, We will rarely need both at the same time.
Second, Even if we have both, it doesn't matter whether we apply `expectIPs` first or `priorIPs` first.

RPRX:
> First, We will rarely need both at the same time.
Second, Even if we have both, it doesn't matter whether we apply `expectIPs` first or `priorIPs` first.

这俩不就是为了配合起来使用的吗，expect 两次，且有一个明确在前面，~~还有人要的话就 third~~

RPRX:
话说我突然想到，`expectedIPs` 作为一个数组，可能也像 https://github.com/XTLS/Xray-core/pull/4673 以前的 `hosts` 一样把内容给合并了

或许直接更改它的行为，改为不合并，确保数组内第一个就是 first，第二个就是 second，以此类推，~~还是说本来就是这样的逻辑~~

patterniha:
`ExpectedIPs`, `unexpectedIPs`, `priorIPs`, `unpriorIPs` 
**Are 4 distinct options**, Although we can have several of them at the same time,but it is a rarely used.

`ExpectedIPs` is clear.

`unexpectedIPs` is reverse-form of `ExpectedIPs` and i explained why we need reverse-form too.

`priorIPs` logic is simple: act same as `expectedIPs` but if no IP matched, the original-IPs still returned without any change, it is useful for workers/MitM configs.


`unpriorIPs` is reverse-form of `priorIPs`

RPRX:
我误解了你，你也误解了我，我本来以为你那个 `priorIPs` 是填 "0.0.0.0/0" 和 "::/0" 用的，作为第二层 expected

但是这个误解正好导致一个更简洁而强大的想法：把 `expectedIPs` 改为逐级匹配的，这一点你还没看懂 https://github.com/XTLS/Xray-core/pull/4666#issuecomment-2843772399

至于如何合并反选，我还得再想想

patterniha:
> It suddenly occurred to me that `expectedIPs`as an array, it might be possible to merge the contents like before #4673`hosts`
> 
> Maybe just change its behavior to not merge, making sure the first one in the array is first, the second one is second, and so on.~Or is this the logic?~

Suppose you want your IPs to be in range A but not in range B.
So you simply set: 
`ExpectedIPs = [A], unexpectedIPs = [B]`

Therefore, the presence of `unexpectedIPs` next to `expectedIPs` is mandatory and the goal cannot be achieved by just changing syntax.

RPRX:
我觉得四个选项叠一起的话逻辑有点混乱，压成两个选项就行，`unexpectedIPs` 是可以加的，它们是合并而不是逐级，且作用于 `expectedIPs` 每一级匹配成功之后（除非 "0.0.0.0/0 或 "::/0"），比如第一级匹配到了但被 unexpectedIPs 否决了就继续下一级

RPRX:
这样就可以 cover 所有需求了，更少的配置项，更强大的功能

patterniha:
Can you give an example of how {10.10.34.0/24 + 2001:4188:2:600:10:10:34:0/120} can be  reversed with using your method?

patterniha:
As i understand, you agree with `unexpectedIPs` but not with `priorIPs` and `unpriorIPs`.

So what is alternative for `priorIPs = [geoip:cloudflare]` in your method?

RPRX:
比如如下配置
```json
"expectedIPs": ["geoip:us", "geoip:uk", "0.0.0.0/0", "::/0"],
"unexpectedIPs": ["10.10.34.0/24", "2001:4188:2:600:10:10:34:0/120"]
```

逻辑是
1. expect US IPs
2. but if unexpected
3. expect UK IPs
4. but if unexpected
5. allow all IPv4
6. allow all IPv6

patterniha:
1. This break current `expectedIPs` behavior. 

    If the IPs contain both uk and us, now we only have us ips.

2. Many changes need to be made in the code.

3. It is definitely more complicated than using 4 options

patterniha:
Let's do this:

I remove `priorIPs` and `unpriorIPs` options and just add `unexpectedIPs`.

Because we can use `redirect-socks-in/out` So these options are not necessary and are just confusing.

and don't change the current `expectedIPs" behavior

Ok?

RPRX:
> and don't change the current `expectedIPs" behavior

这个可以 change 啊，有什么问题吗，可以实现把更细分的 IP 范围放前面以优先匹配，比如 CF、US、ALL

patterniha:
It is 5 am in Iran now. I don't think there is a problem with your solution, but I have to think about it.

RPRX:
太困了会想不清楚，理解，不急，我是睡醒了才看的，~~没想到你这个点还没睡直接回复我了~~

patterniha:
@RPRX 

I think it is better remove `priorIPs` and `unpriorIPs` for now, and merge the PR.
And then open new PR for your changes
Do you agree?

RPRX:
@j2rong4cn 我刚看到 https://github.com/XTLS/Xray-core/pull/4666#issuecomment-2839728820 ，可以开个 PR

RPRX:
@patterniha 这个 PR 加的 `unexpectedIPs` 和删掉的 `allowUnexpectedIPs` 也是围绕 "expected" 服务的，所以就改这个 PR 吧

RPRX:
~~倒是可以把 `disableCache` 和 `finalQuery` 这俩和 "expected" 不相关的分两个 PR 出去~~

patterniha:
> For example, the following configuration
> 
> ```json
> "expectedIPs": ["geoip:us", "geoip:uk", "0.0.0.0/0", "::/0"],
> "unexpectedIPs": ["10.10.34.0/24", "2001:4188:2:600:10:10:34:0/120"]
> ```
> 
> The logic is
> 
> 1. expect US IPs, if we picked an IP, check all `unexpectedIPs`, if no IP was picked or the picked IP is "unexpected" then
> 2. expect UK IPs, if we picked an IP, check all `unexpectedIPs`, if no IP was picked or the picked IP is "unexpected" then
> 3. allow all IPv4, do not check `unexpectedIPs`
> 4. allow all IPv6, do not check `unexpectedIPs`

@RPRX 
There are a few problems:
1. You seperate 0.0.0.0/0 and ::/0 behavior from other ranges behavior. 
2. We can't have both IPv4 and IPv6, and only one of them can be returned, This is a serious problem for happy-eyeballs.

RPRX:
> You seperate 0.0.0.0/0 and ::/0 behavior from other ranges behavior.

这俩就是兜底的，前面所有流程走完了还没选定 IP 就返回第一个 IP，所以不适用 `unexpectedIPs`

> We can't have both IPv4 and IPv6, and only one of them can be returned, This is a serious problem for happy-eyeballs.

A 查询和 AAAA 查询本来就指定了 IP 类型，这里两种都写只是方便这个 DNS Server 配置可以同时 cover 两种情况

patterniha:
> A 查询和 AAAA 查询本来就指定了 IP 类型，这里两种都写只是方便这个 DNS Server 配置可以同时 cover 两种查询

or client/browser IP-request we have two distinct request for IPv4 and IPv6(pass through dns-proxy) so your changes does not affect this case .


But for happy-eyeballs we use domainStrategy= ForceIP and  ** we have one merged request for IPv4 and IPv6** but with your changes only one type of IPs is returned.


DNS behaves differently when we have two separate requests or one merged request.

RPRX:
改成 `, "*"` 吧，且它不适用 `unexpectedIPs`

~~\* 比 any 看起来更规范一些~~

patterniha:
@RPRX 

expectedIPs and unexpectedIPs should behave similarly.

In completion of what you said, I will say:

Let's make `expectedIPs` and `unexpectedIPs` two separate options.

if `"*"` is in `expectedIPs` list , it acts as a `priorIPs`, otherwise act as `expectedIPs`.

same for `unexpectedIPs`, if `"*"` is in `unexpectedIPs` list, it acts as a `unpriorIPs`, otherwise act as `unexpectedIPs`.

///

in fact, we just replace `allowUnexpectedIPs` with `"*"`.

patterniha:
in https://github.com/XTLS/Xray-core/pull/4666/commits/63cff2875eff302fa6ff0bf6297657c1e38fe1ff 

i add new `queryStrategy: UseSystem`.

built-in-dns should not return IPv6 for the system that not support IPv6 (unless user want use returned-IP at the other side), this causes several problems that I can explain if you want.

currently for this problem we can create two configs, one for IPv4-network and one for dual-stack-network, but it is not convenient to change the configs all the time.

so i add new `queryStrategy: UseSystem`,

when `queryStrategy` is `UseSystem`, every time dns-Query call, it check the system to see if it supports IPv6(and IPv4) or not, if it support IPv6(or IPv4), the IPv6(or IPv4) is also returned, otherwise not returned.

`queryStrategy: "UseIP" | "UseIPv4" | "UseIPv6" | "UseSystem"`

j2rong4cn:
> i add new `queryStrategy: UseSystem`.

@patterniha 为什么不用net.InterfaceAddrs()？https://pkg.go.dev/net#InterfaceAddrs

patterniha:
> > i add new `queryStrategy: UseSystem`.
> 
> @patterniha 为什么不用net.InterfaceAddrs()？https://pkg.go.dev/net#InterfaceAddrs

`net.dial("udp6", "[2001:4860:4860::8888]:53")` doesn't send any data(because it is udp), It just tries to bind a port, If successful, it indicates that the system has an interface that supports IPv6 and routing-table is also set correctly to use that interface.

because [2001:4860:4860::8888] is a real public-IP, we make sure that routing-table is also ok, and in practice we can bind a port for a public-IPv6 address.

///

but `net.InterfaceAddrs()`  only show information about interfaces, and it's possible that we can't really bind a port and connect to a public-IPv6 address.

in addition, `net.InterfaceAddrs()` also shows link-local and localhost addresses, but we should not consider them.

///

as a result, using udp-dial is definitely a better and more accurate way. (It also costs zero and only binds a port and then release)

patterniha:
in https://github.com/XTLS/Xray-core/pull/4666/commits/a1436665a6101e5f1d84494eb1d1360fe3df5410 i add `useSystemHosts`:

if true, system-hosts appends to config-hosts at start: https://github.com/XTLS/Xray-core/issues/4677

```
"dns": {
  ...
  "useSystemHosts": true
}
```

patterniha:
@Fangliding

if you reviewed the code and found no problem, Please tell @yuhan6665 and @RPRX to merge, thx.

patterniha:
@RPRX 

I had a goal (Serverless for Iran) and you helped me a lot to achieve my goal, but you neglected me near the end of my work.

~I also helped you and found and fixed nearly 20 small and big bugs in this time.~

**Serverless-for-iran-anti-sanction-version** is not yet complete and needs **these-dns-new-features** and **happy-eyeballs** to complete.

~Needless to say, these features are also very useful for Chinese users and other uses.~

Many users on Telegram ask me for Serverless-for-iran-anti-sanction-version, According to the tests I did, it opens almost all services and websites (except telegram) in all ISPs in iran.

Why does a PR take more than a month to be approved?

RPRX:
~~那我先合了，有要改的以后再改吧~~

Fangliding:
image
所以你检查系统是否支持ipv4v6的方法是向两个硬编码的ip进行dial????

patterniha:
@Fangliding 

I explained in https://github.com/XTLS/Xray-core/pull/4666#issuecomment-2851686886  why I used `net.dial` instead of `net.InterfaceAddrs()`.

It doesn't matter which IP we check for, It is enough to check for a non-reserved IP.

Meo597:
每次 dns 查询，你都额外 dial 两次

我的小霸王学习机受得了
用户手机也受不了

Meo597:
哦，草
不止两次

前脚刚在 LookupIP() 检查完
然后马上又在 QueryIP() 检查

然而 QueryIP() 只被前者调用
上次检查结果都传进去了的

人家原版的 QueryIP() 中再次检查是因为 Nameserver 也有自己的 v4 v6 option

也就是说每次 dns 查询至少无意义地 dial udp 四次
真的没什么开销吗？

~这不是 vibe coding 我吃屎三斤~

Meo597:
```go
package main

import (
	"net"
	"testing"
)

func checkSystemNetwork() (supportIPv4 bool, supportIPv6 bool) {
	conn4, err4 := net.Dial("udp4", "8.8.8.8:53")
	if err4 != nil {
		supportIPv4 = false
	} else {
		supportIPv4 = true
		conn4.Close()
	}

	conn6, err6 := net.Dial("udp6", "[2001:4860:4860::8888]:53")
	if err6 != nil {
		supportIPv6 = false
	} else {
		supportIPv6 = true
		conn6.Close()
	}
	return
}

func BenchmarkCheckSystemNetwork(b *testing.B) {
	for i := 0; i < b.N; i++ {
		checkSystemNetwork()
		checkSystemNetwork()
	}
}
```

---

```code-runner-output
BenchmarkCheckSystemNetwork-4   	     703	   1703517 ns/op	    4623 B/op	      46 allocs/op
```

开销还挺大

patterniha:
@Meo597 

This option is most commonly used on mobile-client-side, people are constantly switching between mobile-data and wi-fi.

in Iran, Most ADSL/VDSL/FTTH do not support IPv6, but IPv6 is supported in most places for mobile-data.

///

The cost of port-binding is almost zero, **less than 0.1ms per bind**, 

Also how many dns-lookup do we have per minute?

checking supported-IP-versions only once, causes users to have to turn VPN off and on after each network change.

This is an important feature for Serverless, please do not change that.

Meo597:
你不是糊了 happy eyeballs 吗？
这个就完全没用了

别告诉我你糊完了自己都没试一下

patterniha:
I add `useSystem`  queryStrategy for UDP not TCP.

if a system doesn't support IPv6, an IPv6 should not selected for freedom-UseIP/ForceIP.

Meo597:
1. 耗时 1.7ms 不是 0.17ms
2. 如果有 fallback 时间继续翻倍
3. 如果像你说的 wifi / 蜂窝 来回切，手机会按需请求 a or aaaa。ios 上只有 tun，请求来的都是 ip，android proxy 模式压根不支持 udp 然后 tcp 你已经有了 happyeyeballs 了。所以你这玩意只能是给 xary-core 内部用的。所以你这个东西的使用场景到底是什么？
4. 如果真像你说的确实手机需要，你至少应该让非移动端 sync.Once

Meo597:
别的都不提，你 QueryIP 为什么还要再查一次，你真的搞明白 dns 这一串函数的调用顺序了吗？

Meo597:
你这个功能纯粹就是臆想的糊上去
我提出问题了现编一个理由出来
你真的用过吗？

patterniha:
> iOS only has TUN, so requests always come in via IP

we have two situations:
1. not using `fakedns`:
   so if we return both IPv4 and IPv6 for browser dns-request, the browser usually choose IPv6 for quic, so if our network doesn't support IPv6, the quic-connection will not be established (in case of TUN, browser does not realize that IPv6 is not available, unless we disable IPv6 in TUN at all)
2. using `fakedns`:
   so we have domain in freedom and if we use "UseIP"/"ForceIP", an IPv6 may be chosen, and the quic-connection  will not be established.
   
///
   
I tested and it takes less than 50ms for 1000 bind-and-close(0.05ms per bind-and-close), so the cost is negligible.

Anyway if you still concern the cost, you can add `useSystemOnce`,
`useSystem` is the key-feature for my Serverless, and should not be changed.

Meo597:
我用 x86 跑，成本都是 1.7ms pre call，虽然是低功耗的 cpu
你 5w 手机秒了 25w PC 太厉害了
你是怎么做性能测试的？上面的 benchmark 跑跑看？

问题关键在于非移动端用它代价太大
你至少应该 PR 一个 patch 让它非移动端时 once

patterniha:
1.7ms only for port binding!!!, I think you measured other-things too (or something is wrong with your benchmark)

This has nothing to do with CPU-power, this is a simple thing.

please run:

```
    a := time.Now()
	for i := 0; i < 1000; i++ {
		c, _ := net.Dial("udp4", "8.8.8.8:443")
		c.Close()
	}
	println((time.Now().Sub(a)).Milliseconds())
```

and send the result.

Meo597:
原来你是这么做 benchmark 的
长见识了

imanabh (approved):

PR #4671, opened 2025-04-29
Improve random IP compatibility: support IPv4, add srcip option, and sync client source IP via sendthrough

ImAubrey:
Improve random IP compatibility: support IPv4, add srcip option, and sync client source IP via sendthrough


RPRX:
~~没看，合了，交给众测~~

PR #4673, opened 2025-04-29
DNS: Extend `hosts` Abilities

patterniha:
### number 1:
currently we cannot set `[]` for a domain in hosts.(the code return error)

but we may want to return no IP (EmptyResponse) for a domain, for example return no IP for ads-domains to block them:

```
"dns":{
    "hosts": {
      "geosite:category-ads-all": []
    }
}
```
so i extend ability to set `[]` or `""` for a domain when we want no IP to return for a domain (EmptyResponse)

### number 2:

suppose we want to return "216.239.38.120" for "youtube.com" but return "216.239.38.119" for other google sites, so we set:

```
"dns":{
    "hosts": {
      "youtube.com": ["216.239.38.120"],
      "geosite:google": ["216.239.38.119"]
    }
}
```

But the code accumulates the results, and we will eventually have:

`"youtube.com": ["216.239.38.120","216.239.38.119"]`

so we don't get the desired result.

**so we should not accumulate matched-results, We should return the first matched-result.**

**because each matched-result can have multiple IPs, so there is no need to accumulate the matched-results.**

**and accumulating the matched-results only causes such problems.**

as a result, now, the first matched-result is returned for a domain in hosts, Instead of accumulating all the matched-results.

### DIFF
--- a/app/dns/hosts.go
+++ b/app/dns/hosts.go
@@ -2,6 +2,7 @@ package dns
 
 import (
 	"context"
+	"sort"
 
 	"github.com/xtls/xray-core/common/errors"
 	"github.com/xtls/xray-core/common/net"
@@ -41,8 +42,6 @@ func NewStaticHosts(hosts []*Config_HostMapping) (*StaticHosts, error) {
 				}
 				ips = append(ips, addr)
 			}
-		default:
-			return nil, errors.New("neither IP address nor proxied domain specified for domain: ", mapping.Domain).AtWarning()
 		}
 
 		sh.ips[id] = ips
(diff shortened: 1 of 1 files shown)


Fangliding:
第一个可以加 隔壁那个rcode也行
第二个。。能的话还是尽量不想break老行为 https://github.com/v2fly/v2ray-core/pull/169 虽然也很久了
其他项目行为 singbox好像只允许address单映射 clashmeta没有复杂的匹配器 只能用通配 逻辑上是很自然的最长匹配优先 xray的地址映射到一组ip 最后append所有结果道理上说得过去 但是xray匹配器是最复杂的 combine的行为让它不那么好用 嗯。。。

patterniha:
@Fangliding 

As I said, appending-results only adds complexity, we can set multiple-IPs for each result, so there is no need to appending results, also, appending-results causes the problem I mentioned.

RPRX:
~~没看，合了，交给众测~~

PR #4702, opened 2025-05-06
DNS-Hosts: appending matched-results again

patterniha:
Unmarshalling json in golang will not respect order.

So, sorting `h.matchers.Match(domain)` in https://github.com/XTLS/Xray-core/pull/4673 is useless and the order of the results is still messed up, Because `json.Unmarshal` messed up the order of domains at the very beginning.

So, I temporarily reverted this to the previous state (appending matched-results).

///

I need to change `json.Unmarshal` for hosts to completely fix this issue.

### DIFF
--- a/app/dns/hosts.go
+++ b/app/dns/hosts.go
@@ -2,8 +2,6 @@ package dns
 
 import (
 	"context"
-	"sort"
-
 	"github.com/xtls/xray-core/common/errors"
 	"github.com/xtls/xray-core/common/net"
 	"github.com/xtls/xray-core/common/strmatcher"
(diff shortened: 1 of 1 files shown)


yuhan6665:
Can you do a proper `git revert`?

patterniha:
> Can you do a proper `git revert`?

No, it should not be reverted, the number-1 in https://github.com/XTLS/Xray-core/pull/4673 is OK.

yuhan6665:
Ok I see it now. Thanks!

patterniha:
> I need to change json.Unmarshal for hosts to completely fix this issue


Done: https://github.com/XTLS/Xray-core/pull/4706

PR #4703, opened 2025-05-06
Workflows: Authenticating the GitHub API call with GitHub token

yin1999:
Authenticating the GitHub API call with GitHub token to avoid [the ratelimte](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users). Referenced documentation: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

Tested in my fork: https://github.com/yin1999/Xray-core/actions/runs/14859797633/job/41721569355

image

### DIFF
--- a/.github/workflows/release-win7.yml
+++ b/.github/workflows/release-win7.yml
@@ -51,7 +51,7 @@ jobs:
           GOSDK=$(go env GOROOT)
           rm -r $GOSDK/*
           cd $GOSDK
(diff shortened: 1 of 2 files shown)


yuhan6665:
Thanks!

PR #4706, opened 2025-05-07
DNS-Hosts: return only first matched-result

patterniha:
we should not append matched-results in `hosts`, reason: https://github.com/XTLS/Xray-core/pull/4673#issue-3028985227

///

I tried to solve this problem in https://github.com/XTLS/Xray-core/pull/4673 but it didn't work, because `json.Unmarshal` had messed up the order and the first matched-result was not necessarily the first matched-result to appear in the json.

///

**but now I changed `json.Unmarshal` as well and the problem is completely solved.**

(In fact, if I knew I could solve "json.Unmarshal" problem so soon, I wouldn't have opened https://github.com/XTLS/Xray-core/pull/4702 at all)

### DIFF
--- a/app/dns/hosts.go
+++ b/app/dns/hosts.go
@@ -6,6 +6,7 @@ import (
 	"github.com/xtls/xray-core/common/net"
 	"github.com/xtls/xray-core/common/strmatcher"
 	"github.com/xtls/xray-core/features/dns"
+	"sort"
 )
 
 // StaticHosts represents static domain-ip mapping in DNS server.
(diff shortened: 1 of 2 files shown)


yuhan6665:
As discussed, we'd like to keep host config simple. Which means host config should just be a map, should not respect any order.
Any matching, fallback behavior should be move to routing or other DNS config. The previous example 
`"geosite:google": ["216.239.38.119"]`
is invalid.
If anyone has a strong use case, we can reopen.

patterniha:
This change was suggested by a telegram-user and I think it should be done, because we can have multiple-ips for each result, and there is no need for appending.
(Also i gave you a more strong use case in telegram)

///

Anyway, this is depends on opinion and it doesn't really matter.

What really matters is **"happy-eyeballs"** and **"dns-new-featutes"** PR, i solved are your problems and answered all your questions But after a month, they still haven't merged. 

I need them for **serverless-for-Iran-anti-sanction-versiom**.

Many users are waiting for this configuration and this configuration solves the problem of many users in iran.

Many users will actually no longer need to buy a VPN in iran.

They will also be useful for other uses and Chinese users will also benefit.

@RPRX  @yuhan6665  @Fangliding @KobeArthurScofield

PR #4713, opened 2025-05-09
Fix UDP original destination on windows

FDeghy:
When listening on 0.0.0.0 for UDP, incoming UDP packets on Windows now have the correct destination address, allowing the core to respond using the appropriate source address (OriginalDestinationAddress).
However, there is a bug: when the outbound tries to create a connection to a sniffed destination (e.g., youtube.com → 173.194.69.190), it mistakenly assumes that the IP is the destination address from the inbound side. As a result, it attempts to bind to 173.194.69.190 on the local system — but that address does not actually belong to the system.

### DIFF
--- a/proxy/dokodemo/fakeudp_other.go
+++ b/proxy/dokodemo/fakeudp_other.go
@@ -1,5 +1,5 @@
-//go:build !linux
-// +build !linux
+//go:build !linux && !windows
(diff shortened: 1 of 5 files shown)


Fangliding:
看懂这里是干嘛的吗就改。。。

PR #4721, opened 2025-05-12
Removing code that was not being executed and should not be executed.

patterniha:
during reading dispatcher codes, I encountered a strange code:

https://github.com/XTLS/Xray-core/blob/72170d8b6be0d83af1624f6d31b81cf73f1a189c/app/dispatcher/default.go#L410-L421

First, **fortunately** this code was not executed at all, because `ob.Target.String()` has prefix("tcp:"/"udp:") and `hosts.LookupHosts(ob.Target.String())` always returned `nil` and as a result `proxied` was always `nil` and the code was not executed at all !!!

Second, this code should not be executed at all, because the logic of this code is **completely wrong**, and we should not use `built-in-dns-hosts` to replace the `ob.Target` and change the destination, because suppose we are in client-side and we want to use `built-in-dns‍` only in client-side, and destination is used in server-side, so we should not change the destination and ob.Target by using `built-in-dns-hosts`, also if we have multiple-ip for a domain this code only replaces one IP.

///

fortunately, this code was not executed at all, and removing it is just to clean up the code.

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -96,7 +96,6 @@ type DefaultDispatcher struct {
 	router routing.Router
 	policy policy.Manager
 	stats  stats.Manager
-	dns    dns.Client
 	fdns   dns.FakeDNSEngine
 }
 
@@ -107,7 +106,7 @@ func init() {
 			core.OptionalFeatures(ctx, func(fdns dns.FakeDNSEngine) {
 				d.fdns = fdns
 			})
-			return d.Init(config.(*Config), om, router, pm, sm, dc)
+			return d.Init(config.(*Config), om, router, pm, sm)
 		}); err != nil {
 			return nil, err
 		}
@@ -116,12 +115,11 @@ func init() {
 }
 
 // Init initializes DefaultDispatcher.
-func (d *DefaultDispatcher) Init(config *Config, om outbound.Manager, router routing.Router, pm policy.Manager, sm stats.Manager, dns dns.Client) error {
+func (d *DefaultDispatcher) Init(config *Config, om outbound.Manager, router routing.Router, pm policy.Manager, sm stats.Manager) error {
 	d.ohm = om
 	d.router = router
 	d.policy = pm
 	d.stats = sm
-	d.dns = dns
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


Fangliding:
https://github.com/XTLS/Xray-core/commit/27224868aba63db8180ee2f2db71aa995930a74d#diff-c3820ebfbe4f1eb7e9d13008e995e0fa8912cad9f058c9f7767aed1df10533d8

tcp:/udp: 前缀似乎确实是没有意料到 但是实际上把它们添加到hosts中是可以被正确识别的

ro的判断有点奇怪 去掉的话似乎可以成为一种在客户端强制将域名解析成IP发送的方法？("tcp:ip.sb:80": "127.0.0.1" 这样的写法其实是可以被正常识别的 而且写成域名会被后续的dns解析器继续解析成IP)

yuhan6665:
I'm for removing deadcode.. It seems we can remove d.dns as well

patterniha:
@Fangliding 

first, this code only use `hosts` to replace the domain to IP, and does not use `dns-servers`, So it's not very practical.

second, if we want to replace the domain with IP, we can use `dialerProxy`, and if we want the IP to participate in routing, we can use `redirect-socks-in/out`.

third, you didn't mention `"tcp:ip.sb:80": "127.0.0.1"` for `hosts` in documentation.

anyway, even adding `"tcp:ip.sb:80": "127.0.0.1"` to documentation doesn't help much because we still can't use `dns-servers`, so it is better to use `dialerProxy` instead.

Fangliding:
just mention

Fangliding:
不过dns-servers的的确确可以被用到 "tcp:example.com:80": "example.com" 后续会被DNS继续解析成IP

yuhan6665:
等真的需要再加回来。。如果都同意这段代码没用那就先删了 删了总归是减小复杂度

Fangliding:
行 那我删了

yuhan6665:
Thanks both!

patterniha:
@yuhan6665 

@Fangliding  reviewed my code in https://github.com/XTLS/Xray-core/pull/4666 and https://github.com/XTLS/Xray-core/pull/4667, please add him to Reviewers to approve my code.

~I'm wonder how you still don't trust me after fixing about 20 small and big bugs‍ and adding many features to Xray-core~

PR #4726, opened 2025-05-14
Sniffer: Fix potential infinite loop

patterniha:
during reading sniffer code, i found two bugs and fixed them:

### 1. infinite-loop bug:

https://github.com/XTLS/Xray-core/blob/72170d8b6be0d83af1624f6d31b81cf73f1a189c/app/dispatcher/default.go#L376

if ` cacheErr != buf.ErrReadTimeout` (cacheDeadline not exceed) and `payload.IsEmpty()`, The code gets stuck in infinite loop, and CPU usage reaches 100%, and the app crashes !

(also if `cacheErr == nil` and `payload` doesn't change and `ErrProtoNeedMoreData`, the same thing happens)

**This bug is probably the main reason for https://github.com/XTLS/Xray-core/issues/3914.**

### 2. fakedns+others:

After reviewing the code, I realized that `"destOverride": ["fakedns+others"]` is no different from `"destOverride": ["fakedns"]`, That means `"destOverride": ["fakedns+others"]` behaves exactly like `"destOverride": ["fakedns"]`.

I also tested this to make sure.

PR #4733, opened 2025-05-16
Sniffer bug: `fakedns` alone should not override domain for tls/http/quic (when there is no map-domain)

patterniha:
currently, `"destOverride": ["fakedns"]` override domain for tls/http/quic protocol when ip is in fake-pool-range and when we have no map-domain.

This shouldn't happen and this is `fakedns+others` job.

///

also, it seems `fakedns+others` has no practical use and can be removed later.

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -224,7 +224,7 @@ func (d *DefaultDispatcher) shouldOverride(ctx context.Context, result SniffResu
 		if strings.HasPrefix(protocolString, p) || strings.HasPrefix(p, protocolString) {
 			return true
 		}
-		if fkr0, ok := d.fdns.(dns.FakeDNSEngineRev0); ok && protocolString != "bittorrent" && p == "fakedns" &&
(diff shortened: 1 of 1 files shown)


yuhan6665:
This is a feature I developed, not a bug. I had discussion with Fangliding.

patterniha:
> This is a feature I developed, not a bug. I had discussion with Fangliding.

@yuhan6665 @Fangliding 

1. documentation says:
`其中 ["fakedns+others"] 相当于 ["http", "tls", "quic", "fakedns"]`

But it is not equal to  ["http", "tls", "quic", "fakedns"]

2. Also, documentation says:
`当 IP 地址处于 FakeIP 区间内但没有命中域名记录时会使用 http、tls 和 quic 进行匹配。`

But this is not correct: https://github.com/XTLS/Xray-core/pull/4726#issuecomment-2882669110, https://github.com/XTLS/Xray-core/pull/4726#issuecomment-2882728850

///

The documentation says two contradictory things, neither of which is correct !!!

PR #4749, opened 2025-05-20
Increase HTTP/2 MaxReadFrameSize from 16KB to 1MB for improved efficiency

radioactiveAHM:
By default, the **MaxReadFrameSize** for the HTTP/2 client is only **16KB**, which is insufficient for high-bandwidth scenarios. In contrast, the default value for HTTP/3 is **6MB**, allowing for more efficient data handling.  

Increasing the **HTTP/2 MaxReadFrameSize** from **16KB to 1MB** enables larger frames to be processed in a single operation, resulting in:  
- **Reduced Overhead**: Minimizing excessive fragmentation of large payloads.  
- **Enhanced Performance**: Improving efficiency when handling high-bandwidth data streams.  

Making this value configurable in the settings file would be a valuable enhancement, but the **Xray** codebase is quite complex, making it difficult to determine the optimal placement for this configuration. Nevertheless, **setting 1MB as the default** is a reasonable choice for proxy applications.

### DIFF
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -209,8 +209,9 @@ func createHTTPClient(dest net.Destination, streamSettings *internet.MemoryStrea
 			DialTLSContext: func(ctxInner context.Context, network string, addr string, cfg *gotls.Config) (net.Conn, error) {
 				return dialContext(ctxInner)
 			},
-			IdleConnTimeout: net.ConnIdleTimeout,
-			ReadIdleTimeout: keepAlivePeriod,
+			IdleConnTimeout:  net.ConnIdleTimeout,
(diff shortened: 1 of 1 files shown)


darkk:
> By default, the MaxReadFrameSize for the HTTP/2 client is only 16KB

Is it true, though? To the best of my understanding, default HTTP/2 frame size is 1 MiB for Go HTTP/2 lirarby since 2014 at the very least: `defaultMaxReadFrameSize` is defined as 1MiB since https://github.com/golang/net/commit/21896bb2f9fea9683232786d7a81dced75e44518

Current Go http2 client uses the value with `configFromTransport()` calling to `setConfigDefaults()` that fills it. It's configurable at least since #1409.

PR #4779, opened 2025-06-04
REALITY protocol: fallback for chrome fingerprint (X25519MLKEM768)

hossinasaadi:
If the client is using version v25.5.16 and the server is on an older version, some Reality configurations using targets that don't support `X25519MLKEM768` will fail. To address this, I've added a temporary `chromeNormal` fallback, which will remain in place for the next few versions until server side's updated to a supported version.

https://github.com/XTLS/Xray-core/pull/3813#issuecomment-2919224335

### DIFF
--- a/transport/internet/reality/reality.go
+++ b/transport/internet/reality/reality.go
@@ -245,6 +245,14 @@ func UClient(c net.Conn, config *Config, ctx context.Context, dest net.Destinati
 			// Do not close the connection
 		}()
 		time.Sleep(time.Duration(crypto.RandBetween(config.SpiderY[8], config.SpiderY[9])) * time.Millisecond) // return
+
+		// temporary fallback for server core that not upgrade to X25519MLKEM768 yet
+		if config.Fingerprint == "chrome" {
+			config.Fingerprint = "chromeNormal"
+			if config.Show {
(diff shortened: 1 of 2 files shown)


Fangliding:
这是故意为之的 旨在督促服务端尽快升级
https://t.me/projectXray/4294992

hossinasaadi:
> 这是故意为之的 旨在督促服务端尽快升级 https://t.me/projectXray/4294992

I agree, but think core should support this for a while.

yuhan6665:
Maybe, use hellochrome_120

hossinasaadi:
> Maybe, use hellochrome_120

chromeNormal is hellochrome_120
Can you explain more?

yuhan6665:
Can you confirm this is the current behavior?
New client + old server + old dest = working
New client + old server + mlkem dest = error
Any other combination = working

Fangliding:
旧的reality服务端没法解析带pq的chrome131
btw就算要改 应该是把default改成旧的120

hossinasaadi:
> New client + old server + old dest = working

done: New client + New server + old dest = working
done: New client + old server + old dest = working
done: New client + New server + New dest = working
done: Any other combination = working
couldn't test (New client + old server + mlkem dest = error), as @Fangliding  said, it should fail

Issue #4791, opened 2025-06-07
Irresponsive UDP transparent proxy


PoneyClairDeLune:
### Integrity requirements

- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

At least on Linux, ever since at least Xray 1.8.x (should be even earlier), the UDP transparent proxy inbounds have always been irresponsive. Whenever a UDP packet is tunnelled in the transparent proxy, only the first packet gets through, while all subsequent packets get seemingly dropped.
All other types of inbounds, including SOCKS5, do not exhibit the same problematic behaviour. V2Ray v4 never exhibited the same problematic behaviour either, as such I'm currently relying on it for UDP transparent proxies. TCP connections get handled fine as well.

### Prerequisites
- A functioning server & client config.

### Steps
- Create a new pair of transparent proxy inbounds to the existing client config.
- Start Xray.
- Utilize anything that starts UDP connections, e.g. initiating a DNS request via `dig` or `nslookup`.

### Client config

<details><pre><code>
{
	"tag": "i-tproxy-4",
	"listen": "127.0.0.1",
	"port": 10811,
	"protocol": "dokodemo-door",
	"settings": {
		"followRedirect": true,
		"network": "tcp,udp"
	},
	"sniffing": {
		"destOverride": ["http", "tls"],
		"enabled": true,
		"routeOnly": true
	},
	"streamSettings": {
		"sockopt": {
			"tproxy": "tproxy"
		}
	}
}, {
	"tag": "i-tproxy-6",
	"listen": "::1",
	"port": 10811,
	"protocol": "dokodemo-door",
	"settings": {
		"followRedirect": true,
		"network": "tcp,udp"
	},
	"sniffing": {
		"destOverride": ["http", "tls"],
		"enabled": true,
		"routeOnly": true
	},
	"streamSettings": {
		"sockopt": {
			"tproxy": "tproxy"
		}
	}
}
</code></pre></details>

- `routeOnly` may not work on V2Ray v4, but I only added that for Xray anyway.

### Server config

N/A

### Client log

```
[...]

PR #4795, opened 2025-06-08
UDP: update last-active-time even when request-data is received

patterniha:
https://github.com/XTLS/Xray-core/issues/4791#issuecomment-2954298022

fix https://github.com/XTLS/Xray-core/issues/4791


RPRX:
对于 UDP 来说，没有回应就应当按不通处理

Fangliding:
你是怎么做到臆想一下bug是什么还没验证没问题就开pr的

patterniha:
@RPRX 

User may use a udp-base-protocol that have separate downstream and upstream connection, so ack-packets may received only from downstream-connection.
 As a result, we should update timer even when sending-data.(otherwise upstream-connection fail)

patterniha:
@RPRX @Fangliding 

Also why is it fixed at 60 seconds and `policy-connIdle` has no effect on UDP ?

Issue #4800, opened 2025-06-10
Freedom outbound doesn't work properly when connection is UDP and address is a domain


patterniha:
during writing serverless-for-Iran-config I encountered this bug, and unfortunately, my work was delayed again.

### Reproduction Method:

1. open **New-incognito-tab** in chrome-android, and go to "cloudflare-quic.com"
2. refresh the page until "HTTP/3" is shown, this show that "cloudflare-quic.com" and "quic" protocol is not block, and everything is working without VPN.
3. import [this config](https://github.com/patterniha/bug-test/blob/main/UDP-BUG-fakedns.json) to v2rayNG and turn on the VPN
4. refresh the page, the "HTTP/2" is shown, indicating that the UDP connection is down.
5. even if you close chrome and open New-incognito-tab and go to "cloudflare-quic.com" again and refresh, "HTTP/2" is still shown, i try this procedure many times, but as long as the VPN is on, "HTTP/3" is not displayed.
6. after turn off VPN (and close chrome and open new-incognito-tab) everything returns to normal and "HTTP/3" is displayed.

///

### Tips:

1. you should open incognito tab instead of normal tab, otherwise chrome-cache affects the results.
2. if you replace `fakedns` with `1.1.1.1` and turn off sniffing, The problem will not occur: https://github.com/patterniha/bug-test/blob/main/NO-BUG-no-sniff.json
3. or if you use `"redirect": "188.114.98.0:0"` (or other cloudflare-ip), the problem will not occur: https://github.com/patterniha/bug-test/blob/main/NO-BUG-redirect.json

### Conclusion:

the Reproduction-Method and Tips2-3 indicates that something is wrong when connection is UDP and address is domain and outbound is freedom (probably something is wrong with `freedom.go`)

Fangliding:
核心可以正常处理目标为域名的UDP流量(包括QUIC)

![Image](https://github.com/user-attachments/assets/d3c024ce-e31b-4e80-967e-40617838c1f4)

也许你应该去问chrome为什么不愿意使用QUIC。如果你找到问题了的确与xray有关那或许可以进行一些修改，但是这很明显超出了这里应该处理的事情

patterniha:
@Fangliding 

I try to solve this problem myself.

do you test my reproduction-method?

dyhkwong:
This should be related to #3351, it it converted from issue to discussion and apparently moderators don't want to further discuss about it.

patterniha:
[...]

PR #4804, opened 2025-06-10
Freedom: Cache UDP resolve result

Fangliding:
https://github.com/XTLS/Xray-core/issues/4800#issuecomment-2959273169
道理上udp就不该有fqdn当目标(((((


Fangliding:
早看到了 已经改了
不过我回头问了一下 这和你遇到的东西没关系 这只是一个小小的refine

patterniha:
@Fangliding 

line 380 and 381 should be swapped

patterniha:
@Fangliding 

line 389 and 390 should be swapped

Fangliding:
这两行换不换有什么区别么

patterniha:
yes ip.domain() is panic

patterniha:
> 道理上udp就不该有fqdn当目标(((((

@Fangliding 

or we have **multiple-IP** or we have **only-one-domain** on single-UDP.

if we have multiple-domain it is impossible to connect response to request.

for example if  we have two domain example-1.com and example-2.com, and both resolved to "2.2.2.2"  so if we receive response from "2.2.2.2" this is for example-1.com or example-2.com ???

**this is logically impossible, and has nothing to do with code.**

///

So we never have multiple-domain on single-UDP and there is no need for `resolvedUDPAddr-map`, only single variable is enough(of course, I haven't deleted it from my PR yet.)

///

Meo597:
`TestCommanderListHandlers` 极其容易炸

rebase 到 v25.6.8 `fbae89d017aee5947598d0a82945b316fdc688ab` 就好了

patterniha:
@Fangliding 

~I separated my PR from your PR.‍~

this PR fix `WriteMultiBuffer` and [my PR](https://github.com/XTLS/Xray-core/pull/4816) fix `ReadMultiBuffer`, and these two PR will solve the UDP problem in Freedom when address is domain or fakedns.

also, you missed a few things in your code:

1. even if `hasStrategy` is false, we should load and store from cache.

2. if `ForceIP` is true, and `built-in-DNS` fail to resolve-IP, we should not fallback to system-DNS (`net.ResolveUDPAddr`)


////////////////////

@RPRX 

these two PR should merge as soon as possible, and new version should be released.

I think you are aware of the sensitive situation in Iran. we are at war and the internet is very limited.

RPRX:
~~话说现在伊朗基本断外网了还能 Serverless-for-Iran 吗~~

patterniha:
> ~Now that Iran is basically disconnected from the Internet, is Serverless-for-Iran still available?~

only youtube ~and porn sites~ (of course, I have to change the DNS, I'm waiting for the freedom-UDP problem to be resolved)

///

but the reverse(bridge-portal) method is working great on home-internet with static-IP.

patterniha:
The internet is back to normal.
but ipv6 and UDP are still not available.

patterniha:
> 
> also, you missed a few things in your code:
> 
> 1. even if `hasStrategy` is false, we should load and store from cache.
> 2. if `ForceIP` is true, and `built-in-DNS` fail to resolve-IP, we should not fallback to system-DNS (`net.ResolveUDPAddr`)
> 
> /////////////////

@Fangliding 

Fix 1 and 2, thx.

Fangliding:
It's done

patterniha:
Two new problems occurred:

1. if `b.UDP.Address.Family().IsDomain()` is `false` in line 385, `ip` is always `nil` and we don't send any data!!!(you need to change `if ip == nil` scope)

2. suppose in `b.UDP.Address, _ = w.resolvedUDPAddr.LoadOrStore(b.UDP.Address.Domain(), ip)` we have `Load` instead of `Store`, so we should always use b.UDP.Address as final-address, but you set `IP:   ip.IP()`(you should use `IP: b.UDP.Address.IP()` instead)

Fangliding:
算了还是改成之前那样吧 我只是不想把continue写两次 
这个use真是烦死了。。

patterniha:
@Fangliding 

i tested but it didn't work, after seeing the code, I realized the problem:

in line 336 you define simple map `resolvedUDPAddr := make(map[string]net.Address)` 

but in line 347 you don't use it and set `resolvedUDPAddr:   utils.NewTypedSyncMap[string, net.Address]()` !!!

so initial-set-map has no effect!!!

///

also it is better to use `targetAddr.Domain()` instead of `targetAddr.String()`, because you use `Domain()` for other usage line 385, 390, 410 (although they are equal).

patterniha:
> Fix 0 len outbounds 

How can the length of outbounds be zero?

And if it is zero, ‍‍`NewPacketWriter` will not be executed at all:

we have:

https://github.com/XTLS/Xray-core/blob/1976d02ec9aaff3d34221a18b4ec8869437e2dc6/proxy/freedom/freedom.go#L105-L107

before that and it will panic.

Fangliding:
> > Fix 0 len outbounds
> 
> How can the length of outbounds be zero?
> 
> And if it is zero, ‍‍`NewPacketWriter` will not be executed at all:
> 
> we have:
> 
> https://github.com/XTLS/Xray-core/blob/1976d02ec9aaff3d34221a18b4ec8869437e2dc6/proxy/freedom/freedom.go#L105-L107
> 
> before that and it will panic.

我这么写可能清楚点 那个ctx可能不是上面看到的ctx 你可以自己用先前的版本试试 就是我在测你昨天那个pr的时候遇到的

patterniha:
i tested and see `len(outbounds)` is always `0` and your initial-set-map does not perform at all.

NewPacketWriter/NewPacketReader ctx is newCtx define in line-174 and has no outbounds.

so we should copy ctx-outbounds to newCtx-outbounds.

///

I fix it in https://github.com/XTLS/Xray-core/pull/4816.

Fangliding:
就这样就行了 不要破坏timeoutonly的语义

patterniha:
> Don't destroy the semantics of timeoutonly

I don't change that, I just copy ctx-outbounds to newCtx-outbounds, please revert your last commit, thx.

patterniha:
~Forget what I said, I hadn't seen your last force-pushed~

PR #4805, opened 2025-06-10
cache update after initial lookup

patterniha:
@Fangliding 

thank you very much for solving this problem. This was very necessary, Otherwise, we couldn't use fakedns or quic-sniffing.

currently, in UseIP/ForceIP mode we have two dns-lookup for each udp-connection, but we should have just one.

so the cache should update after initial-lookup.


Fangliding:
https://github.com/XTLS/Xray-core/pull/4804/commits/f4246e9314a610596291061613c587aeafbbf3c2
这样就够了 但是我还得再强调一遍 这和你遇到的东西没有关系

patterniha:
this is exactly i need

PR #4812, opened 2025-06-12
.

patterniha:
### the problem:

the problem https://github.com/XTLS/Xray-core/issues/4800#issue-3132409092 is because after browser sending UDP-data(quic initial packet) the packet correctly reaches to the final-target but the response-data does not reach to the browser, this problem happen when target-address is domain, let's explain why:

**code-A (freedom.go > ReadMultiBuffer):**

https://github.com/XTLS/Xray-core/blob/fbae89d017aee5947598d0a82945b316fdc688ab/proxy/freedom/freedom.go#L309-L313

**code-B (udp/dispatcher.go > handleInput):**

https://github.com/XTLS/Xray-core/blob/fbae89d017aee5947598d0a82945b316fdc688ab/transport/internet/udp/dispatcher.go#L135-L138

**code-C (socks/server.go):**

https://github.com/XTLS/Xray-core/blob/fbae89d017aee5947598d0a82945b316fdc688ab/proxy/socks/server.go#L237-L243

suppose browser send UDP-socks-request(UDP-associate) and target is cloudflare-quic.com:443.

each request packet consists of header+payload and header is "cloudflare-quic:443".

the response packet is also consists header+payload, **the response-header must also be "cloudflare-quic.com", otherwise browser does not accept the received data**.

to sending packet to "cloudflare-quic.com", Xray-core must resolve it to IP, suppose resolved-IP is "188.114.98.0".
after sending data, `ReadMultiBuffer` function receive response-data and because packet received from "188.114.98.0", it set `b.UDP`(buffer-UDP) address to "188.114.98.0" ---> **code-A**

after udp-dispatcher call a "callback" and pass the buffer(`b`) to that  ---> **code-B**

if for example inbound is socks-protocol the 'callback" is "udpServer" in "socks/server.go > handleUDPPayload".

because `payload.UDP` is not `nil` and is equal to "188.114.98.0:443", **the response-header-address overridden by "188.114.98.0:443", but the request-header was "cloudflare-quic.com" so the browser reject the response.** ---> **code-C**

### In short:

**in short, the request and response header must be same but if address is domain(or fakedns) Xray-core send resolved-IP as a response header and this causes the browser to reject the packet.**

**so if the address is domain or fakedns or changed by `redirect` settings, Xray-core should ignore response-IP-address and send request-header-address as a response-header-address**.

////////////////////////////////////////////////////////////////////////////////

### other changes:

1. there is no difference between sending dns-query to-built-in-dns or system-dns for using resolvedUDPAddr-cache, so even for AsIs-mode we should use cache.
2. if domainStrategy is ForceIP and query to built-in-dns fail, we should not fallback to system-dns.
3. we should use thread-Lock to avoid RaceCondition for resolvedUDPAddr-map.

///

this should be merge to udp-resolve, and then udp-resolve merge to main.

### DIFF
--- a/proxy/freedom/freedom.go
+++ b/proxy/freedom/freedom.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/pires/go-proxyproto"
@@ -238,7 +239,7 @@ func (h *Handler) Process(ctx context.Context, link *transport.Link, dialer inte
 		if destination.Network == net.Network_TCP {
 			reader = buf.NewReader(conn)
 		} else {
-			reader = NewPacketReader(conn, UDPOverride)
+			reader = NewPacketReader(ctx, conn, UDPOverride)
 		}
 		if err := buf.Copy(reader, output, buf.UpdateActivity(timer)); err != nil {
 			return errors.New("failed to process response").Base(err)
@@ -273,7 +274,7 @@ func isTLSConn(conn stat.Connection) bool {
 	return false
 }
 
-func NewPacketReader(conn net.Conn, UDPOverride net.Destination) buf.Reader {
+func NewPacketReader(ctx context.Context, conn net.Conn, UDPOverride net.Destination) buf.Reader {
 	iConn := conn
 	statConn, ok := iConn.(*stat.CounterConnection)
 	if ok {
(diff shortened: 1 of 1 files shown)


Fangliding:
。。。你有试过它可以工作吗？你不会都是等更改合进主线再等v2rayng更新了再去试的吧
关太多了再关要被说了，我也不浪费时间扯了，其他人也看看吧。

PR #4818, opened 2025-06-15
API: Fix issue with inbounduser not finding emails with uppercase letters

fL1pSt3r:
Fixes #3644. 
In the VLESS and Trojan protocols, GetByEmail did not account for the fact that emails are stored in lowercase, so the `api inbounduser` could not find a user if the email was written using uppercase letters.
image


Fangliding (approved):

yuhan6665:
Looks good! Thanks

PR #4840, opened 2025-06-24
VLESS fallbacks: `dest` defaults to "127.0.0.1" -> "localhost"

xqzr:
Fix https://github.com/XTLS/Xray-examples/issues/234


RPRX:
https://github.com/XTLS/Xray-examples/issues/234#issuecomment-3091319391

PR #4855, opened 2025-07-03
Reverse: add `maxConnections` and `maxConcurrency` and increase heartbeat period

patterniha:
The reverse method was one of the methods that worked well in the Iran-Israel 12-Day War and is still a good option.

[It is easy and cheap to get an IP-static or iran-VPS in Iran]

///

currently the max-concurrency-ratio is set to constant 16:

https://github.com/XTLS/Xray-core/blob/cb1afb33e68c56bf4510ad658ab86a0468b8c36c/app/reverse/bridge.go#L75

also, after creating each bridge-worker, it remains until xray-core is restarted and if we don't use mux in our outbound each worker create a connection to portal-server.

also, heartbeat period for each worker is set to 2 seconds.

///

so suppose we have for example 100 connection from portal for a while, so 7 (100/16) worker is created and each worker has a 2 seconds heartbeat-period.
**as a result we send a heartbeat packet every ~300 milliseconds and after the number of connections decreased, we still have 7 workers(->connections) and we send a heartbeat packet every ~300 milliseconds.**

**And this causes waste and reduced performance.**

///

*** **also, we need to be able to limit the number of workers, because After increasing the number of workers, there is no way to reduce them except by restarting xray-core.**

///

### what changed:

1. heartbeat-period increased to 10 seconds.

2. add `MaxConnections` and add `MaxConcurrency` option to bridge-config.(control number of workers)
 
They act like xhttp-xmux-MaxConcurrency/MaxConnections (both can't be greater than zero and if both is zero `MaxConcurrency` is set to 16)
 
The only difference is that ‍‍`MaxConcurrency` control ratio-concurrency  instead of exact-concurrency (because connection created in reverse manner and we can't control over exact-concurrency)


Fangliding:
数据包是这么算的吗 300毫秒一个几十比特在包在你这变成大问题了
这个东西的作用不是保活 是传递draining的控制信息啊
以及xmux的两个max参数是互斥的

patterniha:
I don't change drain-check-period and it is still 2 seconds, I just increase heartbeat to 10 seconds, 2 seconds is too short for heartbeat.

///

> And the two max parameters of xmux are mutually exclusive.

I knew and i added check:

```
if c.MaxConnections > 0 && c.MaxConcurrency > 0 {
		return nil, errors.New("maxConnections cannot be specified together with maxConcurrency")
	}
if c.MaxConnections == 0 && c.MaxConcurrency == 0 {
		c.MaxConcurrency = 16
	}
```

**After increasing the number of workers, there is no way to reduce them except by restarting xray-core, so we need to be able to limit the number of workers.**

patterniha:
anyway if you think heartbeat-period should be 2 seconds, i can revert. but there is no need for 2 seconds heartbeat.
drain-check-period is still 2 seconds i didn't change that.

Fangliding:
我的评价是只把16改成可配置的就行了 ~~别的保不齐会出问题。。~~ 连接是会自己回收的
似乎还好

Fangliding:
再把portal的draining也设置为可控吧 现在默认256 只要设置一个比较低的值就能让服务端快速放弃多开的连接了 加快回收速度 或者给portal设置一个MinConcurrency 如果一个worker不到这个数就直接分配给它 不再取最低的 也能加速服务端放弃多出来的连接

patterniha:
> Then set the portal's draining to be controllable. The default value is 256. Just set a relatively low value to allow the server to quickly abandon the multiple connections to speed up the recycling speed. Or set a MinConcurrency for the portal. If a worker does not reach this number, directly assign it to it instead of taking the lowest number, which can also speed up the server to abandon the extra connections.

also, When the number of connections for a worker reaches zero (and we have at least one another worker), portal should close that otherwise it remain open forever.

Fangliding:
mux worker会16秒检查一次的 如果没有活跃子连接应该会自行关闭才对

patterniha:
> The mux worker will check every 16 seconds and if there is no active sub-connection, it should close itself.

But that doesn't help, **each portal-worker always have at least one sub-connection**:

https://github.com/XTLS/Xray-core/blob/cb1afb33e68c56bf4510ad658ab86a0468b8c36c/app/reverse/portal.go#L226-L229

this sub-connection is used to send heartbeat packet.

///

so the number of sub-connections for each portal-worker will never be zero, **and each worker remains forever**.

This causes us always have a bunch of unused connections.

///

i will fix it soon.

patterniha:
After thinking about it, the main problem is: worker does not close after no active sub-connection,

so I think fixing this and increase heartbeat-period is enough.

PR #4861, opened 2025-07-06
MUX: Refine and Fix some occasional problems

patterniha:
https://github.com/XTLS/Xray-core/blob/cb1afb33e68c56bf4510ad658ab86a0468b8c36c/common/mux/client.go#L291-L297

**if this function run concurrent from two goroutine, some problems may arise, let explain why:**

suppose `m.strategy.MaxConcurrency` is set to 4, and currently we have 3 sessions for `m.sessionManager`.

suppose goroutine-A and goroutine-B run this function concurrency, because we have 3 sessions and `MaxCuncurrency` is set to 4, `m.IsFull()` is `false` for both goroutine, so both goroutine will run `sm.Allocate()`.

although `sessionManager` has it's own Lock, but it does not help in this situation, and `sm.Allocate()`  create a new session for each goroutine. 

**so after running this two goroutine two new sessions created and the number of sessions reach to 5, but `MaxConcurrency` is set to 4, so we exceeded the value of `MaxConcurrency` !!!**

///

**so we need another `sync.Mutex` lock for running this dispatch, to  make sure we never exceed the `MaxConcurrency`/`MaxConnection`.**

also, `m.IsFull()` contain `m.Closed()`, so there is no need to check it again.

### DIFF
--- a/common/mux/client.go
+++ b/common/mux/client.go
@@ -296,7 +296,7 @@ func (m *ClientWorker) Dispatch(ctx context.Context, link *transport.Link) bool 
 	}
 
 	sm := m.sessionManager
-	s := sm.Allocate(m.strategy)
+	s := sm.Allocate(&m.strategy)
 	if s == nil {
 		return false
 	}
--- a/common/mux/client.go
+++ b/common/mux/client.go
@@ -174,7 +174,6 @@ type ClientWorker struct {
 	link           transport.Link
 	done           *done.Instance
 	strategy       ClientStrategy
-	access         sync.Mutex
 }
 
 var (
@@ -277,6 +276,8 @@ func (m *ClientWorker) IsClosing() bool {
 	return false
 }
 
+// IsFull returns true if this ClientWorker is unable to accept more connections.
+// it might be because it is closing, or the number of connections has reached the limit.
 func (m *ClientWorker) IsFull() bool {
 	if m.IsClosing() || m.Closed() {
 		return true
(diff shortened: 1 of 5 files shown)


Fangliding:
这样可以省掉一把锁的样子 不会损失任何性能

patterniha:
@Fangliding 

we should also care about `MaxConnection`, so add it too.

Fangliding:
都是3没有问题 allocate是原子的

patterniha:
@Fangliding 

you should pass `m.Strategy` by reference -> `*m.Strategy`.

otherwise it is possible both goroutine use `3`.

Fangliding:
没看懂你在表达啥 use `3` 的 `3` 是什么

Fangliding:
那就让它copy不就好了 反正这个值又不会变 是个常量 只会读而已

patterniha:
you are right, i just confused for a moment.

anyway, there is no need to revert my last fix.

Fangliding (approved):

patterniha:
i change `server.go` base on our discussion in https://github.com/XTLS/Xray-core/pull/4876.

I only added the minimum changes required, please check, thx.

patterniha:
@Fangliding 

If you see a problem with `server.go`, you can change it yourself.

Which part of my changes is problematic?

patterniha:
Screenshot 2025-07-09 193732

@Fangliding 

I tested, if we don't close writer explicitly, it takes a few seconds to close the connection.
this is because of `policy` settings.

**but all sessions is closed and there is no data to send, so it is useless to wait before closing the connection.**

also, if `ctx.Done` is happen, the connection is remain open.

///

also, returning `s.Close(false)` instead of `buf.Copy(rr, buf.Discard)` in line 238, is definitely the author's mistake.

Fangliding (approved):

PR #4723, opened 2025-05-13
feat(api): Add ListInbounds and ListOutbounds to the gRPC API 

gsergey418:
Fixes #2745. Add the ListInbounds and ListOutbounds methods to the gRPC HandlerService and write the respective CLI commands lsi and lso.

PR #4862, opened 2025-07-06
feat: add new method for get only inbound tags from core. ListTags.

Jolymmiles:
Add new method to get only tags from core instead full inbounds body with a lot of users.

https://github.com/XTLS/Xray-core/pull/4723#issuecomment-2952583032


Jolymmiles:
@Fangliding Could you look pls?

Fangliding:
我不觉得listTags有什么很大用处 这个command列表已经有点乱了

kastov:
> 我不觉得listTags有什么很大用处 这个command列表已经有点乱了

So it is okay that similar method exists already? I mean this PR https://github.com/XTLS/Xray-core/pull/4723.

It can be used for similar purposes, but... it is much simpler and much cheaper to get just TAGS without all that stuff like all users, which is not needed mostly

Fangliding:
正是因为已经有了类似的所以我才觉得没必要再加一个这样的 而且那个pr的功能更全面 而且我确实没有想到列出tag的意义在哪

Fangliding:
否则会导出很多类似的乱七八糟的东西 比如输出dns服务器tag 路由规则tag 诸如此类。。 不可能给每一个小项单开一个API

Jolymmiles:
@Fangliding ok, but what if it is an option of an existing method?

PR #4888, opened 2025-07-12
Simple Dockerfile & .dockerignore

hoodadt:
open to suggestion or critics
add minimal Dockerfile for xray-core build

- Multi-stage build using golang:1.24-alpine
- Module caching via go.mod/go.sum separation
- Installs binary into minimal alpine:3.22
- Includes optimized .dockerignore for clean builds

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,26 @@
+# Binaries and build artifacts
+*.exe
+*.out
+*.test
+*.tmp
+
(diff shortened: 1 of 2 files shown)


Fangliding:
仓库里已经有一个dockerfile了
https://github.com/XTLS/Xray-core/tree/main/.github/docker

PR #4902, opened 2025-07-16
README.md: Add AnyPortal to GUI Clients

AnyPortal:
Introducing AnyPortal, a GUI client for Windows, Linux, macOS, and Android.
Xray is embedded for Android.

The list is getting long, so we suggest some kind of sorting. It would be a conflict of interest to have the list sorted in alphabetical order, so we did not do that.
For now, AnyPortal is added to the end of each supported list.


RPRX:
~~X-flutter 被夺舍了？~~

PR #4909, opened 2025-07-19
UDP: faster close (in some rare situations)

patterniha:
https://github.com/XTLS/Xray-core/pull/4899#issuecomment-3092141586

### DIFF
--- a/app/proxyman/inbound/worker.go
+++ b/app/proxyman/inbound/worker.go
@@ -161,6 +161,7 @@ type udpConn struct {
 	uplink           stats.Counter
 	downlink         stats.Counter
 	inactive         bool
+	cancel           context.CancelFunc
 }
 
 func (c *udpConn) setInactive() {
@@ -203,6 +204,9 @@ func (c *udpConn) Write(buf []byte) (int, error) {
 }
 
 func (c *udpConn) Close() error {
+	if c.cancel != nil {
+		c.cancel()
+	}
 	common.Must(c.done.Close())
 	common.Must(common.Close(c.writer))
 	return nil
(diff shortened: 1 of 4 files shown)


Fangliding:
那么多没合呢要不你还是消停会

patterniha:
udpconn is closed after 2 minutes of inactivity, suppose exactly after 2 minutes a packet is received from inbound.

**because udpconn has no Lock it is possible udpconn is closed while this data passes to udp-dispatcher.**

so new-outbound-connection is created while udpconn is closed, and this connection remain open wrongly.

because target-server can't receive any data, it doesn't send any data after a while and after 60 seconds the wrong-outbound-connection is closed.

///

If you think rare bugs shouldn't be fixed, I have nothing to say to you.

patterniha:
udpconn is closed after 2 minutes of inactivity, suppose exactly after 2 minutes a packet is received from inbound.

**so is possible udpconn is closed while this data passes to udp-dispatcher.**

~possible is possible, even with a very low probability~

so new-outbound-connection is created while udpconn is closed, and this connection remain open wrongly.

because target-server can't receive any data, it doesn't send any data after a while and after 60 seconds the wrong-outbound-connection is closed.

///

If you think rare bugs shouldn't be fixed, I have nothing to say to you.

patterniha:
also, If this happens, we handle the new received packet incorrectly, because we create a new  outbound connection but we ignore it !

so we should run `conn.updateActivity()` in udpWorker lock:

```
if conn, found := w.activeConn[id]; found && !conn.done.Done() {
		conn.updateActivity()
		return conn, true
}
```

patterniha:
this PR doesn't solve the main-problem, we should ‍ run `updateActivity()‍` ‍‍in lock, so there is no need for this PR at all.

patterniha:
I managed to produce an example(with two interface) where udpConn is closed but connEntry is not closed, so this PR should be merged to fix this (rare) situation.

PR #4911, opened 2025-07-19
Vless fallback: we shoud not create fallback-connection after inbound-connection is closed

patterniha:
first do the following test to understand better:

1. run [this xray config](https://github.com/patterniha/bug-test/blob/main/vless_fallback_bug_proof.json)

2. sniff loopback port 8888 with wireshark 

3. run [this](https://github.com/patterniha/bug-test/blob/main/vless_fallback.py) python script
this script create a connection to xray-vless-inbound and close the connection after 1 seconds

you see after inbound-connection is closed, xray create a connection to fallback and then close it immediately!!!

this is just a waste and should be fixed.

Screenshot 2025-07-19 170558

### DIFF
--- a/proxy/vless/inbound/inbound.go
+++ b/proxy/vless/inbound/inbound.go
@@ -206,7 +206,10 @@ func (h *Handler) Process(ctx context.Context, network net.Network, connection s
 
 	first := buf.FromBytes(make([]byte, buf.Size))
 	first.Clear()
-	firstLen, _ := first.ReadFrom(connection)
+	firstLen, errR := first.ReadFrom(connection)
+	if errR != nil {
+		return errR
+	}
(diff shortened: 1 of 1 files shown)


Fangliding:
等其他的合并吧别发这些有的没的了 谁没事打堆syn flood

patterniha:
What does it have to do with syn-flood?
in many situations connection is closed before receiving any data.

CC @RPRX

patterniha:
all other protocols trojan/socks... return error when reading the first data is unsuccessful:

https://github.com/XTLS/Xray-core/blob/83686ebfaa56666b7cf95d31cc53990ef9c57e1f/proxy/trojan/server.go#L163-L166

this is forgotten by the author for vless, and should be fixed.

PR #4926, opened 2025-07-24
All: Del 'else'

xqzr:
Ref #4544

### DIFF
--- a/app/dns/dns.go
+++ b/app/dns/dns.go
@@ -190,13 +190,12 @@ func (s *DNS) LookupIP(domain string, option dns.IPOption) ([]net.IP, uint32, er
 		return nil, 0, errors.New("empty domain name")
 	}
 
+	option.IPv4Enable = option.IPv4Enable && s.ipOption.IPv4Enable
+	option.IPv6Enable = option.IPv6Enable && s.ipOption.IPv6Enable
 	if s.checkSystem {
 		supportIPv4, supportIPv6 := checkSystemNetwork()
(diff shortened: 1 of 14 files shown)


patterniha:
in this situation:

```
var s typeS
if condition {
   s= a1
} else {
   s= a2
}

```
 
it is better to remove `else` by defining `var s = a2`, to avoid double assigning.

///

but we should not remove `else` in this situation:
```
func (s typeS) {
   if condition {
     s = s1
   } else {
     s = s2
   }
}
```

in this situation `s` already assigned(when calling the function), so you should not set `s = s2` before `if`, because if `condition` is true, you assign `s` three times !!! and this makes the code more suboptimal !!!

///

so some of your changes are wrong (dns.go, tun.go, wireguard.go, config.go)

patterniha:
Oh, your changes are basically wrong at all:

```
var c byte = bs[i]
if bs == nil {
	c = s[i]
}
```

you use `bs[i]` even if `bs` is nil !!! and this cause panic !!!

///

**this PR is basically wrong and should be closed.**

Fangliding:
写bench测了一下哪怕是操作[1024]byte这样相当大的变量执行十兆次这样的判断产生的差距不到零点零零几纳秒可能还是来自别的地方 性能这块几乎是没有差距的 这种情况可读性重要一点(虽然两种写法可读性基本没有很大差距) 所以道理上讲两个写法都行 没必要完全统一

PR #4942, opened 2025-07-26
UDP: fix some cone uses like STUN,... when address is domain

patterniha:
although https://github.com/XTLS/Xray-core/pull/4816 has no problem and cover most uses,

but it doesn't help for some special cone uses like STUN, when only initial-request-address can be a domain.

(https://github.com/XTLS/Xray-core/pull/4816 didn't break anything, just didn't fix this situation)

///

this PR also fix situations like STUN when only initial-request-address can be a domain.

///

solving other situations when other requests can also be domain requires more effort, needs the implementation of reverse-IP-lookup-map and some changes to "override.go", but it may be useless, because different domains may have the same-IP, so it is impossible to map IP to domain.

### DIFF
--- a/proxy/freedom/freedom.go
+++ b/proxy/freedom/freedom.go
@@ -285,14 +285,18 @@ func NewPacketReader(conn net.Conn, UDPOverride net.Destination, DialDest net.De
 		counter = statConn.ReadCounter
 	}
 	if c, ok := iConn.(*internet.PacketConnWrapper); ok {
-		isAddrChanged := false
-		if UDPOverride.Address != nil || UDPOverride.Port != 0 || DialDest.Address.Family().IsDomain() {
-			isAddrChanged = true
+		isOverridden := false
+		if UDPOverride.Address != nil || UDPOverride.Port != 0 {
+			isOverridden = true
 		}
+		changedAddress, _, _ := net.SplitHostPort(conn.RemoteAddr().String())
+
 		return &PacketReader{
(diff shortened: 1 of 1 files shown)


RPRX:
Reviewing... 我想了下如果配置中有覆写地址/端口，似乎返回包的来源也没那么重要，可以这么改

patterniha:
yes, if address is override in config, b.UDP remain `nil`, so socks-server use (initial)request-address as response-address.

anyway, response-address should be matched with one of request-addresses.

RPRX:
然后这个 PR 的逻辑是如果返回包的来源和发送包时的去向相同，则修改返回包的来源（可能一样，或改成 domain），LGTM

patterniha:
if the address before resolving is IP, changedAddr=unchangedAddr
otherwise when resolved-IP is recevied, it convert to domain.

also, for fakedns we have another change(domain <-> fakeIP) and "override.go" handle that, so we should not worry about that.

RPRX:
@patterniha You've been invited to join Project X organization

RPRX:
虽然暂时没有仓库的维护权限，但是 @patterniha 展现出来的代码能力是值得认可的，感谢你的每一次贡献

patterniha:
Thank you very much, it's an honor.

Actually, I saw Xray-core on my people's phones and wanted to help develop it so that I could help my people.

KobeArthurScofield:
Thank you, the thing works like I've expected before!

PR #4970, opened 2025-08-02
Root config: Add `version` config (`min` and `max`)

patterniha:
https://github.com/XTLS/Xray-core/pull/4968#issuecomment-3146592323


w0l4i:
> [#4968 (comment)](https://github.com/XTLS/Xray-core/pull/4968#issuecomment-3146592323)

this is for client side or server side ?
where to add this ?

patterniha:
Root config:

```
{
  "version": {
    "min": "25.8.3",
    "max": ""
  },
  "log": {},
  "inbounds": [],
  "outbounds": [],
  ...
}
```

it can be used in both server and client, but its use is to force clients to use specific versions.

w0l4i:
> ,
>     "max": ""

I've tried but it doesn't work well i use older core client like npv tunnel to check is this can prevent client from connecting because of old core version , but it didn't , i tried this code on costume config for client side + on server side too

w0l4i:
image
image
client and server side also server core version is the latest version that released today

patterniha:
This feature added in v25.8.3.
So it doesn't work in older versions.

For older versions you can use new alias
like: direct, tunnel, block(https://github.com/XTLS/Xray-core/pull/4968) to force clients use the latest version.

PR #5014, opened 2025-08-11
fix `MultiBuffer` `Compact` function

patterniha:
Unfortunately, the lack of error return in the two functions `read` and `write` (in `buffer.go`) caused the author to write the function `Compact` (in `multi_buffer.go`) incorrectly and no one has noticed until now.

///

`common.Must2(last.ReadFrom(curr))` does not necessarily append `curr` to `last`, also `common.Must2` has no effect here, because `ReadFrom` does not return any error on failure.

///

for example suppose:

in `Compact` function `for` loop, `last` == `mb[0]` and `curr` == `mb[1]`
and `last.start` == `2` and `last.end` == `8192`
and `curr.start` == `0` and `curr.end` == `2`
so `last.Len() + curr.Len() <= 8192`
but `common.Must2(last.ReadFrom(curr))`  doesn't actually do anything!!!

///

**`Compact` function is used in `tls` and `websocket`, and this caused the Xray-core to occasionally send corrupted data!!! Whoa!!!**

thanks to @Vdu6n1XMrq6694a8 for his/her great Reproduction.

### DIFF
--- a/common/buf/buffer.go
+++ b/common/buf/buffer.go
@@ -258,6 +258,9 @@ func (b *Buffer) IsFull() bool {
 func (b *Buffer) Write(data []byte) (int, error) {
 	nBytes := copy(b.v[b.end:], data)
 	b.end += int32(nBytes)
+	if nBytes != len(data) {
+		return nBytes, errors.New("incomplete write")
+	}
 	return nBytes, nil
 }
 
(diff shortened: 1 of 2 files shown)


patterniha:
also, because `buffer.Read()` did not return error by mistake, these two codes:

https://github.com/XTLS/Xray-core/blob/b1107b9810a32623afff491068494dd9a825f220/common/protocol/quic/sniff.go#L161

and

https://github.com/XTLS/Xray-core/blob/b1107b9810a32623afff491068494dd9a825f220/common/protocol/quic/sniff.go#L231

 did not work correctly and this cause **`SniffQUIC`** didn't work correctly.

Fangliding:
buf最好别动 只是 Compact 函数的问题 还是我来修吧

Issue #5012, opened 2025-08-10
A way to trigger the `ERR_SSL_PROTOCOL_ERROR` error


Vdu6n1XMrq6694a8:
### Integrity requirements

- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

## Summary

I have found a reliable way to reproduce `ERR_SSL_PROTOCOL_ERROR`, which can be reproduced on the latest Xray 25.8.3 with Chrome 139.
This reproduction method is not artificially constructed, but comes from a real example. I have simplified it for ease of description, so that maintainers can reproduce it more easily.

## Reproduction

In this example, there are three configuration files:

* A is the client, exposing port 10000, and uses `vmess+ws` with `"security": "zero"` to connect to B.
* B is the server, exposing port 10001, and uses `vmess+tcp` with `"security": "none"` to connect to C.
* C is another server, exposing port 10002, whose outbound is `freedom`.
The Chrome browser connects to A’s 10000 socks5 via Zero Omega (also known as Switch Omega).

The diagram is as follows:

<img width="781" height="176" alt="Image" src="https://github.com/user-attachments/assets/d8557bc9-60a7-4f15-aab7-c8032d9c882a" />

From my testing, almost any popular website can trigger this error, as it occurs randomly on certain requests. It is recommended to use websites with a large number of requests for testing, such as `x.com`/ `yahoo.com` and `baidu.com` (if you are in CN).

<img width="1032" height="73" alt="Image" src="https://github.com/user-attachments/assets/514756ab-e9ea-4cd2-8cba-7c90b1fee8c4" />

## Analysis

I have tried modifying the configuration to bypass this error and found that the value of `security` is the key factor. The issue can only be reproduced when using `zero` for A→B and `none` for B→C.

Whether TLS is used or not does not affect the outcome.

## Temporary solution

Please do not mix `zero` and `none`; it is recommended to use `zero` consistently.

### Reproduction Method

For easier reproduction, you can use the configuration files I have prepared.

[a.json](https://github.com/user-attachments/files/21706566/a.json) [b.json](https://github.com/user-attachments/files/21706567/b.json) [c.json](https://github.com/user-attachments/files/21706568/c.json)

### Client config

A:

[...]

PR #5015, opened 2025-08-11
Fix compact

Fangliding:
#5014

之前的bug写成test case了


Vdu6n1XMrq6694a8:
Thanks @Fangliding 

I've test this branch and it does fix the https://github.com/XTLS/Xray-core/issues/5012 👍

However, I can still logging like this:

```
[Info] [3437932928] app/proxyman/inbound: connection ends > proxy/vmess/inbound: connection ends > proxy/vmess/inbound: failed to transfer request > websocket: close 1000 (normal)
```

patterniha:
~10 hours non-stop debugging to find this bug~

patterniha:
@Fangliding 

My code is **full-compact**, but your code is **optimized-compact**.

So your code doesn't compact some data.

For example your code doesn't merge the first data into second data in:
[8190,8192], [0, 1000], [0, 1000], ...

**So, your code may send some small packets at beginning of the protocol, and this create a unique-characteristic that GFW can recognized.**

Fangliding:
> @Fangliding
> 
> My code is **full-compact**, but your code is **optimized-compact**.
> 
> So your code doesn't compact some data.
> 
> For example your code doesn't merge the first data into second data in: [8190,8192], [0, 1000], [0, 1000], ...
> 
> **So, your code may send some small packets at beginning of the protocol, and this create a unique-characteristic that GFW can recognized.**

协议最开始的部分都是new出来空buf 你告诉我怎么compact出若干小包？这个函数就是尽力而为将碎片buf合并而已 目的是效率 如果真要硬能合多少合多少那应该是readall才对

patterniha:
```
func Compact(mb MultiBuffer) MultiBuffer {
	if len(mb) == 0 {
		return mb
	}

	mb2 := make(MultiBuffer, 0, len(mb))
	last := mb[0]
	if last.start != 0 && last.Len() < Size {
		b := New()
		common.Must2(b.ReadFrom(last))
		last.Release()
		last = b
	}

	for i := 1; i < len(mb); i++ {
		curr := mb[i]
		if last.Len()+curr.Len() > Size {
			mb2 = append(mb2, last)
			last = curr
			if last.start != 0 && last.Len() < Size {
				b := New()
				common.Must2(b.ReadFrom(last))
				last.Release()
				last = b
			}
		} else {
			common.Must2(last.ReadFrom(curr))
			curr.Release()
		}
	}
	mb2 = append(mb2, last)
	return mb2
}
```

This code apply full-merge, and the performance is not much different from your code in practice, anyway, it is more important that we don't have unique-characteristics.

Fangliding:
bug修了就行 这没啥意义还难看 能是什么指纹 这都是用户产生的数据

patterniha:
> ~10 hours non-stop debugging to find this bug~

~after fixing about 30 bugs and at least 5 major bugs, and adding many features finally @Fangliding  liked my comment :)~

PR #5030, opened 2025-08-17
Some refines related to direct/freedom and `targetStrategy`; More intelligent "useIP"/"ForceIP", enhance "origin" functionality

patterniha:
1. we have duplicate codes related to domainStrategy, one in "transport/internet/config.go" and one in "proxy\freedom\config.go", i remove freedom-domainStrategy-codes and now freedom use internet-domainStrategy-codes.

2. we should not use outbound-handler Address() function to obtain outgoing-local-Addr, because we may set "origin" as sendThrough, and localAddr determined when we receive a connection, so we should use "ob.Gateway" instead, and i add "SetOutboundGateway" to set gateway.

3. suppose in inbound we use "quic-sniffer" or "fakeDNS", and in freedom-outbound we use "useIP"/"ForceIP" targetStrategy, suppose chrome send initial-quic-packet to an IPv4-fakeIP, so the UDP-payload size is 1250, after convert fakeIP to domain, because we use "useIP"/"forceIP" in outbound, the domain can resolve to an IPv6, but if our IPv6-interface-MTU is 1280(minimum possible), we have a packet with 1250+40+8=1298 size but 1298>1280, so we may have fragment-at-IP-level or packet fail to send at all.
regardless of MTU problem, even if our IPv6-MTU >= 1298, GFW can recognize these suspicious-packets, because GFW expect 1250-bytes for IPv4 and 1230-bytes for IPv6, but this can be the opposite.

so only for "UDP" and only for "useIP"/"forceIP" strategy, instead of selecting a random-IP between all IPv4 and IPv6 IPs, we should give priority to originalTarget-IP-type(try to select originalTarget-IP-type first, it means use "useIPv4v6"/"useIPv6v4" instead of "useIP" depending on originalTarget-IP-type)

also, mention that we cannot set two outbounds(one use "useIPv4v6" and one use "useIPv6v4"), because we cant route base on originalTarget-IP-type.


RPRX:
~~Outbound 那个 targetStrategy 用的是哪块代码~~

patterniha:
```
            strategy := h.config.DomainStrategy
			if destination.Network == net.Network_UDP && origTargetAddr != nil && outGateway == nil {
				strategy = strategy.GetDynamicStrategy(origTargetAddr.Family())
			}
			ips, err := internet.LookupForIP(dialDest.Address.Domain(), strategy, outGateway)
```

also you can see `GetDynamicStrategy` in "internet\config.go".

in short, it convert "useIP" to "useIPv4v6" or "useIPv6v4" base on originalDest-IP-type.

patterniha:
instead, we can add `originalTargetType` = "ipv4"/"ipv6"/"domain" to router, i think it is better, not?

patterniha:
for example if originalTarget is IPv4, we use "useIPv4v6" instead of "useIP",

**This does not change any logic, because in "useIP" we select a random IP between all-IP-types, and selected-IP can be any type, and "useIPv4v6" is the same, but it prioritize IPv4 to IPv6.**

**This is necessary so that we don't encounter MTU-problem or GFW-recognize-problem(for freedom)**

patterniha:
@RPRX 

Regardless of MTU-problem, **I found a more important reason**: 

almost all users bypass domestic websites, but if original-request-IP-type != final-sent-request-IP-type on UDP, they send 1250-bytes-on-IPv6/1230-bytes-on-IPv4, while it should be the other way around and **GFW easily detects that the user is using proxy-tools** !!!

so we should send to **same IP-type** as much as possible.

Fangliding:
这两个anyip检查又是什么意思

patterniha:
> 这两个anyip检查又是什么意思

currently, there is no difference between "listen": `""` and  `"0.0.0.0"` and `"[::]"`
in all three cases, we listen on **both** IPv4 and IPv6 and in all three cases we have `"[::]"` for UDP-localAddr.

but if we use `[::]` instead of `nil`, because `[::]` is IPv6, we only get IPv6 when we use "useIP"/"ForceIP" strategy:

https://github.com/XTLS/Xray-core/blob/5464862ee6ba7a994e2d2a2d227387c8ad084822/transport/internet/dialer.go#L88

but this is not correct and we should received both IPv4 and IPv6, because "origin" mean use inbound-local, and inbound-local listen on both IPv4 and IPv6.

///

localAddr is anyIP/anyIPv6 only when network is UDP and we listen on all-interface, so we should use `nil` instead of `[::]` to receive both IP-type.

Fangliding:
我已经把这个功能标记为不支持udp了 udp上这玩意就是损坏的不应该再管了 而且src也没有这个问题

patterniha:
when UDP is listen on one-interface it works, also we should have this feature for UDP, otherwise we should have multiple-outbounds beside multiple-inbounds, but now we only need to have multiple-inbounds and one outbound with `origin` is enough.

Nothing has changed except on "origin" when udp listen on one-interface "gateway" set to that, otherwise we don't set gateway(instead of not set at all)

so now by excepting anyIP everything works properly, and there is no need for multiple-outbounds, only multiple-inbounds is needed.

Fangliding:
如果你执意那么做应该在udpworker知道hub是anyip之后根据源IP把本地IP重置成正确的0000或者[::]

patterniha:
> If you insist on doing that, you should reset the local IP to the correct 0000 or [::] according to the source IP after udpworker knows that the hub is anyip.

I will do it, wait...

patterniha:
> 如果你执意那么做应该在udpworker知道hub是anyip之后根据源IP把本地IP重置成正确的0000或者[::]

done.

patterniha:
Tested and Ready.

RPRX:
~~合，都可以合~~

patterniha:
~I did a force-push just to add a neglected line of code~

RPRX:
> ~I did a force-push just to add a neglected line of code~

~~目前暂时没有要合的 PR 了，如果你还要改的话可以再 force-push~~

patterniha:
@RPRX 

regardless of freedom/direct, we may also encounter this problem in server-side.

suppose our final-target-website is behind cloudflare, and in server-side we resolve the domain to wrong-IP-version, so we may send 1250-bytes-to-an-IPv6, so regardless of MTU-problem, **cloudflare(target-service in general) easily detect that this request come from a proxy-tool (and it can block or send more captcha or ...)**.

///
**in short:**

**it is better to resolve UDP-domains in client-side, to resolve to same-IP-type as original-target, so that we don't encounter MTU-problem/traffic-detection-problem.
but resolve TCP-domains in server-side, to benefit from happy-eyeballs(golang/xray).**

///
**soloutions:**

1. extend `targetStrategy` options, for example use `udp+useIP`, to resolve only when network is UDP.(but it needs clients have built-in-dns)
2. extend `routeOnly` to 4-states, `all/tcp/udp/none`, so we can set `routOnly: udp`, so in case of quic-sniff we still use IP in outbound and sniffed-domain only use in routing , but in case of tls/http-sniff we use domain in outbound too.(but this does not work in case of fakeDNS)

///

**I now think it's unnecessary.**

PR #5046, opened 2025-08-22
Timer: prevent creating redundant check task

patterniha:
```
requestDone := func() error {
		defer timer.SetTimeout(plcy.Timeouts.DownlinkOnly)
                ...
}

responseDone := func() error {
		defer timer.SetTimeout(plcy.Timeouts.UplinkOnly)
                ...
}

```
after one timeout is set, another one will be set after timeout and return, but it is redundant.

//////////////////////////////////////////////////////////////////////////////////////////////

also, if UplinkOnly/DownlinkOnly timeout is already set , splice-timeout should not be set, and I will fix it soon.

https://github.com/XTLS/Xray-core/blob/33272a04994e81509fbd713c8dc1217c39f1e461/proxy/proxy.go#L599-L602

### DIFF
--- a/common/signal/timer.go
+++ b/common/signal/timer.go
@@ -56,12 +56,15 @@ func (t *ActivityTimer) SetTimeout(timeout time.Duration) {
 		return
 	}
 
+	t.Lock()
+	defer t.Unlock()
+	if t.onTimeout == nil {
+		return
+	}
 	checkTask := &task.Periodic{
 		Interval: timeout,
 		Execute:  t.check,
 	}
-
-	t.Lock()
 
 	if t.checkTask != nil {
 		t.checkTask.Close()
@@ -69,7 +72,6 @@ func (t *ActivityTimer) SetTimeout(timeout time.Duration) {
 	t.checkTask = checkTask
 	t.Update()
 	common.Must(checkTask.Start())
-	t.Unlock()
 }
 
 func CancelAfterInactivity(ctx context.Context, cancel context.CancelFunc, timeout time.Duration) *ActivityTimer {


patterniha:
> also, if UplinkOnly/DownlinkOnly timeout is already set , splice-timeout should not be set, and I will fix it soon.

done.

Fangliding:
overridden参数太奇怪了 一个ActivityTimer应该就不存在设置之后无法触发onTimeout的情况除非被故意设置得非常长 这里防止leak的话按目的应该改成 SetTimeoutShorter 或者类似的东西 检查checkTask(如果存在)的interval大于设定的值就重置 否则不重置

patterniha:
@Fangliding 

splice-timeout is 8-hours, **it is greater than both** `connIdle`(5 min), and `UplinkOnly/DownlinkOnly`(2-5 sec).

First, timer set to `connIdle` and if one-side is disconnected, it becomes `UplinkOnly/DownlinkOnly`.

but only when timeout is `connIdle`, we can change it to 8 hours, and if one-side is already disconnected we should not change the timer-timeout.

///

so, if timeout is changed by `UplinkOnly/DownlinkOnly`(=overridden), we should not set splice-timeout to 8 hours, otherwise we can set.
so we need `SetTimeoutIfNotOverridden`.

///

also, in splice, we change the timeout to 8 hours, because we can't update timer when splice, so we set it to long enough value, i change it to 24 hours(default TCP NAT timeout)

Fangliding:
哦我误解了注释里的prevent leak

patterniha:
@Fangliding  changes timer codes, so i rebase this pr, after that.
or @Fangliding  it is better to add "overriden" to your code and fix this problem too, then i close this pr.

Fangliding:
我不确定这是否正确 重新读了一遍我觉得这甚至应该是错的
这里这么操作的原因是splice不执行一般的 io 操作导致无法在每次copy的时候都执行对timer的更新 设置成8小时的作用是**完全禁用超时操作** 执行uplink/downlinkonly的操作后不是两三秒内就关掉 是把需要update的时间缩短到这么短 如果在半关闭后仍有一端在持续传输那么这个连接是完全可以活下去的 你这么做会导致uplink关闭两三秒后downlink必定关闭 这是错误的

patterniha:
>  If one end continues transmitting after a half-close, the connection can survive. Doing this will cause the downlink to close two or three seconds after the uplink closes, which is incorrect.

? usually `DownlinkOnly-timeout` is set after `splice-8-hours-timeout`, and this makes splice-reading closed even if we have continues-transmitting, this is another problem.

patterniha:
anyway, for my use, i set uplinkOnly/downlinkOnly to `0`(in serverless), So that i don't encounter this problem.
I'll check it again later.

PR #5043, opened 2025-08-21
Freedom/Direct: use simple cache instead of map cache for UDP

patterniha:
https://github.com/XTLS/Xray-core/pull/4804#discussion_r2290096423

Issue #3131, opened 2024-03-12
Shadowsocks2022 RAM Usage


rhawsh:
_Hello._

I was in the middle of a test, and I noticed that the SS22 server RAM usage was much higher than other servers.

Server 1 : xray-core v1.8.9 : Shadowsocks2022 (tcp+udp)
Server 2 : xray-core v1.8.9 : Vless-tcp-tls+nginx
Server 3 : xray-core v1.8.9 : Vless-tcp-reality-vision

The uptime is correct and the server load is exactly equal. _(network load and user counts are equal between 3 servers)_

Is SS22 RAM usage always this much compared to other Vless combinations?
![xray-comparison](https://github.com/XTLS/Xray-core/assets/163159839/387f1c7f-f805-44d2-9e11-1995de732d9c)

masbur:
Just like a trojan, the RAM will be full and restart automatically due to OOM (Out of Memory)

I use vless to fallback to many modes like trojan ws, grpc and tcp

Nginx -> Vless TLS / Non TLS -> Trojan [ WS / gRPC / TCP ]

I don't know how to fix it. I've tested on older versions but still getting OOM

RPRX:
目前 Xray 中的 SS2022 是基于 sing 那边的代码，在 Xray 这里好像有点 buggy（比如服务端 UDP），正好 v2fly 前段时间也实现了 SS2022，你测下它的有没有 OOM 问题，没问题的话我们 port 过来即可

RPRX:
刚刚在想 XUDP 的 Global ID 没读取 SS2022 UDP 的那个 ID，回头看了下 VLESS 转 VLESS 好像都没传递 Global ID，还是要写一下

PR #5050, opened 2025-08-22
Trojan-UoT & UDP-nameserver: Fix forgotten release buffer; UDP dispatcher: Simplified and optimized

patterniha:
https://github.com/XTLS/Xray-core/pull/5043#issuecomment-3215492930

simplified and optimize UDP dispatcher code.
also, after closing `connEntry`, it's timer was still active and was not canceled, and i fix that.


patterniha:
Also, probably, the memory leak problem solved in https://github.com/XTLS/Xray-core/pull/5050/commits/6af9a6207b47a2c0f8e446d17852bfc2ab2e2822.

we should release buffers, after we're done with them, but we didn't do that in trojan-uot and udp-dns.

Fangliding:
又在瞎猜了
1 bug是新版本出现的
2 忘记release不会导致leak 它只是把buffer返还给复用池 就算不还的话会被gc回收的

patterniha:
also, this matches the pprof chart:
`freedom-ReadMultiBuffer -> udp-dispatcher-handleInput-callback -> trojan-handleUDPPayload-NewDispatcher-WriteMultiBuffer`

my guess is that the memory leak problem is solved,  ~anyway, this is clearly forgotten-release-buffer and should be fixed.~

@kastov  I am waiting for your confirm

patterniha:
~also we should `Interrupt(c.link.Writer)` instead of `Close(c.link.Writer)` in udp-dispatcher, because we may have some data in link-cache~

@kastov  test https://github.com/XTLS/Xray-core/tree/katsov-test

Fangliding:
Screenshot_2025-08-23-17-58-28-121_com android chrome-edit
这些稀碎东西能不能在你自己的frok试 别推到主仓库来污染眼球 just-test都来了

patterniha:
Oh, i didn't know that even non-main branch are displayed.

kastov:
> ~also we should `Interrupt(c.link.Writer)` instead of `Close(c.link.Writer)` in udp-dispatcher, because we may have some data in link-cache~
> 
> @kastov test https://github.com/XTLS/Xray-core/tree/katsov-test

working on it

kastov:
Here is also fresh dumps. Current ram usage 5.4GB. Yesterday was 3.2GB.

Снимок экрана 2025-08-23 в 18 40 21

[heap+gor+alloc-23082025.zip](https://github.com/user-attachments/files/21950177/heap%2Bgor%2Balloc-23082025.zip)

I will relaunch now with core from branch which you provided

patterniha:
@kastov  katsov-test branch updated, please use the new one.

kastov:
> @kastov katsov-test branch updated, please use the new one.

oh, i already started previous one. i believe it is better to keep that one for at least 24 hours to see results

patterniha:
if possible use new one for 24 h.

kastov:
> @kastov katsov-test branch updated, please use the new one.

btw any ideas how we can stresstest core to see if there any leaks? current way of testing is just so slow... :D

Fangliding:
> Here is also fresh dumps. Current ram usage 5.4GB. Yesterday was 3.2GB.
> 
> Снимок экрана 2025-08-23 в 18 40 21
> [heap+gor+alloc-23082025.zip](https://github.com/user-attachments/files/21950177/heap%2Bgor%2Balloc-23082025.zip)
> 
> I will relaunch now with core from branch which you provided

Try this https://github.com/XTLS/Xray-core/actions/runs/17186238756
Combining two pprofs, I reverted the only most suspicious commit

kastov:
Hi there! Ram usage: 2.8GB

From branch: https://github.com/XTLS/Xray-core/tree/3236d737f665f99d12ee8c74eb4b86e5a4be1c5d (without last 4 commits). 


Src


[heap+alloc+gor+24.zip](https://github.com/user-attachments/files/21957229/heap%2Balloc%2Bgor%2B24.zip)

kastov:
> Try this https://github.com/XTLS/Xray-core/actions/runs/17186238756 Combining two pprofs, I reverted the only most suspicious commit

Working on it

kastov:
@patterniha @Fangliding currently i'm testing https://github.com/XTLS/Xray-core/actions/runs/17186238756. 

I will share results tomorrow i believe, after that i will launch latest core from branch: https://github.com/XTLS/Xray-core/tree/katsov-test

Fangliding:
> @patterniha @Fangliding currently i'm testing https://github.com/XTLS/Xray-core/actions/runs/17186238756.
> 
> I will share results tomorrow i believe, after that i will launch latest core from branch: https://github.com/XTLS/Xray-core/tree/katsov-test

I want to confirm: Is the old version has no problem?

kastov:
> > @patterniha @Fangliding currently i'm testing https://github.com/XTLS/Xray-core/actions/runs/17186238756.
> > I will share results tomorrow i believe, after that i will launch latest core from branch: https://github.com/XTLS/Xray-core/tree/katsov-test
> 
> I want to confirm: Is the old version has no problem?

Yeah, for sure. Something definitely changed, i see the same picture almost on all servers.

I'm using [@remnawave/node](https://github.com/remnawave/node), we have skipped 25.7.x version of core, so before 25.8.x there was 25.6.8. 
Upgrade to 25.8.3 was in v2.1.0.

Shortly after that (17-19 August) , i updated some of my servers to latest remnawave/node, so core was bumped.

And it clearly seen that something is happening, cause almost all servers have the same graphs

CleanShot 2025-08-24 at 22 50 08@2x

CleanShot 2025-08-24 at 22 50 49@2x

CleanShot 2025-08-24 at 22 51 27@2x

CleanShot 2025-08-24 at 22 52 01@2x

kastov:
and of course i checked twice, who is occupying RAM

CleanShot 2025-08-24 at 22 54 51@2x

kastov:
@patterniha which one i need to test from your commits? branch /katsov-test or #5050?

patterniha:
> @patterniha which one i need to test from your commits? branch /katsov-test or #5050?

there is no need, i find the bug, I'm writing a very interesting article about it, many people should apologize, today was a shameful day.

patterniha:
@RPRX @kastov @Fangliding 

this High RAM usage due to large number of uot-trojan-connections from the client side and it has nothing to do with server-side.

when a client use trojan protocol without mux, for each udp-request a new-trojan-connection is created to the server.
so if clients don't use built-in-dns, for each dns-request we have separate connection to server, just to send that one request, **so we create a new connection only for sending and receiving one data!** (**because each dns-request has different source port**)

i tested client-trojan-without-using-built-in-dns and in less than 10 seconds after visiting 2-3 websites hundreds of connections were made to the server(because each websites has multiple subdomain and for each, we create separate connection)
and this connections remain open in server-side at least for 1-minutes:

Screenshot 2025-08-25 102108

///

**but if clients use built-in-dns, all requests redirect to built-in-dns, and only one udp-dispatcher is created, so we only create one connection to the server to send all dns-requests.**

**so when clients use non-mux protocol like trojan, they should use built-in-dns, otherwise they creates many connections to the server, only for sending a dns-requst for each.**

///

**The only question is why has this problem become more apparent recently?**

The answer comes back to v2rayNG.

recently, v2rayNG add new tun2socks-module "hev-socks5-tunnel": https://github.com/2dust/v2rayNG/pull/4800
but "hev-socks5-tunnel" does not separate dns-request from other requests, and all requests go to socks-inbound(Previously, dns-requests go to dokomo-inbound-with-name-dns-in),  so the rule:
```
{
  "inboundTag": "dns-in",
  "outboundTag": "dns-out"
}
```
has no effect in v2rayNG for clients that use new-tun, and all dns-requests directly go to the server, so we have connections as many as dns-requests, and the above problem occurs.

although, this problem is solved in recent versions of v2rayNG: https://github.com/2dust/v2rayNG/pull/4806
but many users still use old-version(1.10.13), or they connected before, and the wrong-configuration is cached.

///

also, in many places in core we create a new buffer, but we didn't release it after we're done with it.
although, it will be recycled by the GC, but we should release all unused buffers for better performance.

///

**update:**

It seems that the increase in uot-connections is due to the possibility of using quic-protocol for clients-that-use-fakedns(see https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3222649191)

kastov:
> > Try this https://github.com/XTLS/Xray-core/actions/runs/17186238756 Combining two pprofs, I reverted the only most suspicious commit
> 
> Working on it

@Fangliding i got fresh results from this version of core. Unfortunately, the leak steal exist as i can see.

Xray RAM usage – 2.5GB

XXX

[alloc+gor+heap+25.zip](https://github.com/user-attachments/files/21967462/alloc%2Bgor%2Bheap%2B25.zip)

kastov:
> this High RAM usage due to large number of uot-trojan-connections from the client side and it has nothing to do with server-side.

So in your opinion if i start 25.6.8 – the leak will steel occurs?

patterniha:
> > this High RAM usage due to large number of uot-trojan-connections from the client side and it has nothing to do with server-side.
> 
> So in your opinion if i start 25.6.8 – the leak will steel occurs?

yes, otherwise, further investigation is required, please test.

kastov:
> > > this High RAM usage due to large number of uot-trojan-connections from the client side and it has nothing to do with server-side.
> > 
> > 
> > So in your opinion if i start 25.6.8 – the leak will steel occurs?
> 
> yes, otherwise, further investigation is required, please test.

Sure, deploying

patterniha:
@kastov 
but the leak should be a little slower, because there is no sync.map for caching-udp-address, but the difference should be very small unless there's a problem with it's recycling, but the code is correct, and it should be recycled correctly.

patterniha:
@RPRX 

this PR includes:

1. in some places, we didn't release unused buffer, after we're done with them, i I safely release them.

2. in udp-dispatcher we didn't cancel timer after we closed `connEntry`, i fix that.

3. we should set `b.UDP` for udp_nameserver-packets otherwise we cannot use redirect-freedom-settings/mux for that.

4. in "freedom.go", the second use of `ResolveUDPAddr` always get IP, so we should use `RawNetAddr` instead for better performance.

5. in "proxy.go", `io.EOF` error should convert to `nil`, otherwise task.run immediately return and `UplinkOnly` has no effect.

///

**READY**, ~also, the changes are simple, you can check it yourself.~

kastov:
> **READY**, ~also, the changes are simple, you can check it yourself.~

Do i need to test it?

patterniha:
No, during the search for your problem, i found these small issues, which are irrelevant to your problem(except 1 & 2, which I had fixed at the beginning, and of course it had little effect on your problem)

kastov:
I know it is too early to make serious conclusions, but...

In my opinion, it is definitely fault of newer version Xray core, not 3rd-party software like v2rayNG (which was not used by my users btw).

As you can see, i make simple annotations for graph, which i made for last week. So, 25.6.8 performs much more better for the same period of time, comparing for each 25.8.3 versions (or versions with fixes from @patterniha and @Fangliding). 

So it shows, what 25.6.8 within the same time period does not have any tendencies to "occupy" more memory.

Diff


I will share end-report tomorrow (full 24h cycle) for 25.6.8 version with pprof dumps.

patterniha:
first, send your full-server-config.
I think about that and i found another reason(i hope this is the main reason):

without https://github.com/XTLS/Xray-core/pull/4816 and https://github.com/XTLS/Xray-core/pull/4942 clients-that-use-fakedns cannot use quic-protocol at all, because request-header-address didn't match with response-header-address and browsers reject the response, **so after this fix, many users use quic-protocol**, and this cause more ram usage.

first, test https://github.com/XTLS/Xray-core/tree/new-kastov-test.
if the problem solved, you should block quic-protocol(udp-443) on your server(quic is not suitable for uot at all because we will have double-retransmission-control)

kastov:
25.6.8, 24h

Current ram usage: 744MB

CleanShot 2025-08-26 at 15 29 04@2x


[heal+gor+alloc_2568.zip](https://github.com/user-attachments/files/21987743/heal%2Bgor%2Balloc_2568.zip)

patterniha:
quic has it's own retransmission-control, when you use tcp-base-transport, for transmitting quic, double retransmission-control will be activated, and this puts more pressure on your server, after https://github.com/XTLS/Xray-core/pull/4816, clients-that-use-fakedns will now be able to use quic-protocol: https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3222649191.
So if you want to support more users on your server, you should block quic(udp-443).

Maybe this is not the main reason, so please test https://github.com/XTLS/Xray-core/tree/new-kastov-test before making any conclusions.

gfw-killer:
Are you still debugging without a full server and client config?

patterniha:
I have it now, I just wanted to see if quic was blocked or not, and I saw that it wasn't.

patterniha:
@RPRX 

[This](https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3224087025) wasn't the main reason either.

The problem wasn't the large number of uot-connections, the problem is that each uot-connection never closes on server-side and remain open forever.

udp-dispatcher does not close main-inbound-connection when closed and only close link, so in case of trojan-uot-connection, connection is closed only after received `fin-ack` from client.

But the strange thing is that v2rayNG doesn't send `fin-ack` after closing the VPN.

although after 2-minutes android-system send `fin-ack`, but the android-network may be disconnected/changed during this time and `fin-ack` does not reach to the server and connection remain open forever.

///

but before version v25.7.25 we have `server-side-keepAlive` and this cause that every uot-connection closes after 15-seconds-idle-timeout-when-no-keepalive-response, but after https://github.com/XTLS/Xray-core/commit/eb433d9462b6fb47759cb2499f724ab1e3954ea9, serverside-keepalive is disabled by default(`lc.KeepAlive = -1`) and this cause trojan-uot-connections remain open forever(if not received `fin-ack`) and we have ram/goroutine leak.

///

this only happen on udp-dispactcher case, because after closing link, it didn't close main-inbound connection.

///

although if we revert https://github.com/XTLS/Xray-core/commit/eb433d9462b6fb47759cb2499f724ab1e3954ea9 the connections closes after 15-seconds-idle-no-keepalive-response, and the problem solved, but we should solve this problem for udp-dispatcher too.

///

@RPRX  I know that you want to release today, but please wait just another day, this is important problem.
@kastov is testing without https://github.com/XTLS/Xray-core/commit/eb433d9462b6fb47759cb2499f724ab1e3954ea9, wait for his confirmation.

RPRX:
具体是啥 problem

> the problem is that each uot-connection never closes on server-side and remain open forever.

XUDP UoT Migration，TCP 断连的话服务端会等几分钟再扔掉出口端口，不过这个是很早就有的功能，与你们说的 problem 无关

patterniha:
i talk about trojan, it does not use XUDP.
also in vless when we use UDP-port-443/53 we still does not use XUDP, but we use simple-dispatcher, 
but in trojan we use UDP-dispatcher.

in inbound-vless after dispatcher-link is closed, `Process` will end too, and because inbound-connection is closed after `Process`, we don't have any problem.

but when we use udp-dispatcher, `Process` will not return, and only link is closed.

so only in case of **trojan-uot-connection**, connection is closed only after:

1. received `fin-ack` from client
2. 15-seconds idle-timeout exceeds, and no response for keepalive sent by server

but in https://github.com/XTLS/Xray-core/commit/eb433d9462b6fb47759cb2499f724ab1e3954ea9 we disabled `server-keepAlive` and if `fin-ack` is not received for any reason, connection remain open forever.

patterniha:
**in short**, if `fin-ack` is not received for any reason, We get stuck in:
https://github.com/XTLS/Xray-core/blob/c2141f09e7fea8cb3d3b760a91cc6091c51d6163/proxy/trojan/server.go#L274

but before https://github.com/XTLS/Xray-core/commit/eb433d9462b6fb47759cb2499f724ab1e3954ea9, we had `server-keepalive`, and it returns error if no keepalive-response received, but now it stuck forever.

RPRX:
> 2 忘记release不会导致leak 它只是把buffer返还给复用池 就算不还的话会被gc回收的

会导致 leak，因为 GC 不会自动调用 `pool.Put(p)`

patterniha:
> Will cause leaks, because GC will not be called automaticallypool.Put(p)

i fix this in this PR, but @kastov says problem still exists: https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3220196315

RPRX:
> i talk about trojan

Trojan ~~？那没事了，有爱玩的就自求多福吧，我可不管~~

patterniha:
first, this PR should be merged: https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3220196315

second, If you give me one day, I'll fix the trojan-uot-problem for the new version.

third, @Fangliding  why you disable `default-inbound-tcp-keepAlive`???

RPRX:
纠正一下

```
// Get selects an arbitrary item from the [Pool], removes it from the
// Pool, and returns it to the caller.
```

Get() 后不 Put() 会被 GC，只是没起到 pool 的作用

patterniha:
~there is no need for one day, i find a easy way, i fix it soon.~

Fangliding:
> first, this PR should be merged: [#5050 (comment)](https://github.com/XTLS/Xray-core/pull/5050#issuecomment-3220196315)
> 
> second, If you give me one day, I'll fix the trojan-uot-problem for the new version.
> 
> third, @Fangliding why you disable `default-inbound-tcp-keepAlive`???

因为参考的nginx没开 服务端就算不开应该还是有 connIdle 来管 不过Trojan uot好像没有正确实现 connIdle。。

patterniha:
> However, the Trojan uot does not seem to implement connIdle correctly.

yes, this is the reason, and i fix it in: https://github.com/XTLS/Xray-core/pull/5064

artoor32:
Also, Shadowsocks 2022 has this memory leak issue #3131 . Does this PR fix the problem in Shadowsocks as well?

patterniha:
> Also, Shadowsocks 2022 has this memory leak issue #3131 . Does this PR fix the problem in Shadowsocks as well?

first, https://github.com/XTLS/Xray-core/pull/5064, fix the trojan memory leak problem, this is just some refines and i found them during searching of main problem.

second, no, i just modify trojan code.

patterniha:
@RPRX 

this PR need rebase, wait...

RPRX:
~~第 1500 个 commit 留给你~~

patterniha:
READY.

PR #5073, opened 2025-08-30
Trojan uot: customizable policy bufferSize

patterniha:
After adding `connIdle`, we can also add `bufferSize`.


RPRX:
https://github.com/XTLS/Xray-core/pull/5067#issuecomment-3239412216

PR #5081, opened 2025-09-02
DNS outbound: Fix some issues

patterniha:
1. because we don't have `UplinkOnly/DownlinkOnly`, after outbound-connection is closed(`response` returned), `request` function(and as a result inbound-connection) remain open for a long time(especially when the network is TCP)

///

2. we have a weird code in:
https://github.com/XTLS/Xray-core/blob/4064f8dd8080b0e278df79e0a42c0b153b1d445e/proxy/dns/dns.go#L393

    it should be: `return 0, err`(dns-query is not sent but we are waiting for a response!)

### DIFF
--- a/proxy/dns/dns.go
+++ b/proxy/dns/dns.go
@@ -209,7 +209,7 @@ func (h *Handler) Process(ctx context.Context, link *transport.Link, d internet.
 				}
 				if isIPQuery {
 					b.Release()
-					go h.handleIPQuery(id, qType, domain, writer)
+					go h.handleIPQuery(id, qType, domain, writer, timer)
 					continue
 				}
 				if h.nonIPQuery == "drop" {
@@ -259,7 +259,7 @@ func (h *Handler) Process(ctx context.Context, link *transport.Link, d internet.
 	return nil
 }
 
-func (h *Handler) handleIPQuery(id uint16, qType dnsmessage.Type, domain string, writer dns_proto.MessageWriter) error {
+func (h *Handler) handleIPQuery(id uint16, qType dnsmessage.Type, domain string, writer dns_proto.MessageWriter, timer *signal.ActivityTimer) {
 	var ips []net.IP
 	var err error
 
@@ -284,7 +284,7 @@ func (h *Handler) handleIPQuery(id uint16, qType dnsmessage.Type, domain string,
 	rcode := dns.RCodeFromError(err)
 	if rcode == 0 && len(ips) == 0 && !go_errors.Is(err, dns.ErrEmptyResponse) {
 		errors.LogInfoInner(context.Background(), err, "ip query")
-		return err
+		return
 	}
 
 	switch qType {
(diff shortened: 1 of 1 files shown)


4ktivated (approved):

RPRX:
`b, err := reader.ReadMessage()` 没有传入 ctx，不还是会在这一行卡着吗？

patterniha:
> `b, err := reader.ReadMessage()`If ctx is not passed in, won't it still be stuck on this line?

after connIdle-timeout exceed -> ctx is canceled ->task.run ends ->Process ends -> 
1. if link.reader is pipe, we have interrupt(reader) after Process in proxyman/outbound/handler.go
2. else if it is inbound-reader(dispatchLink) -> dispatchLink ends -> inbound-Process ends -> inbound-conn closed -> reader closed

so in both case reader is closed and there is no problem.

patterniha:
ready.

patterniha:
also, there is no need for `go h.handleIPQuery(id, qType, domain, writer)` or `go h.rejectNonIPQuery(id, qType, domain, writer)`, there is no need to run them in new goroutine,  we can simply do `h.handleIPQuery(id, qType, domain, writer)` or `h.rejectNonIPQuery(id, qType, domain, writer)`, i will ~fix~  change it .

Fangliding:
好好的异步查询响应本来就可以乱序回又突发奇想了是吧。。。 你改点什么就是fix 然后说我fix了好几十个bug

patterniha:
i said there is no need to run in new goroutine, `write` instantly returned.
so it is better not to make a new goroutine, this is not a bug, just optimization, i didn't say this a bug.

~it is better to call it "change" instead of "fix"~

patterniha:
> also, there is no need for `go h.handleIPQuery(id, qType, domain, writer)` or `go h.rejectNonIPQuery(id, qType, domain, writer)`, there is no need to run them in new goroutine, we can simply do `h.handleIPQuery(id, qType, domain, writer)` or `h.rejectNonIPQuery(id, qType, domain, writer)`, i will ~fix~ change it .

done.

Fangliding:
> i said there is no need to run in new goroutine, `write` instantly returned. so it is better not to make a new goroutine, this is not a bug, just optimization, i didn't say this a bug.
> 
> ~it is better to call it "change" instead of "fix"~

你optimize了个啥 一个goroutine而已 dns本来就可以乱序处理是对的 就是些无意义改动

patterniha:
the main bug in this PR was `return len(b), nil`, and after outbound-connection is closed(or dial failed), inbound-connection remained open for 300s .

also, there is no need for new goroutines, we should simplify the code.

///
@Fangliding 

also, outbound-dns has `userLevel` option:

https://github.com/XTLS/Xray-core/blob/19f890729656bc923ae3dee8426168c93b8ee9c2/infra/conf/dns_proxy.go#L14

but this is forgotten in CN-doc, please add it, thx.

patterniha:
@RPRX 

@Fangliding is right, because `handleIPQuery` use built-in-DNS and it may **take a while** to answer, it is better to run it in new goroutine.
but for `rejectNonIPQuery` there is no need for new goroutine.

so only `handleIPQuery` should run in goroutine, wait...

patterniha:
> so only handleIPQuery should run in goroutine, wait...

done and ready.

~At first I didn't understand what @Fangliding  meant~

PR #5099, opened 2025-09-06
Commands: Fix "with SNI" printing fixed port 443 for `tls ping`

AndyChiang888:
Currently, when Pinging with SNI, port 443 is always used, so should use TargetPort instead of 443.


AndyChiang888:
@RPRX Hi, can be merged?

RPRX:
感谢 PR

Issue #5088, opened 2025-09-05
25.9.5版本的反向代理无法正常使用


kmzs123:
### 完整性要求

- [x] 我读完了 issue 模板中的所有注释，确保填写符合要求。
- [x] 我保证阅读了文档，了解所有我编写的配置文件项的含义，而不是大量堆砌看似有用的选项或默认值。
- [x] 我提供了完整的配置文件和日志，而不是出于自己的判断只给出截取的部分。
- [x] 我搜索了 issues, 没有发现已提出的类似问题。
- [x] 问题在 Release 最新的版本上可以成功复现

### 描述

25.9.5版本更新后之前正常使用的反向代理无法正常使用了，我用相同的配置文件测试了25.8.31版本是正常的

### 重现方式

在25.9.5版本使用反向代理Reverse即可重现

### 客户端配置

<details><pre><code>
{
	"log": {
		"access": "log",
		"error": "log",
		"loglevel": "debug",
		"dnsLog": true
	},
	"outbounds": [{
            "protocol": "freedom",
            "tag": "out"
        }, {
		"tag": "interconn",
		"protocol": "vless",
		"settings": {
			"vnext": [{
				"address": "127.0.0.1",
				"port": 1080,
				"users": [{
					"id": "05f760f7-7543-4089-bcf2-e83e8edaa723",
					"encryption": "none"
				}]
			}]
		},
		"streamSettings": {
			"network": "xhttp",
			"xhttpSettings": {
				"path": "/hello/"
			},
			"security": "reality",
			"realitySettings": {
				"show": false,
				"fingerprint": "chrome",
				"serverName": "newtrackon.com",
				"password": "yLGiEn5Zy8NyEdBDzwJ1PIJjy3bmoQufi_6NxcZ0tg4",
				"shortId": "eb11bd7f06860756",
				"spiderX": "/world"
			}
		}
	}],
	"reverse": {
		"bridges": [{
			"tag": "bridge",
			"domain": "reverse-proxy.xray.internal"
		}]
	},
	"routing": {
		"rules": [{
			"type": "field",
			"inboundTag": ["bridge"],
			"domain": ["full:reverse-proxy.xray.internal"],
			"outboundTag": "interconn"
		}, {
			"type": "field",
			"inboundTag": ["bridge"],
			"outboundTag": "out"
		}]
	}
}
</code></pre></details>

### 服务端配置

<details><pre><code>
{
    "log": {
        "access": "log",
        "error": "log",
[...]

PR #5100, opened 2025-09-07
app/reverse/portal.go: Fix goroutine leak & Add EndpointOverride

patterniha:
https://github.com/XTLS/Xray-core/issues/5088#issuecomment-3263558403

### DIFF
--- a/app/reverse/portal.go
+++ b/app/reverse/portal.go
@@ -92,6 +92,12 @@ func (p *Portal) HandleConnection(ctx context.Context, link *transport.Link) err
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


patterniha:
also, portal-dispatch didn't implement "EndpointOverride" and i added.(otherwise, we have udp-mismatch-address problem when quic-sniffer is enabled in any inbound in portal-side)

patterniha:
also, because mux-client doesn't have `connIdle` mechanism, `muxClient` may never be closed and we still have goroutine-leak.



wait...

RPRX:
。。。

patterniha:
it is ok, i will open another PR.

RPRX:
Mux 的修复另开个 PR 吧

PR #5110, opened 2025-09-08
MUX: Prevent goroutine leak

patterniha:
fix problems:

1. after using `DispatchLink`, we may have goroutine-leak when connection between mux-client and mux-server cut.

2. in some cases in mux `dispatchLink` wait on wrong things and it cause code-stuck and goroutine leak.

3. for reverse, we may have goroutine-leak when a sub-client-connection cut

///

4. most NAT TCP translations time out is 24 hours, so change splice-timeout to 24 hours.

### DIFF
--- a/app/reverse/bridge.go
+++ b/app/reverse/bridge.go
@@ -55,7 +55,7 @@ func (b *Bridge) cleanup() {
 			activeWorkers = append(activeWorkers, w)
 		}
 		if w.Closed() {
-			w.timer.SetTimeout(0)
+			w.Timer.SetTimeout(0)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


Fangliding:
你真的可以在旧版本触发出来这些bug吗

patterniha:
> 你真的可以在旧版本触发出来这些bug吗

this bugs appear after using `DispatcherLink`, for example if connection between mux-client and mux-server cut(with no fin-ack), the code stuck in:

https://github.com/XTLS/Xray-core/blob/ab1fa13ebebdb153fce3502cbcaa35d88dcd5123/common/mux/server.go#L294

so we should implement "monitor" for mux-server too(when we have no session and server-mux is idle for a 60 seconds, ServerWorker should be closed, it is safe because client-mux closes after 16-seconds idle-time)

///

or on the client-side, "DispatcherLink" waited on `fetchInput` and as a result waited on:

https://github.com/XTLS/Xray-core/blob/ab1fa13ebebdb153fce3502cbcaa35d88dcd5123/common/mux/client.go#L278

but if sub-client-inbound-connection cut(with no fin-ack), the code stuck here, because now `s.input` is not pipe-reader and `common.Interrupt(s.input)` has no effect, so we should implement `done.Instance` for `Session` and `DispatcherLink` should wait on it.

RPRX:
没有破坏 XUDP UoT Migration 吧

patterniha:
It has nothing to do with XUDP, just added "monitor" for mux-server-side.
also dispatcherLink should not wait on inbound-reader, because inbound-reader may ends only after dispatcherLink ends, and this cause stuck.(when no fin-ack received)

patterniha:
checked again and ready.

patterniha:
rebased.

PR #5200, opened 2025-09-30
External config: Add unix socket HTTP loader support

kastov:
Adds support for loading configuration from HTTP endpoints served over Unix domain sockets using the http+unix:// protocol scheme.


yuhan6665:
Looks good to me!

PR #5210, opened 2025-10-06
Router: Use built-in-dns only once for all rules (in "IPOnDemand"/"IPIfNonMatch" mode)

patterniha:
**we should have the same resolved-IPs for a domain for all IP-rules**, otherwise wrong-outbound may be chosen (discussed in https://github.com/XTLS/Xray-core/pull/5138).

also, currently, if for example we have 10 IP-rules and our dns-server is unavailable, we wait a long time to choose the outbound (at least 4 seconds for each = 40 seconds!).

so if we encounter an error for an IP-rule (failed to resolve the domain to IP), we should use that for all other IP-rules and **don't use built-in-dns again for every IP-rule**.

///

also note that, this is only for one-dispatch and everything is reset for the next-dispatch.

### DIFF
--- a/features/routing/dns/context.go
+++ b/features/routing/dns/context.go
@@ -12,20 +12,15 @@ import (
 // ResolvableContext is an implementation of routing.Context, with domain resolving capability.
 type ResolvableContext struct {
 	routing.Context
-	dnsClient   dns.Client
-	resolvedIPs []net.IP
-	hasError    bool
+	dnsClient dns.Client
+	cacheIPs  []net.IP
+	hasError  bool
 }
 
 // GetTargetIPs overrides original routing.Context's implementation.
 func (ctx *ResolvableContext) GetTargetIPs() []net.IP {
-
-	if ips := ctx.Context.GetTargetIPs(); len(ips) != 0 {
-		return ips
-	}
-
-	if len(ctx.resolvedIPs) > 0 {
-		return ctx.resolvedIPs
+	if len(ctx.cacheIPs) > 0 {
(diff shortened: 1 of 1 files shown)


Meo597:
这次是真的拉了一坨大的，弄出个 Breaking Change

本来非 AsIs 目的是不信任原始请求中的 IP，路由模块自己重新解析一下
现在直接废了

你真的只顾自己的伊朗需求，不考虑广泛使用的透明代理等用途吗？

patterniha:
@Meo597 

right, i should not change the place of `if ips := ctx.Context.GetTargetIPs(); len(ips) != 0 {return ips}`.

the priority of built-in-dns-result should be higher than original target-ip.

Is it okay for you now?

Meo597:
> Is it okay for you now?

np

patterniha:
So we reached an agreement.

Meo597:
Yes, caching errors in a single dispatch is acceptable.

This PR doesn't break any functionality.

And it can resolve the edge case issue you described.

I believe performing a DNS lookup for every IP rule is a code defect.

The lazy query in the source code design was originally intended to reduce latency.

Because if an IP rule cannot be matched due to a parsing error, skipping it and moving to the next IP rule may lead to unexpected behavior.

Meo597 (approved):

RPRX:
~~既然 @Meo597 看过了就先合了然后交给众测吧~~

RPRX:
~~话说伊朗的英文习惯是括号前不加空格吗~~

PR #5243, opened 2025-10-20
DNS: Remove unnecessary checkSystemNetwork()

Meo597:
是否真有必要在每个 DNS 查询都检查 v4 v6 先不提

至少没必要检查 1+n 次

这两行实在太蠢了

### DIFF
--- a/app/dns/nameserver.go
+++ b/app/dns/nameserver.go
@@ -217,11 +217,7 @@ func (c *Client) IsFinalQuery() bool {
 
 // QueryIP sends DNS query to the name server with the client's IP.
 func (c *Client) QueryIP(ctx context.Context, domain string, option dns.IPOption) ([]net.IP, uint32, error) {
-	if c.checkSystem {
-		supportIPv4, supportIPv6 := checkSystemNetwork()
-		option.IPv4Enable = option.IPv4Enable && supportIPv4
-		option.IPv6Enable = option.IPv6Enable && supportIPv6
-	} else {
+	if !c.checkSystem {
 		option.IPv4Enable = option.IPv4Enable && c.ipOption.IPv4Enable
 		option.IPv6Enable = option.IPv6Enable && c.ipOption.IPv6Enable
 	}


patterniha:
you remove this feature completely if we use it in dns-server-settings (not global-dns-settings)

we have two place for `queryStrategy`, one in global-dns-settings, and one in per-dns-server-settings.

we may only set `useSystem` only for one dns-server and not for all-servers (this also has practical applications)

///

it is acceptable to optimize the code to use at most one `checkSystemNetwork` per DNS-LookupIP call, but you remove it completely!!!

```
"servers": [
      {
        "address": "DNS1",
        "queryStrategy": "useSystem"
      },
      {
        "address": "DNS2",
        "queryStrategy": "UseIP"
      }
],
"queryStrategy": "UseIP"      
```

for example `checkSystemNetwork` is not preformed for this config at all !!!

Meo597:
第一，谁给你权力乱关 PR 的，垃圾代码写出来不让人说了?

第二，既然你非要让每个服务器都配置 useSystem，在 LookupIP() 把结果传参进 QueryIP() 就行了

patterniha:
because I am sure that this is wrong, we may set `useSystem` only for one dns-server not all servers.

Meo597:
只是改个if 的问题就能兼容你这种奇葩需求，到你这直接 close PR

好家伙
github 要是交给腾讯运营，你这不得冲个 vip 就把我号给封了

patterniha:
> Just changing the if problem can be compatible with your weird requirements

Ok, just open another PR for that (It's not as easy as you think), i don't have any problem with you I just close a completely-wrong PR.

///

**also, we usually set `useSystem` to global-queryStrategy-settings, so `c.checkSystem` is `false` and only `s.checkSystem` is true, so there is no need for your optimize.**

**so we use `checkSystemNetwork` only once per DNS.LookupIP, even if we have 10 DNS-server, unless someone set `useSystem` in all DNS-servers !**

patterniha:
> so c.checkSystem is false and only s.checkSystem is true

I think you didn't pay attention to this.

Meo597:
你 server 没继承全局，文档没注明

如果用户全局和 server 都开，确实要查 4 次

这玩意就算真的移动端需要，也有很大优化空间

我不像你一样有那么多过剩的精力，你自己来吧

patterniha:
> Your server does not inherit the global, the document does not indicate

all problems is solved when we talk logically, sometimes you're right and sometimes I'm right.

RPRX:
邀请了 @Meo597 加入 org，等下也给个 Write 权限，~~这样就公平了~~

Meo597:
~这不比打游戏刺激~

RPRX:
已给 @Meo597 Write 权限，虽迟但到，~~主要是防止又被关 PR~~

Meo597:
终于不需要靠 force push 重启 test 了
~能打败魔法的只有魔法~

PR #5354, opened 2025-11-27
Update protocol.go

xujiegb:
Refactor EncodeUDPPacket to handle buffer size dynamically and improve error handling.

### DIFF
--- a/proxy/shadowsocks/protocol.go
+++ b/proxy/shadowsocks/protocol.go
@@ -218,19 +218,42 @@ func EncodeUDPPacket(request *protocol.RequestHeader, payload []byte) (*buf.Buff
 	user := request.User
 	account := user.Account.(*MemoryAccount)
 
-	buffer := buf.New()
 	ivLen := account.Cipher.IVSize()
+
+	const maxAddrLen = 259
+
+	tagLen := int32(0)
+	if account.Cipher.IsAEAD() {
+		tagLen = 16
+	}
+
+	need := int32(ivLen) + maxAddrLen + int32(len(payload)) + tagLen
+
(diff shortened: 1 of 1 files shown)


xujiegb:
@fujiwarabach Please check it.

fujiwarabach:
Close this PR

fujiwarabach:
@xujiegb We should leave this problem to @RPRX .

RPRX:
~~啥情况~~

alphonca99:
@fujiwarabach @xujiegb Although other issues caused the UDP packets to be too large, xray-core crashes when the UDP packets exceed 8k, so a PR might be worth trying.

xujiegb:
> @fujiwarabach @xujiegb Although other issues caused the UDP packets to be too large, xray-core crashes when the UDP packets exceed 8k, so a PR might be worth trying.

My main concern is that this change may lead to higher memory usage, so I’m closing this PR.

fujiwarabach:
@fujiwarabach @xujiegb We don't need to investigate this for Arch Linux users; they build their own desktop environments, so let them figure it out themselves. This issue doesn't exist on Fedora Linux or Red Hat Enterprise Linux.

Let me remind you again: if you see ArchLinux users encountering problems, just ignore them.

xujiegb:
> @fujiwarabach @xujiegb We don't need to investigate this for Arch Linux users; they build their own desktop environments, so let them figure it out themselves. This issue doesn't exist on Fedora Linux or Red Hat Enterprise Linux.
> 
> Let me remind you again: if you see ArchLinux users encountering problems, just ignore them.

Okay

Issue #4346, opened 2025-02-03
Solution to the xpadding leak


ghost:
Hello, I'd like to thank all of you for your efforts in improving xray.

In the recent update, you removed the option to disable xpadding, which is a positive step towards reducing the chance of detection. However, it seems you haven't considered that xpadding can cause traffic to be identified on certain CDNs!

I believe one solution to this problem would be to use a random 3 or 4 character string instead of `x_padding`, and also random characters instead of `X` or `0`.

The prerequisite for this is that the client should accept any response header, and the server should accept any query (meaning it should be sensitive to the path but not sensitive to the query parameters).

Thank you again for your efforts. Please think more about my suggestion before dismissing it, as this is important for those of us using specific CDNs.

Fangliding:
https://t.me/projectXtls/670

AiAndNet2030:
> https://t.me/projectXtls/670

I had seen that message, and while it's true for large CDNs like Cloudflare but smaller and country-specific CDNs often try to enforce restrictions using one or a few rules. The presence of `x_padding` is something that helps them do this with a single rule. 

For example, I noticed that on one CDN, adding a standard user-agent like Google Chrome resolved the connection issue.

Therefore, please prevent blocking by randomizing the xpadding.

@Fangliding 
@RPRX

AiAndNet2030:
> > Therefore, please prevent blocking by randomizing the xpadding.
> 
> 我们的意思很明确，等这样的 blocking 出现后再采取措施规避它

As I said, this doesn't happen with well-known and public CDNs like Cloudflare, but it does happen with local CDNs. However, if your intention is for me to wait until they block and then report it, that's fine, and I'll wait. But I wish you paid as much attention to the inner layers as you do to the outer layers, such as adding default headers and so on, because many Xray users use various CDNs.

Thanks for all your work to make X-ray better.

@RPRX

RPRX:
> However, if your intention is for me to wait until they block and then report it, that's fine, and I'll wait.

Exactly.

hexband:
> > Therefore, please prevent blocking by randomizing the xpadding.
> 
> 我们的意思很明确，等这样的 blocking 出现后再采取措施规避它

Since yesterday russian CDN named cdnvideo already throws 403 if request contains x_padding param. This blocking has already been implemented and is in effect, and many people in Russia have now lost internet access.

RPRX:
[...]

PR #5414, opened 2025-12-12
XHTTP transport: New options for bypassing CDN's potential detection

paqx:
Hi!

As you might know, there are currently white lists for mobile internet in many regions of Russia. It's hard set up an xray server to bypass the restriction because you need either a white-listed IP or a white-listed CDN that supports POST requests. This commit targets the latter case. There are few white listed CDNs in Russia and unfortunately they try to block xray requests. I myself bumped into one such case. CDNVideo started to block any POST requests whose query string matches the pattern `x_padding=XXXXX`. Also, some requests included an X-Padding header. Both the query param and header have a very unusual name and can be easily spotted. My goal was to bypass this by allowing users to change the name of the query param. I ran a test and successfully bypassed CDNVideo's filter. It's not a perfect solution because they can start blocking requests by the value but anyway.

What I did:

1. Added xPaddingQueryParam. It can be set in the config to change the name of the query param used to pass the padding data. The default is x_padding
2. Renamed X-Padding header to X-Signature - a lot more common and widely used by actual CDN users.
3. Changed the way the padding data is generated. It now consists of chars X and Z, not only X.

Further improvements:

- Other CDNs disabled POST requests but PUT and PATCH requests are often not blocked. I plan to allow users to choose an HTTP method to pass data.
- The padding value is still very noticable. I don't understand the idea well enough yet but it the post said X was chosen because it's compressed into 8 bits. I checked the encoding table and the only suitable alternative was char Z. Not much and I wonder if it's possible to make the data look more random. In this case it will be very hard to block it because it can disguise as cache busters.
- Use a more realistic user agent. I noticed that the requests are currently passed with the `Go-http-client/2.0` UA. Also doesn't look like a real CDN user. 

Thanks


Fangliding:
这个部分暂时不需要pr 不过它是计划中的 https://github.com/XTLS/Xray-core/issues/4346#issuecomment-3567313166

不用X换用随机ASCII字符的方法非常简单 只需要确定字符集(eg base62)然后计算它在哈夫曼编码下的压缩率然后将预期的xpadding长度除以它做修正就行了 最后的结果会趋近xpadding的长度而且还会额外附加一点随机性 这早在全部用X填充之前我就想到了 不过问题不大而且这个字母占8个位确实很有缘所以当初就没提这一点(当然复杂一点的话精确长度也是完全可以做到的)

hxehex:
> 这个部分暂时不需要pr 不过它是计划中的 [#4346 (comment)](https://github.com/XTLS/Xray-core/issues/4346#issuecomment-3567313166)
> 
> 不用X换用随机ASCII字符的方法非常简单 只需要确定字符集(eg base62)然后计算它在哈夫曼编码下的压缩率然后将预期的xpadding长度除以它做修正就行了 最后的结果会趋近xpadding的长度而且还会额外附加一点随机性 这早在全部用X填充之前我就想到了 不过问题不大而且这个字母占8个位确实很有缘所以当初就没提这一点(当然复杂一点的话精确长度也是完全可以做到的)

I thought the pr was in english, and you are replying in chinese. where is your conscience or respect? what were you thinking when making this reply and then closing this pr?

and then, you did not give an exact valid reason on why you closed this pr. all you said is that you can handle yourself the random generation of something I still don't understand. you ignored the main purpose of this pr, that is clearly in it's title:
# XHTTP: rename X-Padding to X-Signature, add xPaddingQueryParam 

..or if you can't read english, here you go:
# XHTTP：将X-Padding重命名为X-Signature，新增xPaddingQueryParam参数

the #4346 issue you pointed to is dead. nothing was made or changed since (1 month) the last reply in it (not counting mine reply)

hxehex:
also, the xray phrase "Xray, Penetrates Everything." right now is invalid - because it does not "penetrate" CDNvideo

RPRX:
~~我们看得懂英文，回复用中文纯粹是因为对我们来说码字速度更快，没有别的意思~~

我觉得从你改了多个部分的代码以及“Further improvements”的思考来看，你对 Xray 的架构有不浅的研究、完全理解 XHTTP 的工作原理、且有发散性的思维，也就是说你是一个持续改进 XHTTP 的不错的人选，所以我会 reopen 这个 PR，只是怎么改还需讨论

~~就像 Vision Seed 一样，本来在等 GFW 动手，如果 XHTTP 再不加自定义的话等下隔壁出个类似的传输层号称啥都能改，绷不住~~

> Renamed X-Padding header to X-Signature - a lot more common and widely used by actual CDN users.

修改默认名称会破坏 XHTTP 新旧版本之间的兼容性，应当加选项

> Added xPaddingQueryParam. It can be set in the config to change the name of the query param used to pass the padding data. The default is x_padding
> Changed the way the padding data is generated. It now consists of chars X and Z, not only X.

然后关于这两点我思考过 https://github.com/XTLS/Xray-core/issues/4346#issuecomment-3567313166 ，如果要改应当加上七八个选项，或者就干脆不带 padding，你觉得哪一种更好？

RPRX:
> also, the xray phrase "Xray, Penetrates Everything." right now is invalid - because it does not "penetrate" CDNvideo

~~我突然觉得 Xray 这名字挺贴切的，对不同墙的穿透性不同，但毕竟不是 Gamma-ray，尽力而为，如果完全断网那只能星链~~

hxehex:
> > also, the xray phrase "Xray, Penetrates Everything." right now is invalid - because it does not "penetrate" CDNvideo
> 
> ~我突然觉得 Xray 这名字挺贴切的，对不同墙的穿透性不同，但毕竟不是 Gamma-ray，尽力而为，如果完全断网那只能星链~

:face_with_diagonal_mouth: starlink is not available nor it does work in Russia (officially)

hxehex:
> 修改默认名称会破坏 XHTTP 新旧版本之间的兼容性，应当加选项

I guess neither me or @paqx thought about that

I think maybe we should revert that one change but still allow customizing the header name

Fangliding:
> > 这个部分暂时不需要pr 不过它是计划中的 [#4346 (comment)](https://github.com/XTLS/Xray-core/issues/4346#issuecomment-3567313166)
> > 不用X换用随机ASCII字符的方法非常简单 只需要确定字符集(eg base62)然后计算它在哈夫曼编码下的压缩率然后将预期的xpadding长度除以它做修正就行了 最后的结果会趋近xpadding的长度而且还会额外附加一点随机性 这早在全部用X填充之前我就想到了 不过问题不大而且这个字母占8个位确实很有缘所以当初就没提这一点(当然复杂一点的话精确长度也是完全可以做到的)
> 
> I thought the pr was in english, and you are replying in chinese. where is your conscience or respect? what were you thinking when making this reply and then closing this pr?
> 
> and then, you did not give an exact valid reason on why you closed this pr. all you said is that you can handle yourself the random generation of something I still don't understand. you ignored the main purpose of this pr, that is clearly in it's title:
> 
> # XHTTP: rename X-Padding to X-Signature, add xPaddingQueryParam
> ..or if you can't read english, here you go:
> 
> # XHTTP：将X-Padding重命名为X-Signature，新增xPaddingQueryParam参数
> the #4346 issue you pointed to is dead. nothing was made or changed since (1 month) the last reply in it (not counting mine reply)

我完全看得懂英文 反正很多人都要借助翻译软件译成自己的母语我还不如也直接敲中文尽可能保留原始意思
我打一开始就看了这些更改没什么很大的建设性 重命名header的名字和把里面填充的字母从全X改成XZ混合并没有什么意义 要检测也就改几行规则的事情 改名字还顺便破坏兼容性 但是这个header确实打算开放修改 我才贴了那条link

hxehex:
@paqx will reply about this problem a bit later he can't answer right now

paqx:
@RPRX 

Thank you for reopening the PR. Russian users really need these changes at the moment to bypass censorship.

> Based on your modifications to multiple parts of the code and your consideration of "Further improvements," I believe you have a deep understanding of Xray's architecture, a complete grasp of XHTTP's working principles, and divergent thinking. In other words, you are a good candidate for continuous improvement of XHTTP. Therefore, I will reopen this PR; however, how to modify it still needs discussion.

Thanks, but actually I do web developing in PHP/Python + DevOps. Yesterday was the first day I started to dig the xray code and I haven't coded in the go programming language before. Therefore, I hope you will help me implement the things I outlined in the first post.

> Changing the default name will break compatibility between new and old versions of XHTTP; an option should be added.

I will return X-Padding as the default header name for backward compatibility and try to make it possible to change the header name via the config file.

> Then I thought about these two points https://github.com/XTLS/Xray-core/issues/4346#issuecomment-3567313166 . If we were to change it, we should add seven or eight options, or just remove the padding altogether. Which do you think is better?

Could you please explain the idea behind the seven or eight options? In my PR, I implemented the possibility to set any query param name instead of x_padding. Doesn't this allow users to set any name and not chose from just 7-8 options? Or do you mean something else?

As I mentioned, the xray code is new to me. I don't yet understand why some decisions were made. I gathered my knowledge about XHTTP mainly from [this post](https://github.com/XTLS/Xray-core/discussions/4113#discussion-7614138). My current understanding is that the x_padding query param as well as the X-Padding header were added to make the size of HTTP requests more random (so that censors cannot easily detect HTTP requests of the same length and block them). If that's true, I don't think it's a good idea to completely remove the padding. I realize that the primary goal of xray is to bypass censors and not CDNs. However, CDNs in Russia decided to cooperate with censors so I think making xray less detectable for CDNs is necessary too (but not by making it less stealthy for censors).

@Fangliding 

> Renaming the header and changing the padding from all Xs to a mix of XZs is meaningless. Detection would only require changing a few lines of rules. Renaming it also breaks compatibility. But this header is indeed intended to be open for modification, which is why I posted that link.

I agree that my current changes do dot provide enough protection but I needed to start from something. It currently works and allowed me to bypass the CDNVideo filters. They simply added some regex that searches for x_padding query param followed by a sequence of Xs.

I only changed Xs to a sequence of X+Z because  [this post](https://github.com/XTLS/Xray-core/discussions/4113#discussion-7614138) specifically mentions that it was chosen since it gets compressed into 8 bits and it sounded like it's important for xray operation. And the only suitable letter that gets compressed into the same 8 bits is Z. So I used what I could use. It would great if you could explain why the padding data is generated like a sequence of Xs now.

Obviously if CDNVideo notice some other param followed by a sequence of Z+X, they will easily block it again. But if I manage to change the way the padding data is generated and make it look random like some hash, it will be a lot harder for them.

For example, real CDN users can often use params like _dc for cache busting and if the padding looks something like this:
```
_dc=iubwefwe2423bfwebnklewb2b23t23
```

I doubt they will decide to block it because it might cause serious issues for many real users. Isn't it what XHTTP is all about?

> The method of using random ASCII characters instead of 'X' is very simple.  You just need to determine the character set (e.g., base62), calculate its compression ratio under Huffman coding, and then divide the expected xpadding length by this ratio to make the adjustment. The final result will approximate the xpadding length and also add a bit of randomness. I thought of this before using all 'X's for padding, but it wasn't a big problem, and the fact that this character occupies 8 bits was quite coincidental, so I didn't mention it at the time (of course, a more complex method could achieve the exact length).

I will try to implement this in the PR.

Thanks.

gfw-killer:
I think the new options could be
1. the padding position (query-string / header / cookie / multipart)
2. it's key-name
3. padding character-set (repeated X or random A-Z,a-z,0-9)
4. upload method (POST / PUT / PATCH)
it is also possible to upload data using GET request but not all webservers supports it, same about OPTIONS/TRACE/CONNECT/HEAD
There is also 'Resumable Upload APIs' like Tus; https://developers.cloudflare.com/stream/uploading-videos/resumable-uploads/ (with no need to gRPC support))

gfw-killer:
I remember there was another obvious characteristic too
`POST /yourpath/sameUUID/seq`
the connection uuid could also have different position and forms, and also the seq

paqx:
Hi everyone!

I just finished the new padding generation method and also added capability to pass and receive padding in different places. At the same time I tried to preserve backward compatibility. 

The new parameters that I added are:
1. xPaddingObfsMode
2. xPaddingKey
3. xPaddingHeader
4. xPaddingPlacement
5. xPaddingMethod

**xPaddingObfsMode** - for backward compatibility. If's set to false (the default value), the padding will be generated and placed like before.

**xPaddingKey** - the name of the key for storing the padding value. Depending on the xPaddingPlacement, it can be:
1. name of the query param placed inside of a URL which in turn is placed inside of an HTTP header
2. name of the cookie
3. name of the HTTP header
4. name of the query parameter

multipart (I understand it as padding in the request/response body) seemed difficult for me, so I decided to skip it for now.

**xPaddingHeader** - the name of the header. Only makes sense when padding is passed as a header value or a query inside of a header value.

**xPaddingPlacement** - only works when xPaddingObfsMode=true. Defines a place to put the padding (queryInHeader, cookie, header, query).

**xPaddingMethod** - how to generate padding: repeat-x (like now) or tokenish. For tokenish I did some testing and found out that Huffman encoding gives ~20% size reduction for base62 sequences. So I generate a sequence taking into account this compression ratio and then try to adjust it until the validation tolerance (currently 2 bytes) is reached.

Files
- transport/internet/splithttp/browser_client.go
- transport/internet/splithttp/client.go
- transport/internet/splithttp/hub.go
Have easy access to the padding config. 

However, files
- app/dns/nameserver_doh.go
- transport/internet/tls/ech.go
don't. And I just don't know how to pass the padding configs there. So far I simply removed padding from them and my question is whether it's fine. If not, could you suggest better ideas?

I did some testing and the new features seem to work. But it would be great to have the code reviewed.

In the meantime, I will try to implement more upload methods besides POST.

Thanks.

Fangliding:
你可以用这个utlils函数替换其他两个地方的padding（说实话因为header是写死的它们意义不大 可以保持X 不知道为什么要删掉。。）

paqx:
@Fangliding 

Thanks for the comment. If they are not very meaningful, let's revert the changes. I used the helper function that you added and returned the padding to `app/dns/nameserver_doh.go` and `transport/internet/tls/ech.go`.

I also added another parameter. Now it's possible to specify the uplink HTTP method using uplinkHTTPMethod. 

Currently, some Russian CDNs block xray by simply disabling the POST method, but other methods usually work. One of such CDNs is Yandex CDN (and their partner EdgeCDN). Using the new uplinkHTTPMethod I switched to the PUT method and it worked. I also tested the HEAD method and it worked too.

Now I would like to change the user agent in requests to something that looks more like a real user and maybe do something about the UUID and seq in each request URL.

Fangliding:
超级大量的put和head请求比post更加奇怪 本质还是猫鼠游戏

paqx:
I can see that Russian CDNs are blocking xray with care. Examples:

- CDNVideo used a precise regex that matches `x_padding=XXXX`. Simple changing X to a combination of X+Z bypasses the filter. 
- Other CDNs like Yandex only disabled the POST requests, but not PUT, PATCH, TRACE, etc. I don't have any statistics, but as a web developer I haven't seen any websites use CDNs with methods other than GET. Even POST looks unusual to me. In my opinion, they could just leave only the GET method and disable all others in advance. And yet they didn't do it. 

Why? My guess is that real CDN customers might have special needs and may actually use methods other then POST in their web apps. And CDNs can't extend their filters forever. Because at some point it might severely impact their real paying customers.

So I still think the changes make sense.

As a last resort, I thought about using GET requests for uplink (via query or headers). I guess it will be super-slow, but it's better than nothing.

hxehex:
> 超级大量的put和head请求比post更加奇怪 本质还是猫鼠游戏

wdym

gfw-killer:
> 本质还是猫鼠游戏

All this proxy stuff is a cat-and-mouse game, GFW use JA3, proxy devs use uTLS, they find vulnerability to detect uTLS, proxy devs find and patch or invent things Naiveproxy, GFW continues and proxy devs continue too
The difference is only, that when there is no TLS, it's harder, but still not impossible, raising the number of false-positives will do the job

Also a reminder that when user use a CDN that is made by the local government, there must be option to disable the padding completely, as the CDN is literally whitelisted by the local GFW
it's more important to have options to change the connection uuid and seq positions, and even multiple forms with different lengths for the uuid (`POST /yourpath/sameUUID/seq`)

> As a last resort, I thought about using GET requests for uplink (via query or headers). I guess it will be super-slow, but it's better than nothing.

Performance will not be very bad, because upload traffic is low in normal web browsing, except when user uploads a file to somewhere
And your local CDN may support a huge max-header-size, or maybe they set a better size limit on cookies or query-string

Check this one too (I think most CDNs will not support it, but your local CDN may do)
https://www.baeldung.com/cs/http-get-with-body

Fangliding:
> All this proxy stuff is a cat-and-mouse game, GFW use JA3, proxy devs use uTLS, they find vulnerability to detect uTLS, proxy devs find and patch or invent things Naiveproxy, GFW continues and proxy devs continue too
> The difference is only, that when there is no TLS, it's harder, but still not impossible, raising the number of false-positives will do the job

那能一样吗 用utls是为了模仿流量更常见的client hello 这用更少见其他方法代替post 不是弄反了?

gfw-killer:
My view is that having many options will increase the count of false-positives for them and is beneficial
Currently detecting XHTTP does not even need a RegEx, this options will help a lot
If the author make no breaking change and keep the design minimal and clean, it will bother no one

nedogimov-m:
> Check this one too (I think most CDNs will not support it, but your local CDN may do) https://www.baeldung.com/cs/http-get-with-body

A lot of Russian CDNs support this

yuhan6665:
Thank @paqx for your work!
For some reason the reverse test cases are hanging..

paqx:
@yuhan6665 Hi! Thanks for the information. I'll check why it happens when I finish the remaining features I would like to add .

paqx:
I added parameters to change the placement and keys of session Id (uuid) and seq  `(POST /yourpath/sameUUID/seq)`. By default, they will be placed in the path for backward compatibility.

- sessionPlacement - can be path, query, cookie or header.
- sessionKey - the name of the key (not applicable for path)
- seqPlacement - can be path, query, cookie or header. If sessionPlacement is path, seqPlacement must also be path.
- seqKey - the name of the key (not applicable for path)

The last thing I would like to implement in this PR is the ability to use the GET method for uplink transmission and also pass the data in the query, header or body instead of the request body.

RPRX:
~~我其实懒得去设计哪些地方应该怎么改，要不直接在 XHTTP 内允许多人 PR 多种变体得了，只要两端都是 Xray-core 时能连上就行~~

RPRX:
~~然后看哪个变体比较能经过时间的检验，最终会被纳入 XHTTP 官方版的设计，非 Xray 的实现也可以同步过去了~~

paqx:
Finally, I managed to add the possibility to use GET requests for both uplink and downlink. Also, it's possible to pass uplink data in headers or cookies instead of the request body (not actually required for now, but I feel like disabling any requests except for GET is the first thing CDNs will think of when they find out any non-GET method can be used for uplink).

For placing data in headers or cookies, there's the new uplinkDataPlacement parameter (can be header or cookie). Initially, I wanted to make it possible to use the query for passing uplink data, but it appeared to be useless. The encoded data is too large and the CDN I tested it with simply dropped the requests. With headers and cookies I was able to bypass this by splitting data into several headers/cookies. 

And I made fixes for the failing tests.

I think I've done everything I wanted to do in this PR.

@RPRX @Fangliding Could you please review the changes. Let me know if you have any suggestions, ideas or threats (just kidding).

Thanks

RPRX:
~~看样子应该不是伊朗人，有一位伊朗贡献者 @hossinasaadi 几天没出现过了~~

hxehex:
> 几天没出现过了

https://radar.cloudflare.com/ir
https://x.com/CloudflareRadar/status/2009312093749801350

internet in iran is nonexistent right now because there is a civil war going on

RPRX:
@paqx 下个稳定版本会带上这个 PR 以及正式版 Vision Seed

zeretrelle:
> I think I've done everything I wanted to do in this PR.

> 
> Thanks

Amazing job!  Thank you for all your efforts. 

Would it be too much to ask for some information on how to use these new settings, perhaps with a few examples?

Cheers

paqx:
@zeretrelle 

> Would it be too much to ask for some information on how to use these new settings, perhaps with a few examples?

No, I was actually going to write a guide on how to use the new parameters. I'll do it soon.

paqx:
# Rationale

Russian mobile operators block normal access to the Internet. Only some whitelisted websites are accessible. Others are either completely blocked or severely throttled. There are two primary methods to bypass these restrictions:

1. **Use a VPS with a whitelisted IP** and masquerade traffic as a whitelisted website.
2. **Proxy traffic through a whitelisted CDN** using Xray-core's XHTTP protocol.

However, CDNs started to detect and block XHTTP traffic. Currently, the detection is straightforward:
*   **CDNVideo** blocks traffic by the distinctive x-padding (a long sequence of the character `X`).
*   **Other CDNs** (e.g., Yandex Cloud, VK Cloud) have disabled POST requests (used by XHTTP for uplink data transmission).

# Solution

This PR adds new parameters to the XHTTP protocol configuration to make it harder to detect and block its traffic.

**New Configuration Parameters:**
*   `xPaddingObfsMode`
*   `xPaddingKey`
*   `xPaddingHeader`
*   `xPaddingPlacement`
*   `xPaddingMethod`
*   `uplinkHTTPMethod`
*   `sessionPlacement`
*   `sessionKey`
*   `seqPlacement`
*   `seqKey`
*   `uplinkDataPlacement`
*   `uplinkDataKey`
*   `uplinkChunkSize`

---

# How to Use the New Parameters

## 1. Bypassing Filtering by X-Padding

If a CDN blocks traffic based on the x-padding, the goal is to:
1.  Rename the `x-padding` parameter.
2.  Make its value look like a random token.

### Parameter Description

*   **`xPaddingObfsMode`** (`boolean`, default: `false`)
    Enables padding obfuscation. Defaults to `false` for backward compatibility.

*   **`xPaddingKey`** (`string`)
    The name of the key that stores the padding value. Its meaning depends on `xPaddingPlacement`:
    *   Name of a query parameter inside a URL (if placement is `queryInHeader`).
    *   Name of a cookie.
    *   Name of an HTTP header.
    *   Name of a URL query parameter.

*   **`xPaddingHeader`** (`string`)
    The name of the HTTP header. Only relevant when `xPaddingPlacement` is `header` or `queryInHeader`.

*   **`xPaddingPlacement`** (`string`)
    Defines where to put the padding value. Possible values: `queryInHeader`, `cookie`, `header`, `query`. **Only works when `xPaddingObfsMode` is `true`.**

*   **`xPaddingMethod`** (`string`)
    Defines how to generate the padding value.
    *   `repeat-x`: The default method (a long sequence of `X` characters).
    *   `tokenish`: Generates a random Base62 token.

### Example

Stick with key names commonly used by real CDN users to make it difficult for the CDN to filter XHTTP traffic.

```json
{
    "network": "xhttp",
    "security": "tls",
    "xhttpSettings": {
        "host": "<YOUR_CDN_RESOURCE_URL>",
        "path": "/some-path",
        "xPaddingObfsMode": true,
        "xPaddingMethod": "tokenish",
        "xPaddingPlacement": "queryInHeader",
        "xPaddingKey": "_dc",
        "xPaddingHeader": "X-Cache"
    }
}
```

**Result:** The padding will be passed as a query parameter inside a URL which in turn is placed in a request header:
```
X-Cache: https://<YOUR_CDN_RESOURCE_URL>/some-path/?_dc=9cfMQ4q8zo6WgC5pYYQd6Il6J1jY6xHQnPYsP22zV8HQipoufDF6krZF2nSlRk9HKJliuuYrYCkSrtt7T5aCD5ZbLGU2J6Se6JUCVSxCDVy2vghVbWh2n1SBuRMn67mHu08S4CUWlQTa5RptDi15OlN5nIN5JK1a4nw04L05jA1eaJypgaBr2rEJjeza1vFgGmpUfbKqqJIYFSZQdAO1rPMa74sTUF3Xi1T1HmPYBSfS7UEqxFMnJENxMun2RxYMTx3cKxLpbBzwKvkL45ZHT8LTDFY4YMIffjx5NYca6JzvazJQCkTSb5qKowrKy7j6zV6uV95t9kXWED41uLH4Vb06zbZqHx9HlHkD5ZaI3BHyfkrVbgkVvXumYiUf1mbh7o3XcaOub6cLaWTfnWHkHOoVrGMPkZvt39UtQynqL7BnppnG38dvN5A1bAMrbUXaJxClZ4jvXdEthgieSHIXrXrqvFAbLMCTAYp7bK5EtrMMRyw4tKtDB8TPLAS40CRDplDPtij47SBf0E9YzOP377ElhzOREZRYf9DrBvLy9ixFKMBv99lXkirUccWKWipzqqDUtEcAsSItGKTfJStO9keFhr26OhzdJhiVpKD8ZK8eTtAcGfuIXq8bnWK6bJDp2w7WBIXM8FYhqsXkghdJ8xYGI8eMqQhdrGrRET2dmp7nlAFj86KxomfKP
```

---

## 2. Bypassing Filtering by Uplink HTTP Method

If a CDN blocks XHTTP by disallowing POST requests, you can switch to a different HTTP method.

### Parameter Description

**`uplinkHTTPMethod`** (`string`)
Changes the HTTP method used for uplink data transmission. Use any method that supports a request body (e.g., `PUT`, `PATCH`) ans is allowed by the CDN.

**Example:**
```json
{
    "network": "xhttp",
    "security": "tls",
    "xhttpSettings": {
        "host": "<YOUR_CDN_RESOURCE_URL>",
        "path": "/some-path",
        "uplinkHTTPMethod": "PUT"
    }
}
```

---

## 3. Bypassing Future Potential Filtering

Current CDN filters are relatively simple, but that might change. The following parameters are intended to further help disguise XHTTP traffic.

### Hiding Session ID and Sequence Number

XHTTP appends a session UUID and sequence number to the path (e.g., `/path/507d9107-8354-4ccf-b174-96387334b3b0/2`). The following parameters can be used to move and rename these values.

*   **`sessionPlacement`** (`string`): Where to place the session ID. Options: `path`, `query`, `cookie`, `header`.
*   **`sessionKey`** (`string`): The key name for the session ID (not applicable if `placement` is `path`).
*   **`seqPlacement`** (`string`): Where to place the sequence number. Options: `path`, `query`, `cookie`, `header`. **If `sessionPlacement` is `path`, `seqPlacement` must also be `path`.**
*   **`seqKey`** (`string`): The key name for the sequence number (not applicable if `placement` is `path`).

Stick with key names commonly used in real CDN traffic.

**Example:**
```json
{
    "network": "xhttp",
    "security": "tls",
    "xhttpSettings": {
        "host": "<YOUR_CDN_RESOURCE_URL>",
        "path": "/some-path",
        "sessionPlacement": "header",
        "sessionKey": "X-Auth-Token", // Optional. Default: `X-Session` for header, `x_session` for cookie/query.
        "seqPlacement": "query",
        "seqKey": "page" // Optional. Default: `X-Seq` for header, `x_seq` for cookie/query.
    }
}
```

### If Only GET is Available

If a CDN only allows the `GET` method and uplink data cannot be sent in a request body, it is possible to use the `GET` method for both downlink and uplink data transmission. The uplink data will be passed as cookies or HTTP headers.

**Important:** This requires the `mode` to be explicitly set to `"packet-up"` on **both client and server** configurations.

*   **`uplinkDataPlacement`** (`string`): Where to place the split uplink data fragments. Options: `cookie` or `header`.
*   **`uplinkDataKey`** (`string`): The base name for the key(s) used to pass the data fragments. The client will automatically split the data into multiple chunks and the server will reassemble it.
*   **`uplinkChunkSize`** (`uint32`): Maximum size of each uplink data chunk in bytes (only applicable when `uplinkDataPlacement` is not `body`).
    - **Minimum:** 64 bytes
    - **Recommended values:**
      - For `"cookie"` placement: **3072 bytes** (3KB default)
      - For `"header"` placement: **4096 bytes** (4KB default)
    - When not specified, appropriate defaults are automatically selected based on `uplinkDataPlacement`.
   
**Example:**
```json
{
    "network": "xhttp",
    "security": "tls",
    "xhttpSettings": {
        "host": "<YOUR_CDN_RESOURCE_URL>",
        "path": "/some-path",
        "mode": "packet-up", // MUST be set explicitly when uplinkDataPlacement is not `body`.
        "uplinkDataPlacement": "header",
        "uplinkDataKey": "X-Payload", // Optional. Default: `X-Data` for header, `x_data` for cookie.
        "uplinkChunkSize": 4096
    }
}
```

RPRX:
~~其实序号可以加密一下，现在是这样做的吗~~

paqx:
> The serial number could actually be encrypted; is that what's being done now?

No, it's not done at the moment. I considered encryption when I was making the changes, but I then realized that it adds too much complexity and doesn't bring any significant benefits. Even if the session ID and seq values are passed as plain-text, you can still make it really hard to understand if they belong to XHTTP and not to real CDN users.

## Examples

### Headers for SeqKey:
- `X-Content-Range`
- `Segment-Number`
- `Chunk-Number`
- `X-Part`

### Query/Cookie Parameters for SeqKey:
- `range` or `r`
- `part`
- `segment` or `seg`
- `chunk`
- `offset`
- `fragment`

### Headers for Session ID:
- `X-Session-ID` or `X-Session-Id`
- `X-Request-ID`
- `X-Client-ID`
- `X-Auth-Token`
- `X-API-Key`
- `X-User-ID`

### Query/Cookie Parameters for Session ID:
- `session_id` or `sid`
- `token`
- `auth_token`
- `api_key`
- `key`
- `signature`
- `expires`
- `policy`
- `st` or `s_t`
- `visitor_id`
- `client_id`
- `user_token`
- `auth`
- `remember_me_token`

And you can make countless combinations:

1. Pass session ID as the `X-Client-ID` header and seq as the `chunk` cookie:
```json
{
    "xhttpSettings": {
        "sessionPlacement": "header",
        "sessionKey": "X-Client-ID",
        "seqPlacement": "cookie",
        "seqKey": "chunk"
    }
}
```
2. Pass session ID as the `visitor_id` cookie and seq as the `Segment-Number` header:
```json
{
    "xhttpSettings": {
        "sessionPlacement": "cookie",
        "sessionKey": "visitor_id",
        "seqPlacement": "header",
        "seqKey": "Segment-Number"
    }
}
```
3. Pass session ID as the `signature` query param and seq as the `fragment` query param:
```json
{
    "xhttpSettings": {
        "sessionPlacement": "query",
        "sessionKey": "signature",
        "seqPlacement": "query",
        "seqKey": "fragment"
    }
}
```

And so on. Your fantasy is the only limitation.

Seems to me that's more than enough to masquerade XHTTP traffic well enough without encryption. What do you think?

gfw-killer:
Thank you @paqx
It seems like xPaddingObfsMode option can be removed, when default settings are compatible with old settings, there is no need to xPaddingObfsMode option
And there should be a limit on `uplinkDataKey` like the early data of WebSocket https://github.com/XTLS/Xray-core/pull/375
If it does reuse the same H2 connection, the performance of it should be good

i mentioned early data of websocket! it could be good if we have options to change it's placement and key too
other clients already support a custom header key for websocket early data
https://wiki.metacubex.one/config/proxies/transport/?h=early#__tabbed_1_4
but xray have no option for it yet!

paqx:
@gfw-killer 

Thanks for the comment.

Now that I am reviewing my own code again I can see it's probably possible to drop the xPaddingObfsMode parameter (originally I thought it would make the code much cleaner). I will try to do it.

> And there should be a limit on uplinkDataKey like the early data of WebSocket https://github.com/XTLS/Xray-core/pull/375
> If it does reuse the same H2 connection, the performance of it should be good

I don't quite catch it yet. I added two parameters: uplinkDataPlacement and uplinkDataKey. If the uplinkDataPlacement is either cookie or header:
1. the client takes the request body, splits it into chunks and appends to the request as cookies or headers. 
2. The uplinkDataKey is used as the cookie/header names. If the uplinkDataKey, for example, is set to `x_data`, then the first chunk is put into the x_data-0, the second chunk into the x_data-1 and so on.
3. The request body is removed.

The existing ScMaxEachPostBytes parameters already sets the size of the request body, so it seems like it's already possible to:
1. set the size of the uplink data
2. set the data placement (body, cookies, headers)
3. set the names of the data cookies/headers.

Or do you mean something else?

> i mentioned early data of websocket! it could be good if we have options to change it's placement and key too
> other clients already support a custom header key for websocket early data
> https://wiki.metacubex.one/config/proxies/transport/?h=early#__tabbed_1_4
> but xray have no option for it yet!

I assume it's something not related to disguising the padding in XHTTP, but a similar thing that could be implemented for the websocket mode? If yes and there's a need for such parameters, I could try to add them.

gfw-killer:
> If the uplinkDataKey, for example, is set to x_data, then the first chunk is put into the x_data-0, the second chunk into the x_data-1 and so on.

I mean there should be an option to change the chunk max size
because header max size limit of each cdn/reverse proxy is different

> The existing ScMaxEachPostBytes parameters already sets the size of the request body

so client will use the size of ScMaxEachPostBytes as the max size for each chunk?

> I assume it's something not related to disguising the padding in XHTTP, but a similar thing that could be implemented for the websocket mode? If yes and there's a need for such parameters, I could try to add them.

Sorry, yes it's not related to this PR, but it has the same goal
It's about Early Data feature of WebSocket (and HTTPUpgrade) transport
The fixed header name is "Sec-WebSocket-Protocol" in Xray-core
It's placement is good, it will be better if we have an option to change it's header name

KiGamji:
> also, the xray phrase "Xray, Penetrates Everything." right now is invalid - because it does not "penetrate" CDNvideo

По инсайдам от "многоуважаемого" Роскомнадзора и работающих там людей — 

> Yandex Cloud и VK Cloud в этом году не ограничат от глобального инета; отрубать от глобального инета серверы не будут пока просто потому что в РФ нет нормального оборудования для CDN (звучит как-то странно, не знаю), а те кто может его поставить - накинули санкции

Это вполне может говорит о том, что рубильник для зарубежного интернета именно на CDN, которые находятся в так называемых "белых спискам" — могут дёрнуть в любой момент, впрочем, как и на домашнем интернете. И в то же время это уже исключительно наша забота, потому что в Китае GFW такие действия себе позволить не может) Но промолчать не смог.

KiGamji:
@paqx @hxehex Nonetheless, with the recent news of whitelisted subnets being banned for VPS assignment purposes—anything I could help you with in that PR?

RPRX:
~~内幕消息：我和 @iambabyninja 正在开发一个新的传输层，预计很快就会与大家见面，后续也会类似 XHTTP 有多种模式~~

paqx:
> I mean there should be an option to change the chunk max size
> because header max size limit of each cdn/reverse proxy is different

I can add a dedicated parameter to define a chunk size. I didn't do it because I had already introduced a lot of new parameters and I wanted to bloat the configuration. Currently, I hard-coded the chunk sizes: 4KB if data is passed via headers and 3KB if data is passed via cookies. These chunk sizes for headers and cookies should be safe for most CDNs.

> so client will use the size of ScMaxEachPostBytes as the max size for each chunk?

Not exactly. The client will use ScMaxEachPostBytes as the max size of all data passed in a single request:
1. If the data is passed in the request body, the data is not split in chunks (current default behavior);
2. If the data is passed via headers, it's split into 4KB chunks.
3. If the data is passed via cookies,  it's split into 3KB chunks.

> Sorry, yes it's not related to this PR, but it has the same goal
> It's about Early Data feature of WebSocket (and HTTPUpgrade) transport
> The fixed header name is "Sec-WebSocket-Protocol" in Xray-core
> It's placement is good, it will be better if we have an option to change it's header name

Got it. Ok, I can take a look at it.

> Nonetheless, with the recent news of whitelisted subnets being banned for VPS assignment purposes—anything I could help you with in that PR?

@KiGamji Thanks for the interest. All in all, the PR is already complete. I implemented parameters that can be used to bypass current CDN filters (Yandex Cloud, VK Cloud, EdgeCenter, CDNVideo) and several parameters that can help bypass future potential filters (that I would implement if I were a CDN). So currently it's more about refactoring than adding something new. For example, @gfw-killer suggested to remove the `xPaddingObfsMode` parameter. I originally added it to make it easier for me to implement the new features.

> По инсайдам от "многоуважаемого" Роскомнадзора и работающих там людей —
> 
> Yandex Cloud и VK Cloud в этом году не ограничат от глобального инета; отрубать от глобального инета серверы не будут пока просто потому что в РФ нет нормального оборудования для CDN (звучит как-то странно, не знаю), а те кто может его поставить - накинули санкции
> 
> Это вполне может говорит о том, что рубильник для зарубежного интернета именно на CDN, которые находятся в так называемых "белых спискам" — могут дёрнуть в любой момент, впрочем, как и на домашнем интернете. И в то же время это уже исключительно наша забота, потому что в Китае GFW такие действия себе позволить не может) Но промолчать не смог.

I don't really believe they will ever disconnect the Russian internet from the global internet. Maybe for a few days, but not forever. It will simply kill the already slowly dying economy. Because most goods are manufactured from imported components. Without the Internet, how is a factory supposed to order the components? Using a phone and hard-copy catalogs received by mail? Nobody gave me a viable answer to this question yet.

KiGamji:
> I don't really believe they will ever disconnect the Russian internet from the global internet. Maybe for a few days, but not forever. It will simply kill the already slowly dying economy. Because most goods are manufactured from imported components. Without the Internet, how is a factory supposed to order the components? Using a phone and hard-copy catalogs received by mail? Nobody gave me a viable answer to this question yet.

Да им по сути своей наплевать, Hetzner, AWS, Cloudflare побанили как эксперимент, не произошло ничего, дальше пошли отключения мобильного интернета. А пустить в перспективе чистый белый список для всех, кроме "своих" и корпоративного сегмента — как будто бы уже ничего и не мешает.

---

They don't really care at this point. They banned Hetzner, AWS, Cloudflare as an experiment — nothing happened, so they moved on to mobile internet shutdowns. And rolling out a strict whitelist for everyone except "their own" and the corporate sector — seems like nothing's stopping them anymore.

RPRX:
~~虽然我习惯打中文但大家能否尽量用英文，因为俄文看不懂，tor 浏览器没有自带翻译~~

KiGamji:
> ~虽然我习惯打中文但大家能否尽量用英文，因为俄文看不懂，tor 浏览器没有自带翻译~

Fair, sorry about that. I was just discussing some local (Russian) context with a couple of people here — thought it would be easier in Russian. Actually, I'd also just include an English translation next time, if that works? Would be nice if Chinese comments could get the same treatment when possible, though...

RPRX:
~~只是一个个人请求，带上俄文的英译会很有帮助，因为 tor 浏览器没有内置 Chrome 那种自带翻译，对于中文你们直接开翻译就行~~

KiGamji:
Well, we're actually not all on Chrome either, for obvious reasons 😅 Would be great if all comments could get translations too when there's a mixed audience

RPRX:
> ~内幕消息：我和 @iambabyninja 正在开发一个新的传输层，预计很快就会与大家见面，后续也会类似 XHTTP 有多种模式~

https://github.com/XTLS/Xray-core/pull/5581#issuecomment-3797134147

paqx:
@RPRX 

Using a cloud storage as a proxy sounds like an awesome idea. Maybe I could later do some research to see if something similar is possible with Russian cloud storages like Yandex Disk.

Thanks.

KiGamji:
> Using a cloud storage as a proxy sounds like an awesome idea. Maybe I could later do some research to see if something similar is possible with Russian cloud storages like Yandex Disk.
> 
> Thanks.

`@nexts` DM in telegram in case you would like some help and testing.

gfw-killer:
> I can add a dedicated parameter to define a chunk size. I didn't do it because I had already introduced a lot of new parameters and I wanted to bloat the configuration. Currently, I hard-coded the chunk sizes: 4KB if data is passed via headers and 3KB if data is passed via cookies. These chunk sizes for headers and cookies should be safe for most CDNs.

there is no other choice, there must be many options to make the protocol more flexible and support more scenarios
It's a need to let us to select the chunk limit, because it's not same in all CDNs and will also change in the future
You can even add an option to change the HTTP response status code of the server, or set multiple codes and randomize it in each request
Or change uplinkHTTPMethod to accept multiple values, and randomize requests between them to make them less suspicious
And there will be no problem with all this options, for proxy providers it's features, and as we can share them in Extra of XHTTP, there is no need for GUI Clients to change anything, they just need to update the core

And i have some questions about the tests you made to select this default values
Before you do the tests, have you increased the limit of Xray & Nginx max accepted header size
And have you tried to use multiple chunks in to the same or different placements in a single GET request to carry the maximum possible data in a GET request? if there be a 4kb limit on a single header or cookies, and a 64kb overall limit on headers, it does matter

paqx:
> there is no other choice, there must be many options to make the protocol more flexible and support more scenarios
> It's a need to let us to select the chunk limit, because it's not same in all CDNs and will also change in the future

Ok, I added the new `uplinkChunkSize` parameter and updated the documentation above (https://github.com/XTLS/Xray-core/pull/5414#issuecomment-3770071786).

> You can even add an option to change the HTTP response status code of the server, or set multiple codes and randomize it in each request
> Or change uplinkHTTPMethod to accept multiple values, and randomize requests between them to make them less suspicious

Sounds like a good idea. I could add these capabilities later.

> Before you do the tests, have you increased the limit of Xray & Nginx max accepted header size

No, I didn't even change the settings of Xray & Nginx. In my early tests, I put the uplink data in a single chunk into the query param, a header and a cookie. And tried to proxy the traffic via a CDN. But it all failed because the size of the data was too big. The CDN just didn't didn't let the data through and dropped it. And this is how I realized I need to split the uplink data into chunks and spread it across multiple cookies/headers. I used some reasonable chunk sizes (3KB for cookies and 4KB for headers) and it worked.

> And have you tried to use multiple chunks in to the same or different placements in a single GET request to carry the maximum possible data in a GET request? if there be a 4kb limit on a single header or cookies, and a 64kb overall limit on headers, it does matter

Do you mean whether I always sent data in either headers or cookies, or tried to use both cookies+headers at the same time? if yes, then I didn't try to use both cookies+headers. I used either headers or cookies. In both cases everything worked, so I didn't dig deeper. Although spreading data between both cookies+headers could be possible (if it's necessary to pass more data in a single request, e.g. to make uploads faster).

As I understand, the limits are different for different servers, proxies and browsers:
https://stackoverflow.com/questions/686217/maximum-on-http-header-values
https://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key

The idea is to ensure that in your specific case you don't exceed the limits of the strictest intermediary proxy (the one with the smallest header or cookie size limits).

gfw-killer:
Thank you ❤️❤️❤️❤️
In Iran we really need this PR to merge and also https://github.com/XTLS/Xray-core/pull/5560 @RPRX 

GFW enables a hard mode after 1-2 pm then after 5-6 pm they enable the hell mode until 1 am
they use all the possible ways to detect proxies, asn, port, fingerprint, connection count, reconnect time, init connection size, single connection traffic and living time & etc.
Even the Real Delay feature of apps that automatically tests each proxy in the subscription is used to detect and block proxies
~If grpc had the features and performance of xhttp, it could help more~
I mean the GFW operators really thinks out of the box! they know all the possible characteristics of each protocol/transport and even apps

RPRX:
~~如果你们觉得没什么问题了的话倒是可以先合并先发版，改动太多懒得看了~~

RPRX:
~~我的意思是改动挺多的，比起来现在讨论还需要哪些改动、该怎么设计，不如先发版给有需要的人用上，以后再讨论设计~~

没有涉及新配置项时不影响原版 XHTTP 连接吧？测试一下 @paqx

paqx:
@RPRX 

I think the current parameters should suffice for now. Everything else could be added in separate PRs (if necessary).

I tried to make everything backward compatible. That's why I added the xPaddingObfsMode. By default, it's set to false and the padding should be added as before. In this case it should be compatible with older clients, though I haven't actually tested it. I can set up a client with my new changes and a server without them (and vice versa) to see if they can communicate properly.

RPRX:
@paqx 那你测试一下吧

boyboytemp:
I've done a limited testing and found an error with one particular configuration:
Server w/o changes:  mode: stream-one,  all other defaults
Client from this branch: mode: stream-one, all other defaults
An error is printed on every connect and nothing works:
`2026/01/29 16:38:15.996633 [Info] transport/internet/splithttp: stream-up mode is not allowed`

paqx:
@boyboytemp 

Could you please share redacted configs of the client and server? Also, have you tried the same configs without the changes in this PR?

boyboytemp:
> Also, have you tried the same configs without the changes in this PR?

Yes it works when I use the old client.

Server version: 26.1.13

Server is behind Nginx (Nginx config is default from examples):

<details>

<summary>server.json</summary>

``` json
{
  "dns": {
    "servers": [
      "https+local://8.8.8.8/dns-query",
      "https+local://1.1.1.1/dns-query",
      "https+local://8.8.4.4/dns-query",
      "8.8.8.8",
      "8.8.4.4",
      "localhost"
    ]
  },

  "log": {
    "access": "/var/log/xray/access.log",
    "error": "/var/log/xray/error.log",
    "loglevel": "debug"
  },
  "inbounds": [
    {
      "listen": "/dev/shm/xrxh.socket,0666",
      "protocol": "vless",
      "settings": {
        "clients": [
          {
            "id": "1"
          }

        ],
        "decryption": "none"
      },
      "streamSettings": {
        "network": "xhttp",
        "xhttpSettings": {
          "mode": "stream-one",
          "path": "1"
        }
      }
    }
  ],
  "outbounds": [
    {
      "tag": "direct",
      "protocol": "freedom",
      "settings": {
        "domainStrategy": "UseIPv4v6",
        "fragment": {
          "packets": "tlshello",
          "length": "50-200",
          "interval": "2-15"
        }
      }
    },
    {
      "tag": "blocked",
      "protocol": "blackhole",
      "settings": {}
    }
  ],
  "routing": {
    "domainStrategy": "AsIs",
    "rules": [
      {
        "type": "field",
        "ip": [
          "geoip:private"
        ],
        "outboundTag": "blocked"
      }
    ]
  }
}
```
</details>


<details>

<summary>client.json</summary>

``` json
{
  "log": {
    "loglevel": "debug"
  },
  "dns": {
    "servers": [
      {
        "address": "https+local://1.1.1.1/dns-query"
      },
      {
        "address": "https+local://8.8.8.8/dns-query"
      },
      "8.8.8.8",
      "9.9.9.9",
      "1.1.1.1.",
      "localhost"
    ]
  },
  "inbounds": [
    {
      "tag": "socks",
      "port": "14000",
      "listen": "0.0.0.0",
      "protocol": "socks",
      "settings": {
        "udp": true
      }
    }
  ],
  "outbounds": [
    {
      "tag": "vless-xhttp",
      "protocol": "vless",
      "settings": {
        "vnext": [
          {
            "address": "1",
            "port": 443,
            "users": [
              {
                "id": "1",
                "encryption": "none"
              }
            ]
          }
        ]
      },
      "streamSettings": {
        "network": "xhttp",
        "xhttpSettings": {
          "path": "1",
          "mode": "stream-one"
        },
        "security": "tls"

      }
    },
    {
      "tag": "direct",
      "protocol": "freedom",
      "settings": {}
    },
    {
      "tag": "blocked",
      "protocol": "blackhole",
      "settings": {}
    }
  ],
  "routing": {
    "domainStrategy": "IPOnDemand",
    "rules": [
    ]
  }
}
```
</details>

boyboytemp:
I've also noticed that `xPadding*` settings in `downloadSettings` should be in sync with uplink settings from`xhttpSettings`.
For example, I use different `xPaddingPlacement` in this configuration:

```json
"xhttpSettings": {
  "mode": "stream-up",
  "xPaddingObfsMode": true,
  "xPaddingMethod": "tokenish",
  "xPaddingPlacement": "cookie",
  "xPaddingKey": "id",
  "downloadSettings": {
    "network": "xhttp",
    "xhttpSettings": {
      "xPaddingObfsMode": true,
      "xPaddingMethod": "tokenish",
      "xPaddingPlacement": "header",
      "xPaddingKey": "X-FOO",
    }
    
  }
}
```
It results in an error: 

`2026/01/29 17:30:34.397826 [Info] transport/internet/splithttp: invalid padding (queryInHeader=X-Padding, key=client_id) length:0`

Up http request: `Header: cookie: id=...`
Down http request: `Header: x-padding: ...` Note that the header name is not `x-foo` as I requested.
When I put settings in sync everything works fine. Except that it seems that `xPaddingKey` seems to be ignored since it is always `x-padding` for me when using `"xPaddingPlacement": "header"`.

paqx:
@boyboytemp 

Thanks! I will take a look and make a fix.

RPRX:
上下行分离的配置是无关的，请不要改变这个定义，比如你上行过某个 CDN、下行过另一个 CDN，但它们适用不同的 XHTTP 变体

RPRX:
@paqx 这个 PR 可以合并了吗

paqx:
> The configuration for separating upstream and downstream traffic is irrelevant; please do not change this definition. For example, even if you upstream through one CDN and downstream through another, they may still apply to different XHTTP variants.

Ok, I let's not change it.

> @paqxCan this PR be merged?

@boyboytemp pointed out (https://github.com/XTLS/Xray-core/pull/5414#issuecomment-3817865901) that there's some issue with the stream-one mode and I haven't checked it yet. I would suggest to postpone merging until I figure out why it fails. Otherwise, there's a chance that existing stream-one set-ups may break.

RPRX:
@paqx 测试一下吧，~~明天发版也不是不行~~

gfw-killer:
@RPRX Could you please change the Stream-one content-type to gRPC in both sides to support single connection over CDNs with gRPC support?
A problem of Stream-up is that it makes 2 connection at the same time and GFW already used this behavior to detect the connection
I know we can use different SNI and IP for Downlink and Uplink, but it's troublesome and when one of them connections is bad the performance will be very low
and if i use gRPC transport, it's featureless and has bad performance
as Stream-one is mainly for direct connection, it will make no problem

RPRX:
@gfw-killer 我没记错的话 stream-one 的默认 header 就是 gRPC，stream-up 即使是 maxConcurrency 1 也是同一条 TCP/QUIC

gfw-killer:
You are right! stream-one works for CF! Thank you

paqx:
@RPRX Added a fix for the stream-one mode. I did some quick testing and it seems to work.

@boyboytemp Could you please double-check?

dvzubarev:
> @boyboytemp Could you please double-check?

It works, thank you!



>  The configuration for separating upstream and downstream traffic is irrelevant; please do not change this definition. For example, even if you upstream through one CDN and downstream through another, they may still apply to different XHTTP variants.
> 
> Ok, I let's not change it.

I'm not sure that I understand translation. But I think that if it is required that `up.xPaddingPlacement` == `down.xPaddingPlacement` then it might be better to check it after configuration loading. Because now when they are different nothing works. But I agree that this is minor thing and can be done later. 

@paqx Could you please look into why `xPaddingKey` does not affect the header name when `xPaddingPlacement == header`.
Given this configuration:

```json
        "xhttpSettings": {
          "path": "poo",
          "mode": "stream-up",
          "xPaddingObfsMode": true,
          "xPaddingMethod": "tokenish",
          "xPaddingPlacement": "header",
          "xPaddingKey": "X-FOO",
          "sessionPlacement": "header",
          "sessionKey": "X-BAR",
          "uplinkHTTPMethod": "PUT"
} 
```
I look into headers that arrives to the server via wireshark and there is no `x-foo` header but the default one `x-padding` instead.

iCHAPALAH:
> @paqx Could you please look into why `xPaddingKey` does not affect the header name when `xPaddingPlacement == header`.

The header name is defined by 'xPaddingHeader' parameter

Avaritia-ZOV:
**CDN in Russia is dead.**

Why did you whine and ask for functionality if it is publicly available? You won yourself a month of CDN on the whitelist, and then there will be requirements for legal entities to obtain CDN. Keep it up! 
Did you really need to do that now, when there are other methods available? You have literally **destroyed** your last chance in the future.

KiGamji:
не буду тебя огорчать, но глобальный интернет в РФ, по слухам, держат ещё из-за их почти полного отсутствия (CDN). так что нет, хоть это и не панацея на ближайшие месяцы, но обесценивать чужой труд 一 это глупо.

---

I won't upset you, but as it happens, the global internet in russia is still standing precisely because near-total absence of CDN services, or so the rumors go. So no, even if it's not a silver bullet, dismissing other people's work is just dumb.

KiGamji:
> when there are other methods available?

suggest any, except for fighting and paying other authorities for a whitelisted CIDR range

Avaritia-ZOV:
> Я не буду вас расстраивать, но, как оказалось, глобальный интернет все еще стоит именно из-за их почти полного отсутствия, или так ходят слухи.

That's ridiculous, I haven't heard anything funnier today. 
It's so completely absent that anyone can go and buy themselves a Yandex VPS and get around the whitelists)). 
And you didn't understand the essence of my words either.
The fact that the CDN bypass has been made publicly available will mark the imminent end of CDN for individuals, nothing more.

KiGamji:
> It's so completely absent that anyone can go and buy themselves a Yandex VPS and get around the whitelists

bro is living under a rock or something

KiGamji:
you do realize the whitelisted ranges are now prohibited for sale by legal authorities, right? and that being said, the CIDR whitelists aren't centralized in any way

KiGamji:
that means, if your yandex VPS suddenly works in Rostov with T2, it'd NOT work in St. Petersburg with Yota. and that is not just from me making up things, you either ask people who sell that kind of stuff, or I provide screenshots myself. CDNvideo is literally the widest working option right now, and they are NOT giving whitelisted IP addresses anymore

hxehex:
please read my russia-mobile-internet-whitelist repo readme, it covers the most basics that you should know

RPRX:
@paqx XDRIVE 有开工吗，~~没有的话我就开始 vibe 了~~

paqx:
@RPRX 

Please start implementing XDRIVE. I am a little busy these days and cannot work on it yet.

RPRX:
@Fangliding ~~要不你来 vibe 一下 XDRIVE，新月新配额~~

Fangliding:
你们先前的讨论我都没看 顶多出来我再看看
~~不是说已经和谁计划好了吗 怎么还是新建文件夹啊~~

KiGamji:
I could help, but cba, just messing with TUN for now... ~~Please take a look at my closed PR when you have time and give any suggestions~~

RPRX:
@Fangliding XDRIVE 的先前讨论全在 https://github.com/XTLS/Xray-core/pull/5581 ，往下翻

代码也有了，Google Drive https://github.com/XTLS/Xray-core/pull/5581#issuecomment-3812422766 ，S3 stores https://github.com/paqx/s3proxytest ，把它们弄成 Xray 的传输层就行

Fangliding:
过年回家我还要带孩子 我平时在core写的代码都是不vibe的(

这么大一个个的SDK真的打算弄进来吗 这目测下可比之前的acme大多了

Fangliding:
稍微看了一下 可能的问题是无视本来的目标和无视sockopt

Meo597:
我正在苦哈哈的重写 geomatcher
看到你们都在 vibe
突然就不想努力了

RPRX:
> 过年回家我还要带孩子 我平时在core写的代码都是不vibe的(

~~看一下接不接吧，接的话我邮件给你发那些被删掉的 token，可以测试一下~~

> 这么大一个个的SDK真的打算弄进来吗 这目测下可比之前的acme大多了

~~其实手写都行~~

> 稍微看了一下 可能的问题是无视本来的目标和无视sockopt

还有 security 和 finalmask 也不适用，服务端入站也不 listen，不过 dialer-proxy 和 domain-fronting 可以搞一下

K088228:
~~其实手写都行~~

I checked—the code works without the SDK, using only the REST API. It doesn’t look very pretty, but it’s reasonably readable, even though there’s a lot of boilerplate.

~~看一下接不接吧，接的话我邮件给你发那些被删掉的 token，可以测试一下~~

Those tokens are no longer valid. When you need credentials, I can generate new ones or provide you with a few Google accounts.

paqx:
> Are they really planning to include all these huge SDKs? They look much bigger than the previous ACME.

I used the SDK because I needed to quickly test if it's possible to use an S3 store for proxying. But then also mentioned that they would make the binaries too bloated and that it would be better to use REST API instead.

obezbashka:
> это значит, что если ваш VPS от Яндекса вдруг заработает в Ростове с T2, то он НЕ будет работать в Санкт-Петербурге с Yota. И это не просто мои домыслы, вы можете спросить у людей, которые продают такие вещи, или я сам предоставлю скриншоты. CDNvideo — это буквально самый широкий из доступных вариантов, и они БОЛЬШЕ НЕ предоставляют IP-адреса из белого списка


Они все еще предоставляют адреса в белых списках, и еще десятки cdn, часть из которых можно обойти лишь открыв на точке  выхода вебсокеты, реализовав реверс прокси(раннее это работало на edge cdn, сейчас же это не работает) если у нас будет возможность пропускать xhttp не только по post запросам, то остальное дело техники, сигнатура xhttp по post запросам сильно палится.

но к тому же, ркн злоупотребляет не только этим, хостеров в виде вк клауда и яндекса дополнительно сканили, сканил грчц и откидывал абузы, так же это затронуло и CDN сервисы, я так и не понял, был ли это сканнинг, ибо cdn отлетели через неделю после вк клауда и яндекса, так что похоже что публичные списки впн серверов, которые лежат на гитхабе, просто отфильтровали по IP и пронесли абузами.

hippo2025:
@paqx Can the new parameters be passed to the "extra" field? This sounds like a stupid question, but somehow this doesn't work for me. If I just copy parameters into extra field, nothing works. What am I doing wrong? Can you please give an example where your parameters are passed into extra field?

hann-wang:
> @paqx Can the new parameters be passed to the "extra" field? This sounds like a stupid question, but somehow this doesn't work for me. If I just copy parameters into extra field, nothing works. What am I doing wrong? Can you please give an example where your parameters are passed into extra field?

https://github.com/XTLS/Xray-core/blob/main/infra/conf/transport_internet.go#L264-L275

Those params work in the "extra" field. If the extra field exists, the "xhttpSettings" should only contain "host", "path", "mode" and "extra".

maoxikun:
没有完整的客户端和服务端example吗？双端的xhttpSettings都按照此PR中的示例配置连接成功，但是nginx的access日志中很多408没人发现吗？
```json
{
    "network": "xhttp",
    "xhttpSettings": {
        "host": "<YOUR_CDN_RESOURCE_URL>",
        "path": "/some-path",
        "xPaddingObfsMode": true,
        "xPaddingMethod": "tokenish",
        "xPaddingPlacement": "queryInHeader",
        "xPaddingKey": "_dc",
        "xPaddingHeader": "X-Cache"
    }
}
```

fish4terrisa-MSDSM:
> 没有完整的客户端和服务端example吗？双端的xhttpSettings都按照此PR中的示例配置连接成功，但是nginx的access日志中很多408没人发现吗？
> 

#5739 這個裡面有完整配置
https://github.com/XTLS/BBS/issues/25#issuecomment-4270761269

maoxikun:
> #5739 這個裡面有完整配置 [XTLS/BBS#25 (comment)](https://github.com/XTLS/BBS/issues/25#issuecomment-4270761269)

客户端和服务端都写在xhttpSettings里面也会生效（Referer中的padding消失并且能连上），但是nginx的access日志中上行POST请求会出现408（使用此PR的配置多访问几个网站几分钟之内就能复现）

实测放extra里面一样408

xx-xxxx-xxxx:
其实对于CDN来说，同一个uri大量的POST返回的body为空，GET返回大段内容，也是一个非常明显的特征。建议上下行path uri可以分离，并且可以将POST返回的随机填充内容放入body

zakachkin:
Есть мысль по поводу `uplinkHTTPMethod` в XHTTP.

На мой взгляд, было бы удобнее и хитрее не использовать один фиксированный метод для всех upload-запросов, а дать возможность чередовать несколько методов, например `PUT` и `PATCH`.

Условно:

```json
"uplinkHTTPMethods": ["POST", "PUT", "PATCH"]
```

или даже с исключениями:

```json
"uplinkHTTPMethods": ["POST", "PUT", "PATCH"],
"disabledUplinkHTTPMethods": ["PUT"]
```

Смысл в том, что разные CDN по-разному пропускают методы. Где-то работает только `POST`, где-то может проходить `PATCH`, а `PUT` может отдавать `405`. Было бы удобно указывать список разрешённых методов, а Xray мог бы выбирать/чередовать их для upload-запросов.

Например, если CDN не поддерживает `PUT`, пользователь просто добавляет его в excluded/disabled list, и XHTTP продолжает использовать только рабочие методы.

На мой взгляд, такая рандомизация или ротация методов могла бы сделать XHTTP-поведение менее шаблонным и более похожим на обычный API-трафик, особенно при работе через CDN.

Ahmad-2213:
XHTTP needs to be documented & Updated , Everytime we should look for PRs to see what's changed...
~It's a headache~

Ahmad-2213:
Arvancloud CaaS has Also restricted the Methods to GET only as I've observed , all other methods return NOT ALLOWED 405.
The Setup I use is XHTTP + XTLS + TLS  (plain http is forbidden in Forough DC ~which makes RPRX really happy xD~) 
And it seems that with body + GET , mlkem is troublesome and sometimes not handled correctly by nginx ingress & CDN, Furthermore,  if the UplinkPlacement is header/cookie, I Always encounter The header is too large 431, even with different parameters & offloading session/padding to query; in this situation where UplinkPlacement is header/cookie, some sites like chatgpt.com fail to open with the same header error; IDK if the reason is mlkem but what i do know is that it's not working properly.

Ahmad-2213:
> XHTTP needs to be documented & Updated , Everytime we should look for PRs to see what's changed... ~It's a headache~

@Meo597 Can you do this favour?

hippo2025:
> Arvancloud CaaS has Also restricted the Methods to GET only as I've observed , all other methods return NOT ALLOWED 405. The Setup I use is XHTTP + XTLS + TLS (plain http is forbidden in Forough DC ~which makes RPRX really happy xD~) And it seems that with body + GET , mlkem is troublesome and sometimes not handled correctly by nginx ingress & CDN, Furthermore, if the UplinkPlacement is header/cookie, I Always encounter The header is too large 431, even with different parameters & offloading session/padding to query; in this situation where UplinkPlacement is header/cookie, some sites like chatgpt.com fail to open with the same header error; IDK if the reason is mlkem but what i do know is that it's not working properly.

I had the same problem. Notice that uplinkChunkSize only limits the size of *one* header, but the total size of all headers in one GET query can be as large as scMaxEachPostBytes. Don't know about nginx, but with Caddy I could only make this work with scMaxEachPostBytes also around 4096, not much more.

It's sad this is mentioned nowhere, everyone will run into this same issue.

You might also wish to set scMinPostsIntervalMs to lower value, because with default settings Xray can only send 33 GET queries per second and we have to limit scMaxEachPostBytes. I haven't experimented with this though.

Ahmad-2213:
@hippo2025 
Thanks for genuine reply , I've been playing with scMaxEachPostBytes , though it's title is a little confusing because you don't have only POST anymore.
As I tested , the problem solved with scMaxEachPostBytes: 4096, however, it's so resource intensive , but it'll make it work.
Thanks for reaching out.

hippo2025:
> @hippo2025 Thanks for genuine reply , I've been playing with scMaxEachPostBytes , though it's title is a little confusing because you don't have only POST anymore. As I tested , the problem solved with scMaxEachPostBytes: 2048, however, it's so resource intensive , but it'll make it work. Thanks for reaching out.

You can also change nginx settings to allow larger total size of headers in GET queries. But be careful, too large can be blocked by CDN.

Ahmad-2213:
> > @hippo2025 Thanks for genuine reply , I've been playing with scMaxEachPostBytes , though it's title is a little confusing because you don't have only POST anymore. As I tested , the problem solved with scMaxEachPostBytes: 2048, however, it's so resource intensive , but it'll make it work. Thanks for reaching out.
> 
> You can also change nginx settings to allow larger total size of headers in GET queries. But be careful, too large can be blocked by CDN.

Snippets are not allowed and disabled by CDN.
the crucial part for this modification to work properly is to configure "scMaxBufferedPosts" aswell accordingly, we're reducing the maxsize , therefore this value should be upscaled so the connection won't be disconnected immaturely.
we need to keep some balance between resource usage and these terms .

Meo597:
> > XHTTP needs to be documented & Updated , Everytime we should look for PRs to see what's changed... ~It's a headache~
> 
> @Meo597 Can you do this favour?

Unlike @Fangliding who maintains all the documentation and core PRs, I only focus on the parts I use. Unfortunately, I haven't used xhttp to this day.

Ahmad-2213:
> > > XHTTP needs to be documented & Updated , Everytime we should look for PRs to see what's changed... ~It's a headache~
> > 
> > 
> > @Meo597 Can you do this favour?
> 
> Unlike @Fangliding who maintains all the documentation and core PRs, I only focus on the parts I use. Unfortunately, I haven't used xhttp to this day.

I appreciate it, thanks.

PR #5417, opened 2025-12-13
Wireguard: Decouple server endpoint DNS from address option

Meo597:
Previously, Wireguard server endpoint's domain resolution was incorrectly constrained by the local `address` option. For example, `ForceIPv6v4` might fail to resolve AAAA records for the server domain if no IPv6 was explicitly configured in the `address` option.

This commit decouples the server endpoint's domain resolution from the local `address` configuration. It ensures the Wireguard server address is resolved independently, allowing its `domainStrategy` to function correctly without being limited by the client's local network or `address` settings.


yuhan6665:
Yeah I think you are right. I was copying this piece from the process() and didn't pay too much attention. Thanks for the fix!
You can remove the code instead of comment out

Meo597:
> You can remove the code instead of comment out

done

Issue #5066, opened 2025-08-27
Suppress "REALITY localAddr: X DialTLSContext" Logs


nopoz:
Is there a way to suppress these logs on the client? These lines are clogging up stdout on the client and they don't add any value as far as I can tell.

Client stdout:

```
REALITY localAddr: 10.1.2.3:62573	DialTLSContext
REALITY localAddr: 10.1.2.3:62573	DialTLSContext
REALITY localAddr: 10.1.2.3:62642	DialTLSContext
REALITY localAddr: 10.1.2.3:62644	DialTLSContext
REALITY localAddr: 10.1.2.3:62641	DialTLSContext
REALITY localAddr: 10.1.2.3:62643	DialTLSContext
```

Log settings on the client:

```
  "log": {
    "access": "none",
    "loglevel": "warn"
  }
```

If there is no way to suppress currently, can this be added? 

Thanks.

RPRX:
这是 REALITY Spider X 的日志，代表收到了目标网站的真证书，原因可能是你的流量被重定向了或者你被 MITM 了，警惕你的运营商

RPRX:
这个是没检查 config.Show 的，忘了当初是不是故意的，我想了下也不用改，如果有人试图攻击你，则 REALITY 必须要提醒你

PR #5427, opened 2025-12-15
REALITY client: Clearer log when receiving real certificate

ari-ahm:
The fmt.Printf call inside UClient's DialTLSContext was executing unconditionally.
This caused log pollution when using xray-core as a library dependency, printing to stdout even when debug mode was disabled.(REALITY localAddr: xxx	DialTLSContext)

This patch wraps the print statement with `if config.Show`, consistent with other debug logs in the file.
This cleans up the output when using the package as a library.

### DIFF
--- a/transport/internet/reality/reality.go
+++ b/transport/internet/reality/reality.go
@@ -180,7 +180,7 @@ func UClient(c net.Conn, config *Config, ctx context.Context, dest net.Destinati
 		fmt.Printf("REALITY localAddr: %v\tuConn.Verified: %v\n", localAddr, uConn.Verified)
 	}
 	if !uConn.Verified {
-		errors.LogError(ctx, "REALITY: peer verification failed (potential MITM or redirection)")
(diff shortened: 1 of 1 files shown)


RPRX:
如果有人看到了这个 log 那说明有福报了 https://github.com/XTLS/Xray-core/issues/5066#issuecomment-3229899426

要改成先判断 if show 的话要在外面再加个 error 级别的警报，比如你的连接被重定向了/你正在遭受中间人攻击

ari-ahm:
I have updated the code based on your feedback.

I added errors.LogError outside the spider loop. This ensures the user receives a clear, high-severity warning about the verification failure (potential MITM/Redirection) exactly once.
I kept the if config.Show guard on the internal DialTLSContext log.

To be honest, the raw DialTLSContext print message didn't mean much to 99% of users and was mostly seen as log spam. By replacing the repetitive prints with a single, clear Error log, we keep the "blessing" of the security warning, but in a format that users can actually understand and act upon.

PR #5436, opened 2025-12-17
feat: optional DNS pinning for XHTTP w/ downloadSettings

0x3mp7y:
This PR fixes XHTTP issue when main XHTTPSettings and it's downloadSettings have the same domain but DNS returns different A records that points to different servers, for example when using DNS Load Balancing.

- Add opt-in per-context DNS “pin” so repeated dials to the same domain reuse the first chosen IP (transport/internet).
- Enable pinning only for XHTTP when downloadSettings.address is missing / empty  or "same" value, and allow downloadSettings to inherit the primary address while keeping other settings (transport/internet/splithttp).
- Handle reused HTTP connections (pin based on actual remote IP once a connection is obtained).
- Add unit tests for pinning and for different values ("same" or empty address).

### DIFF
--- a/transport/internet/dialer.go
+++ b/transport/internet/dialer.go
@@ -225,6 +225,15 @@ func checkAddressPortStrategy(ctx context.Context, dest net.Destination, sockopt
 
 // DialSystem calls system dialer to create a network connection.
 func DialSystem(ctx context.Context, dest net.Destination, sockopt *SocketConfig) (net.Conn, error) {
+	// Opt-in DNS pinning (only when ctx carries a pin store).
+	pinDomain := ""
+	if dest.Address.Family().IsDomain() {
+		pinDomain = dest.Address.Domain()
+		if ip, ok := getDNSPin(ctx, pinDomain); ok {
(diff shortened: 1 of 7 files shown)


Fangliding:
唔 AI垃圾

0x3mp7y:
So what's wrong? Issue is there, I checked everything and don't see any issues in my current implementation. Tho it's really assisted with AI

Test files and comments can be removed on merging.

Fangliding:
写的非常丑而且是非必要的 xhttp上下行分流有义务让它的目标最后指向相同的端点 这些分散的数据被一起发送有助于进一步打乱特征 你破坏了这些（它们会在最常见的场景比如CDN中出现）
协议在设计的时候不考虑对端多IP是多个端点的情况 你的场景非常少见 为了它而加这么多无用的东西是不合适的

0x3mp7y:
Ok I got your point of view, but it's still could be useful in some scenarios with multiple endpoints behind DNS name + downloadSettings. I'll try to short the code as much as possible without creating any new files etc...

PR #5437, opened 2025-12-17
XHTTP: fix DNS multiple endpoints w/ downloadSettings

0x3mp7y:
Added optional parameter "same" for xhttp's downloadSettings.address so it fix splitting download and upload to multiple endpoints behind single DNS record and as a result that leads to not functional connection.

I do get it that is works right because up/down streams are separate dealers so that's why A records resolved independently, but this still can be useful for DNS load balancing while keeping other download settings (for example different xmux parameters for download and upload).

I rewrote the code and removed all useless code from previous PR (thought at first this can be useful in future with other things).

### DIFF
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -9,6 +9,7 @@ import (
 	"net/http/httptrace"
 	"net/url"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -39,6 +40,13 @@ var (
 	globalDialerMap    map[dialerConf]*XmuxManager
 	globalDialerAccess sync.Mutex
 )
+
+type pinState struct {
+	domain string
+	ip     net.IP
+}
+
+type pinKey struct{}
 
 func getHTTPClient(ctx context.Context, dest net.Destination, streamSettings *internet.MemoryStreamConfig) (DialerClient, *XmuxClient) {
 	realityConfig := reality.ConfigFromStreamSettings(streamSettings)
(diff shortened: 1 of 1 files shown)


Fangliding:
I still don't think this is appropriate

0x3mp7y:
I agree it can be done via another third party tools, but I don't understand what's wrong there. It doesn't touch current logic in anyway. So why don't add it. Only pros and no cons. @RPRX what's your opinion?

RPRX:
大致看了一下，上下行分离如果填同一个域名但有多个 DNS 记录时确保让上下行走同一个 IP 吗，这没有必要，会影响分离效果

**但反方向是有必要的，比如说故意让它们走不同 IP 或一边 IPv4 另一边 IPv6，你有兴趣的话可以 PR 一下**

还可以实现简化配置，在非上下行分离配置的基础上加一个选项即可自动上下行分离走不同 IP

0x3mp7y:
The current testings I made shows way better performance (that's also was mentioned in example usage in XHTTP thread) for using multiple xmux connections for client's inbound traffic and a single connection for outbound. So there's a reason to split up/down even for single address. Also it can be used within single address, but with different SNIs. Many possibilities. Maybe code is not perfect, that's why I allowed edit by maintainers in a first place

> 大致看了一下，上下行分离如果填同一个域名但有多个 DNS 记录时确保让上下行走同一个 IP 吗，这没有必要，会影响分离效果
> 
> **但反方向是有必要的，比如说故意让它们走不同 IP 或一边 IPv4 另一边 IPv6，你有兴趣的话可以 PR 一下**
> 
> 还可以实现简化配置，在非上下行分离配置的基础上加一个选项即可自动上下行分离走不同 IP

PR #5455, opened 2025-12-20
TLS ECH: Increase DOH timeout

patterniha:
because most DOH are blocked, we should use fragment/domainFronting/other-proxy dialer for DOH to obtain the ech key.

but using fragment/other-proxy dialer causes more delay, so 5 seconds timeout may not be enough.

~According to some reports in the Persian group, using fragment with less than 2 seconds delay does not work on some ISPs in Iran.~


Fangliding:
直接改默认值就好了 5秒是当时从直连的角度随手写的

PR #5468, opened 2025-12-23
common/uuid: fix panic when parsing 32-len invalid UUID string.

ari-ahm:
his PR fixes a bug in uuid.ParseString where specific malformed UUID strings could cause the application to panic due to an index out of range error.

The problem specifically occurred on string lengths of 32. This is because anything longer than 32 bytes results in a larger initial capacity for the byte slice, which inadvertently prevented the panic (though the logic was still flawed). However, with exactly 32 bytes, the missing bounds check caused a crash when processing the byte groups if the format didn't match expectations.

Updated ParseString to verify slice bounds before accessing text[0] or slicing text[:byteGroup].
Added a unit test in uuid_test.go covering a malformed UUID string (missing the last character) to ensure it returns an error instead of panicking.

panic message on new test with old uuid.go code :
```
--- FAIL: TestParseString (0.00s)
panic: runtime error: slice bounds out of range [:12] with capacity 8 [recovered, repanicked]

goroutine 9 [running]:
testing.tRunner.func1.2({0x778060, 0xc00001a2b8})
        /usr/lib/go/src/testing/testing.go:1872 +0x237
testing.tRunner.func1()
        /usr/lib/go/src/testing/testing.go:1875 +0x35b
panic({0x778060?, 0xc00001a2b8?})
        /usr/lib/go/src/runtime/panic.go:783 +0x132
github.com/xtls/xray-core/common/uuid.ParseString({0x7a357e, 0x20})
        /home/arian/Codes/Xray-core/common/uuid/uuid.go:92 +0x40a
github.com/xtls/xray-core/common/uuid_test.TestParseString(0xc0001188c0)
        /home/arian/Codes/Xray-core/common/uuid/uuid_test.go:48 +0x25c
testing.tRunner(0xc0001188c0, 0x7aef68)
        /usr/lib/go/src/testing/testing.go:1934 +0xea
created by testing.(*T).Run in goroutine 1
        /usr/lib/go/src/testing/testing.go:1997 +0x465
exit status 2
FAIL    github.com/xtls/xray-core/common/uuid   0.006s
```

### DIFF
--- a/proxy/vless/encoding/encoding.go
+++ b/proxy/vless/encoding/encoding.go
@@ -93,7 +93,7 @@ func DecodeRequestHeader(isfb bool, first *buf.Buffer, reader io.Reader, validat
 
 		if request.User = validator.Get(id); request.User == nil {
 			u := uuid.UUID(id)
-			return nil, nil, nil, isfb, errors.New("invalid request user id: %s" + u.String())
+			return nil, nil, nil, isfb, errors.New("invalid request user id: " + u.String())
 		}
 
 		if isfb {


RPRX:
顺便 https://github.com/XTLS/Xray-core/pull/5426#issuecomment-3685940916

yuhan6665:
Thanks looks good to me!

PR #5479, opened 2025-12-30
Enhance router configuration with route caching features

clcc2019:
- Added route caching options to the Config message in config.proto, including route_cache, route_cache_ttl, and route_cache_max_size.
- Updated the Router struct to support caching, including initialization and cache invalidation logic.
- Implemented cache retrieval and storage in the routing logic to optimize performance.
- Added methods for cache statistics and invalidation in the Router.

This update improves performance by allowing cached routing results, reducing the need for repeated rule evaluations.

### DIFF
--- a/app/router/config.pb.go
+++ b/app/router/config.pb.go
@@ -7,12 +7,13 @@
 package router
 
 import (
+	reflect "reflect"
+	sync "sync"
+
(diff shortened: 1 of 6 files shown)


Fangliding:
不要在这里提交AI生成的代码

PR #5513, opened 2026-01-09
Fix FullCone NAT: safe cone assertion, race condition, cleanup

Jolymmiles:
- P0: Safe type assertion for cone context value with default true
- P1: Fix race condition - create cancel before goroutine spawn
- P1: Add defer-based cleanup in udpWriter for panic safety
- P1: Add debug logging when FindRoute fails
- P2: Replace common.Must with tolerant error handling in cleanup
- P2: Extract named udpConnEntry struct for maintainability

Preserves FullCone semantics: 2-tuple routing, 5-min timeout, return address rewriting, XUDP GlobalID support.

### DIFF
--- a/proxy/tun/handler.go
+++ b/proxy/tun/handler.go
@@ -28,6 +28,15 @@ import (
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 )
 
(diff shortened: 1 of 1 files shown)


Fangliding:
这就别凑热闹了吧

PR #5521, opened 2026-01-11
Proxy: Tun: Implement idle connection timeouts

Owersun:
Consider PolicyManager configuration

### DIFF
--- a/proxy/tun/handler.go
+++ b/proxy/tun/handler.go
@@ -2,6 +2,7 @@ package tun
 
 import (
 	"context"
+	"time"
 
 	"github.com/xtls/xray-core/common"
 	"github.com/xtls/xray-core/common/buf"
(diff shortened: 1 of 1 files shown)


RPRX:
~~从大规模用 DispatchLink() 开始 idle timeout 都是交给 outbound 了~~

Owersun:
ok. perfect, then I'll just implement net.Conn for udp, and depend on that dispatcher will call .Close() on it when idle timeout is reached.

RPRX:
不过 UDP 的话那个“自动清理来源二元组 map”还得留着，因为 outbound 非 UoT 时可能不会把 close 传导过来，~~我不确定~~

Owersun:
Well, adding more changes don't make sense then... The whole idea was that udp will be pure net.Conn, only implementing Read/Write/Close/... and lifecycle is controlled the handler, which will close connections that don't have any bytes for some time...

RPRX:
~~可能非 UoT 时也有 idle timeout close 吧，我真的不确定，@yuhan6665 @Fangliding~~

Owersun:
Tested it with xHTTP outbound, outbound signal the connections to be closed without a need for additional lifecycle handling on inbound side.

RPRX:
~~我想起来了，DispatchLink() 靠的是 return 来传导 close，所以只要 outbound 有 idle timeout 就能传导过来~~

PR #5525, opened 2026-01-12
TUN inbound: Make udp_fullcone pure side effect free udp connection

Owersun:
This makes udp_fullcone usage independent side effect free implementation of net.Conn
This allows to substitute other imeplentations ip stacks could use, but with custom logic allowing FullCone NAT
The code is clearly split to the places it belong, and handler/stack_gvisor/udp_fullcone don't depend on each other and don't know about each others implementation as they should be.
Both stack and connection are nice 200-300 line files.
Connection handling was remove from udp_fullcone and delegated to be part of lifecycle handler should do.
Handler will close connections calling .Close() on net.Conn, when outbound signal that the connection is complete, which will make udp connections close and clean up.
Running for 12 hours with two devices behind a router shows proper gofunc cleanup, keeping the number of running coroutines going up and down with connection number.

PR #5531, opened 2026-01-13
TUN inbound: Close connection when handling is done

Owersun:
dispatcher.DispatchLink calls conn.Close() (or rather common.Interrupt()) on incoming connection only if error occurs.
conn.Close() is not called in case of successful handling (connection finishes or completes by outbound policy idle timeout), when DispatchLink return with no error and HandleConnection function just ends.
This breaks lifecycle handling, and leave connections (both TCP/UDP) dangling in gVisor while already closed from upstream side.
Add defer conn.Close(), always signalling handled connection to close to the input (tun) side, send finish packets (TCP) and cleanup handling (UDP).

### DIFF
--- a/proxy/tun/handler.go
+++ b/proxy/tun/handler.go
@@ -102,6 +102,10 @@ func (t *Handler) Init(ctx context.Context, pm policy.Manager, dispatcher routin
 
 // HandleConnection pass the connection coming from the ip stack to the routing dispatcher
 func (t *Handler) HandleConnection(conn net.Conn, destination net.Destination) {
+	// when handling is done with any outcome, always signal back to the incoming connection
+	// to close, send completion packets back to the network, and cleanup
+	defer conn.Close()
+
 	sid := session.NewID()
 	ctx := c.ContextWithID(t.ctx, sid)
 


RPRX:
~~确定没问题的话我就先合了~~

RPRX:
~~幸好还没发版，躲过一次背刺~~

Owersun:
~yeah, sorry for that. we did so many changes so quick, that it's hard to properly do long testing of it. things slipped. when there are many connections open, if I look at prof there is like 500 more coroutines and then 300 less in few minutes. which looks "fine", but actually wasn't.
I spend yesterday and today testing and found out that there is no .Close() call on sucessfull finish, and it is required for proper lifecycle (otherwise gVisor don't know that connection is no longer open)
If you wait with the release one more day, so I'll monitor the situation with this change, see how prof performs, how many are connections there and how my devices perform - that would be perfect~

RPRX:
~~没事，今天先发版让大家测测吧，隔几天再发一版也是常态~~

RPRX:
@Owersun https://github.com/XTLS/Xray-core/pull/5525#issuecomment-3738108497 有进展吗

Owersun:
I was monitoring the implementation that was release for whole last week.
After this change it all look proper and good, this change was final cherry on the top, and the implementation is proper after that, as designed. No crashes, no issues.
I'll start looking at how to implement simple routing next week.

PR #5551, opened 2026-01-16
README.md: Add Happ RU to iOS & macOS Clients

mangustyura:


iambabyninja:
@RPRX, @Fangliding following the recent removal of the happ app from the Russian App Store region (https://github.com/XTLS/Xray-core/discussions/4759) a separate version was released to ensure continued access to the service without requiring users to change their region or switch Apple IDs.
Accordingly, there are now two versions available: one for the global market and one for the local market.

Issue #5079, opened 2025-09-01
Outbound Interface problem


gfw-killer:
I have a Warp Outbound
```
    {
      "tag": "warp",
      "protocol": "freedom",
      "streamSettings": {
        "sockopt": {
          "interface": "warp"
        }
      }
    }
```
and route some Domains to it, to change IP
Problem is that when wg-quick fails, 'warp' outbound connects directly using my server IP and leaks my server IP to those routed websites

* routing some domains is an example, i use it for more, that is why i use wg-quick instead of xray wg outbound, for a better performance

xqzr:
```shell
curl --interface warp ip.sb
```
work?

patterniha:
set log to `debug/info` and check if you see: "failed to apply socket options".

The code doesn't return error when `applyOutboundSocketOptions` return error(for example, when "interface" is not available) and it only shows log (it continues to run without setting "interface"):

https://github.com/XTLS/Xray-core/blob/4064f8dd8080b0e278df79e0a42c0b153b1d445e/transport/internet/system_dialer.go#L133-L135

Fangliding:
要怪怪go的RawConn.Control 没法返回error吧
当然可以捕获 但是捕获的之后关掉这个conn也已经dial出去了 可能破测试好看了点不过还是掩耳盗铃级 建议用 sendThrough 指定localIP没有这个问题

gfw-killer:
> curl --interface warp ip.sb
> work?

Problem occurs when interface does not exist or gets removed while Xray is running, Outbound must fail instead of using another Interface and leak the proxy server IP

> It is recommended to use sendThrough to specify localIP without this problem

I was hopeful to `sockopt interface` get fixed, but an alternative is also good. if it was possible to set interface name in sendThrough everyone could easily switch to it as a safe alternative to the unsafe `sockopt interface`

PR #5629, opened 2026-01-31
SocketConfig: Fail fast on unavailable network interface

KiGamji:
Validate interface at config load; return non-retryable error at runtime.

Fixes CPU/memory leak caused by infinite retry loop when bound interface is unavailable.

### DIFF
--- a/common/retry/retry.go
+++ b/common/retry/retry.go
@@ -1,12 +1,16 @@
 package retry // import "github.com/xtls/xray-core/common/retry"
 
 import (
+	stderrors "errors"
 	"time"
 
 	"github.com/xtls/xray-core/common/errors"
 )
 
 var ErrRetryFailed = errors.New("all retry attempts failed")
+
+// ErrNoRetry stops retry immediately when wrapped in error chain.
+var ErrNoRetry = errors.New("no retry")
 
 // Strategy is a way to retry on a specific function.
 type Strategy interface {
@@ -27,6 +31,9 @@ func (r *retryer) On(method func() error) error {
 		err := method()
 		if err == nil {
 			return nil
+		}
+		if stderrors.Is(err, ErrNoRetry) {
+			return err
 		}
 		numErrors := len(accumulatedError)
 		if numErrors == 0 || err.Error() != accumulatedError[numErrors-1].Error() {
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"math"
+	gonet "net"
 	"net/url"
 	"runtime"
 	"strconv"
@@ -1214,6 +1215,12 @@ func (c *SocketConfig) Build() (*internet.SocketConfig, error) {
 		happyEyeballs.Interleave = c.HappyEyeballsSettings.Interleave
 		happyEyeballs.TryDelayMs = c.HappyEyeballsSettings.TryDelayMs
 		happyEyeballs.MaxConcurrentTry = c.HappyEyeballsSettings.MaxConcurrentTry
+	}
+
+	if c.Interface != "" {
+		if _, err := gonet.InterfaceByName(c.Interface); err != nil {
+			return nil, errors.New("invalid interface: ", c.Interface).Base(err)
+		}
 	}
 
 	return &internet.SocketConfig{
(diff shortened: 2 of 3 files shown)


Fangliding:
如果有无限循环问题应该修复loop的地方
retry据我所知应该都只有5次而已

KiGamji:
@Fangliding
The retry itself is limited to 5, yes. The issue is that clients keep sending new requests, each spawning 5 retries with log output. With interface down, you get thousands of failing connections per minute, each doing 5 retries — that's the resource drain.

Two separate fixes here:
1. Validate interface at config load — fail early, don't start with invalid config
2. Don't retry on "interface not found" — it won't appear in 100ms, just fail the connection immediately

This isn't about fixing a loop, it's about failing fast on a known-unrecoverable error.

KiGamji:
> With interface down, you get thousands of failing connections per minute, each doing 5 retries — that's the resource drain.

image

Fangliding:
如果问题在此 那任何导致它连接可能失败的错误最后都会导致此类问题

KiGamji:
Network errors (timeout, connection refused, etc.) are transient — retry in that cases makes sense, the next attempt may succeed.

"Interface not found" is not transient — if the interface doesn't exist, retrying 100ms later won't help. Same for config validation: why allow Xray to start with an interface that doesn't exist...?

This specifically targets non-recoverable errors, not all connection failures.

KiGamji:
> why allow Xray to start with an interface that doesn't exist...?

That being said, if the interface disappears during runtime — as a misconfiguration or a service restart — it will simply drop the connections unless it is reconnected, avoiding CPU and memory leaks. That also matches the normal non-TUN behavior.

Fangliding:
那就只启动时检测接口吧 这个err太丑了

Fangliding:
有大量的配置错误都不可恢复 没有意义

KiGamji:
> 那就只启动时检测接口吧 这个err太丑了

It's just that startup validation alone won't help if the interface disappears at runtime (e.g. VPN reconnect, Wi-Fi toggle). That's when the leak happens...

I am open to suggestions on a cleaner approach for the runtime case, I guess

Fangliding:
可以为retry添加一个全局的速率限制 只是需要一个合适的令牌桶值

KiGamji:
I mean, that's a bigger, *unnecessary* change affecting all retry behavior, not just this case. Unlike the current solution, which targets specifically the cases where the interface disappears.

KiGamji:
@RPRX @Owersun what do you guys think?

Fangliding:
wrap一层err层层往上递尝试parse才是更糟糕的 判断err类型有很多坑 指不定哪里改一下返回值就失效了 加一个limiter只用几行代码就行了 而且可以处理其他情况

Owersun:
I'm not sure I have a word here, but my opinion is that's a decent suggestion. If config doesn't match environment - fail and exit is proper reaction, similar to some other "startup errors" that make the start ambiguous. Similar to what should happen when e.g. socket proxy is configured to listen, but at the time of start the port is occupied.

gfw-killer:
sockopt interface has other problems too https://github.com/XTLS/Xray-core/issues/5079
isn't it better to switch to sendThrough as @Fangliding suggests?
if it could get interface ip like "interface:eth0", it could replace sockopt interface
do the memory leak exist when using sendThrough too?

RPRX:
说到 sockopt，现在里面有能分享的有不能分享的（比如 interface），@Fangliding 我们应该把这两种区别开才好做后面的分享

KiGamji:
Well, sendThrough binds by IP, not the interface name. So if the interface goes down, its IP becomes unavailable too — same leak scenario.

The underlying issue is retrying on non-recoverable socket errors, regardless of whether you use interface or sendThrough.

Fangliding:
> 说到 sockopt，现在里面有能分享的有不能分享的（比如 interface），@Fangliding 我们应该把这两种区别开才好做后面的分享

我就觉得这玩意不该分享 本来应该是高级选项都 use at own risk
开出来破事太多（这个pr本身也是一例）

gfw-killer:
Isn't it better to add a 10 second delay after each failed retry? instead of a panic or just allowing the infinite retry with no delay
and of course an error log

KiGamji:
> > 说到 sockopt，现在里面有能分享的有不能分享的（比如 interface），@Fangliding 我们应该把这两种区别开才好做后面的分享
> 
> 我就觉得这玩意不该分享 本来应该是高级选项都 use at own risk 开出来破事太多（这个pr本身也是一例）

If you are fine with this, sure

image

RPRX:
@KiGamji 先写个 auto-route 吧 https://github.com/XTLS/Xray-core/issues/5594#issuecomment-3790257752 ，~~这个 PR 的话我想想~~

Issue #5631, opened 2026-01-31
XHTTP sessionPlacement unsupported path


Aichikaze:
### Integrity requirements

- [x] I have read all the comments in the issue template and ensured that this issue meet the requirements.
- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

According to https://github.com/XTLS/Xray-core/pull/5414#issuecomment-3770071786, session id can be placed in path, but when specifying "sessionPlacement": "path", Xray returns the error infra/conf: unsupported session placement: path

### Reproduction Method

Set sessionPlacement in the config to path

### Client config

{
  "log": {
    "loglevel": "debug"
  },
  "inbounds": [
    {
      "tag": "SOCKS",
      "port": 10808,
      "listen": "127.0.0.1",
      "protocol": "socks",
      "settings": {
        "udp": true
      },
      "sniffing": {
        "enabled": true,
        "destOverride": ["http", "tls"]
      }
    }
  ],
  "outbounds": [
    {
      "tag": "PROXY",
      "protocol": "vless",
      "settings": {
        "address": "ip",
        "port": 443,
        "id": "uuid",
        "encryption": "none"
      },
      "streamSettings": {
        "network": "xhttp",
        "security": "tls",
        "tlsSettings": {
          "alpn": ["h3"],
          "serverName": "domain"
        },
        "xhttpSettings": {
          "host": "host",
          "path": "/path",
          "mode": "packet-up",
          "extra": {
            "sessionPlacement": "path"
          }
        }
      }
    }
  ]
}

### Server config

{
  "log": {
    "loglevel": "debug"
  },
  "inbounds": [
    {
      "tag": "XHTTP",
      "listen": "127.0.0.1",
      "port": 7777
      "protocol": "vless",
      "settings": {
        "clients": [],
        "decryption": "none"
      },
      "sniffing": {
        "enabled": true,
        "destOverride": ["http", "tls"]
      },
      "streamSettings": {
        "network": "xhttp",
        "xhttpSettings": {
          "path": "/path",
          "sessionPlacement": "path"
        }
      }
    }
  ],
  "outbounds": [
    {
      "tag": "DIRECT",
      "protocol": "direct"
    }
  ]
}

[...]

PR #5634, opened 2026-02-01
Fix: XHTTP sessionPlacement unsupported path

paqx:
Fix for issue https://github.com/XTLS/Xray-core/issues/5631

# Cause

Config parser didn't expect that someone could explicitly set the `SessionPlacement` and `SeqPlacement` parameters to their default values (`path`).

# Fix

I added `path` to the list of allowed values for the `SessionPlacement` and `SeqPlacement` parameters.

### DIFF
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -334,7 +334,7 @@ func (c *SplitHTTPConfig) Build() (proto.Message, error) {
 	switch c.SessionPlacement {
 	case "":
 		c.SessionPlacement = "path"
-	case "cookie", "header", "query":
+	case "path", "cookie", "header", "query":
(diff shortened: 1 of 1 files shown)


RPRX:
新版本的 XHTTP+REALITY 应该是确实有问题，你测一下顺便修一下吧，~~今天再发一版~~

paqx:
> The new version of XHTTP+REALITY does seem to have a problem. You should test it and fix it while you're at it.Another version will be released today.

Is it about this https://github.com/XTLS/Xray-core/issues/5632? I will check it as well, but I cannot guarantee I will be able to reproduce it. The issue report is not very informative. I was hoping the author could share the configs and error messages.

PR #5702, opened 2026-02-17
Fix: resolve data races in splithttp transport

OneMiny:
## Description

Fixes three critical data races in the `splithttp` package identified through static analysis and `go test -race`. All three affect connection stability under concurrent workloads.

---

### 1. `WaitReadCloser` race — `client.go`

- **Problem:** `ReadCloser` was read/written across goroutines without synchronization, causing undefined read behavior and potential nil dereference.
- **Fix:** Guarded `ReadCloser` with `sync.Mutex`; used `sync.Once` for safe channel close.

### 2. Upload buffer use-after-handoff — `dialer.go`

- **Problem:** `buff.Len()` was called after buffer ownership was transferred to the pipe, racing with the consumer goroutine.
- **Fix:** Captured buffer length _before_ `WriteMultiBuffer` call.

### 3. `uploadQueue` unguarded fields — `upload_queue.go`

- **Problem:** `closed` and `reader` were read in `Read()` without lock, while `Close()` mutated them under lock.
- **Fix:** Extended `writeCloseMutex` scope to cover the `Read()` path.

## Impact

- Eliminates undefined behavior, potential panics, and flaky connection drops under load
- No API or config changes — drop-in safe

## Test

```bash
go test -race ./transport/internet/splithttp/...
```

### DIFF
--- a/transport/internet/splithttp/client.go
+++ b/transport/internet/splithttp/client.go
@@ -88,7 +88,7 @@ func (c *DefaultDialerClient) OpenStream(ctx context.Context, url string, sessio
 		req.Header.Set("Content-Type", "application/grpc")
 	}
 
-	wrc = &WaitReadCloser{Wait: make(chan struct{})}
+	wrc = &WaitReadCloser{wait: make(chan struct{})}
 	go func() {
 		resp, err := c.client.Do(req)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


Fangliding:
请至少找个真人来提交pr
此类提交需要复现步骤证实会导致使用问题

PR #5714, opened 2026-02-21
Config option allowing usage of forward proxy for selected vless "reverse" users.

zolg:
The current code deliberately prohibits outgoing traffic for vless users with the “reverse” enabled.

The reasons for this are clear, and as a default option, it is useful.

However, in some cases, it is undesirable.
In particular, when “user” is a router behind NAT and xray on the server serves both internet access for users of the router's network and forwarding ports to services on the router's internal network.

This simple patch adds the `allowForward` parameter to the `reverse` object in order to selectively allow usage of a forward proxy by selected users.

```
"inbounds": [
   {
      "protocol": "vless",
...
      "settings": {
         "clients": [
          {
            "id": "bidirectional user",
            "reverse": {
               "tag": "remote-gw",
               "allowForward": true
            }
         ]
      }
   }
]
```


RPRX:
同一个入站内支持多个 users

PR #5732, opened 2026-02-26
API: Fix Online Map

kastov:
After this functionality was introduced last year, there were some doubts about its usefulness. Unfortunately, after it began to be actively used, those concerns were confirmed. In its current state (without these fixes), Online Map is practically unusable and largely pointless.

The logic based on the last 20 seconds simply does not work. Due to the specific nature of how Xray operates, many connections live much longer than 20 seconds, and even if this time could be adjusted, it would not fix the underlying problem.

A fairly common situation was the following: no IP addresses are shown, yet traffic is actively flowing (being recorded in statistics), which indicates that a connection is active. However, it is impossible to determine which IP address it originates from — even though that is precisely what the Online Map feature is supposed to provide.

This PR fixes the incorrect Online Map logic: an IP address is now retained for as long as the connection remains active. Once the connection is closed, the IP address is removed. This is the correct behavior.

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -183,12 +183,9 @@ func (d *DefaultDispatcher) getLink(ctx context.Context) (*transport.Link, *tran
 		if p.Stats.UserOnline {
 			name := "user>>>" + user.Email + ">>>online"
 			if om, _ := stats.GetOrRegisterOnlineMap(d.stats, name); om != nil {
-				sessionInbounds := session.InboundFromContext(ctx)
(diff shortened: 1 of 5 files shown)


kastov:
~also, there were issues in the current implementation: several data races in the internal map, last cleanup occurring outside the mutex, and a lock that was blocking all Manager operations.~

Fangliding:
@hossinasaadi

hossinasaadi (approved):
Looks good to me. This is cleaner and more effective, thanks for the improvement.

PR #5737, opened 2026-02-27
XTLS Vision: Defer Splice handoff until write completes

HeXis-YS:
## Summary

This PR addresses an intermittent Vision handoff race that can lead to TLS stream breakage (`ERR_SSL_PROTOCOL_ERROR`), as reported in #4878.

## Root Cause

In `VisionWriter.WriteMultiBuffer`, `inbound.CanSpliceCopy` could become `1` before the current direct-write buffer had fully flushed.  
That allowed splice mode to start while Vision was still writing to the same socket, creating a race at the handoff boundary.

## Changes

- Updated `VisionWriter.WriteMultiBuffer` in `proxy/proxy.go`:
  - keep splice readiness in a local variable during direct-write switch;
  - perform `w.Writer.WriteMultiBuffer(mb)` first;
  - only then set `inbound.CanSpliceCopy = 1` (if still eligible).

This keeps existing Vision behavior intact while making splice handoff ordering deterministic.


RPRX:
@yuhan6665

yuhan6665:
@HeXis-YS thanks for your work.
I understand you want to make sure the buffer is written first and cleared, and then send the signal by setting the CanSpliceCopy flag. Although I believe there are more than one thread (reader) running and more than one buffer involved. So your proposal maybe better in some cases, but maybe not in others.
My test is this - see if we can get rid of the time.Sleep() in CopyRawConnIfExist() completely, and only rely on code signal to switch to splice copy - I feel like we may need another signal. If we can do that we probably really solve the root cause.

HeXis-YS:
@yuhan6665 
I have tested it on my server and it works fine. However, my test scenarios are very limited.
Please feel free to test. Any ideas or suggestions are welcome.

yuhan6665:
~~Can you try remove time.Sleep() in CopyRawConnIfExist() and test it again?~~ nevermind I just see you did it

yuhan6665:
@HeXis-YS sorry but I can't understand what you added for the ready channel. I don't think it does anything more since there is no other new code signal than the existing place to call EnableSpliceCopy()
If you agree, can you remove the three new commits? And keep the removal of time.sleep() in CopyRawConnIfExist()

Maybe we can do a beta release and see if the issue is fixed.

Fangliding:
话说vless入站不是早就改成dispatchlink不走pipe了吗 就一个gouroutine在搬东西 还有竞争?

HeXis-YS:
Done.
These 3 commits were actually intended to fix the problem raised by Copilot, but it seems Copilot wasn't that smart.

yuhan6665:
> 话说vless入站不是早就改成dispatchlink不走pipe了吗 就一个gouroutine在搬东西 还有竞争?

Seems you are right, let's give it a try

PR #5746, opened 2026-03-01
Merge hosts and servers of the built-in DNS server from config files,…

26X23:
This PR merges hosts and servers of the built-in DNS from two configuration files and overrides other DNS fields from the latest configuration file.

I have one configuration file per every outbound and one common configuration file with inbound, routing and dns. When using tun inbound I need to route a DNS request for the outbound's address to the domestic DNS via freedom protocol with a special sockopt/mark. Since outbound's address is different in every configuration file I want to write one DNS server with the address of one outbound in "domains" in the per outbound configuration file. Without this PR I have to enumerate addresses from all outbounds in the "domains" of the DNS server in the common configuration file.

### DIFF
--- a/infra/conf/dns.go
+++ b/infra/conf/dns.go
@@ -2,6 +2,7 @@ package conf
 
 import (
 	"bufio"
+	"cmp"
 	"encoding/json"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Fangliding:
这个问题预计是 #5594 的一部分 合并dns server可能导致某些不可知的行为（鉴于这个模块非常乱）就像我们不合并路由规则一样

PR #5770, opened 2026-03-07
Log: Enhance error handling by directing error messages to stderr

ImMohammad20000:
## Summary
This PR fixes console log stream routing so error logs go to `stderr` instead of `stdout`.

## Changes
- Added `IsError` to `HandlerCreatorOptions` to distinguish error vs non-error logger creation.
- Updated logger initialization:
  - Access logger uses `IsError: false`
  - Error logger uses `IsError: true`
- Updated console handler creator to:
  - use `CreateStderrLogWriter()` when `IsError` is true
  - keep `CreateStdoutLogWriter()` otherwise
- Updated CLI startup/config failure prints in `main/run.go` to `stderr` via `fmt.Fprintln(os.Stderr, ...)`.

## Why
Previously, both access and error logs were written to `stdout`, which is incorrect for standard stream semantics and makes operational log handling (systemd, Docker, shell redirection, log collectors) harder.

### DIFF
--- a/app/log/log.go
+++ b/app/log/log.go
@@ -53,7 +53,8 @@ func New(ctx context.Context, config *Config) (*Instance, error) {
 
 func (g *Instance) initAccessLogger() error {
 	handler, err := createHandler(g.config.AccessLogType, HandlerCreatorOptions{
-		Path: g.config.AccessLogPath,
+		Path:    g.config.AccessLogPath,
+		IsError: false,
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


Fangliding:
error loger 不全是 error 老问题了 倾向于不break

M03ED:
same as every pr/issue @Fangliding close it and @RPRX should reopen it.

ImMohammad20000:
If there is any problem with error logs send to stderr we can just send logs in run file to stderr

PR #5773, opened 2026-03-07
mKCP transport: Make sure ACKs are limited within MTU

LjhAUMEM:


patterniha:
@RPRX 
In the Iranian group, many people requesting for official release to use XDNS, the main reason is because of client-side, all android/ios apps are updated based on the official release version.

~i didn't see https://t.me/projectXtls/1754~

RPRX:
@patterniha 新版本将很快到来，我知道 XDNS 相对于 DNSTT 的优势是支持 UDP，但为什么有人说 XDNS 能通过更多 DNS？DNSTT 和 SlipStream 不能限制成更小的 MTU 吗？另外伊朗的全国性断网增强了，仍然能查询国际上的 TXT 记录？

patterniha:
@RPRX 

https://github.com/Mygod/slipstream-rust/pull/5#issuecomment-3751304180

https://github.com/Mygod/slipstream-rust/pull/5#issuecomment-3766162518

TXT record is still available, also  i don't think it can be easily blocked in all resolvers, unless they block at the very beginning (dpi), which can be bypassed with noise (or small ip-header-ttl or wrong udp-header checksum)

I also write two way query data transfer which works with A record too: https://github.com/patterniha/QQ-Tunnel

///

I tested and XDNS works without problem, but  speed is not good, so:

1- we should use different source port to send each query
2- we should be able to use multiple resolvers To distribute queries between them and bypass rate limit.

Issue #5800, opened 2026-03-13
Commit b7a22c72 causes HTTPS reverse proxies to error out


PoneyClairDeLune:
### Integrity requirements

- [x] I have read all the comments in the issue template and ensured that this issue meet the requirements.
- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

The commit [`b7a22c72`](https://github.com/XTLS/Xray-core/commit/b7a22c729be98e613d90ab8e1140b72ea009696b) causes transports based on HTTP (e.g. WebSocket and gRPC) to become unable to pass through reverse proxies.

### Reproduction Method

1. Connect to any server behind a reverse proxy with version 26.2.6, preferrably with the WebSocket transport.

### Client config

<details>
```json
{
	"log": {
		"loglevel": "info",
		"dnsLog": true,
		"maskAddress": "/24+/48"
	},
	"dns": {
		"tag": "d-dns",
		"servers": [
			"8.8.8.8",
			"1.1.1.1",
			{
				"address": "8.8.4.4",
				"skipFallback": true,
				"domains": ["__SERVER__"]
			},
			{
				"address": "1.0.0.1",
				"skipFallback": true,
				"domains": ["__SERVER__"]
			}
		]
	},
	"inbounds": [{
		"tag": "i-forwarder",
		"protocol": "tunnel",
		"listen": "127.0.0.1",
		"port": 23456,
		"settings": {
			"address": "11.15.0.1",
			"port": 12345,
			"network": "tcp"
		}
	}],
	"outbounds": [{
		"tag": "o-block",
		"protocol": "blackhole",
		"settings": {
			"response": {
				"type": "none"
			}
		}
	}, {
		"tag": "o-direct",
		"protocol": "freedom",
		"settings": {
			"domainStrategy": "AsIs"
		}
	}, {
		"tag": "o-remote",
		"protocol": "vless",
		"settings": {
			"address": "__SERVER__",
			"port": 443,
			"flow": "",
			"id": "__UUID__",
			"encryption": "none"
		},
		"streamSettings": {
			"network": "ws",
			"security": "tls",
			"wsSettings": {
				"path": "__PATH__",
				"host": "__SERVER__"
			},
[...]

PR #5802, opened 2026-03-14
Xray-core: More robust browser header masquerading (chrome, firefox, edge)

PoneyClairDeLune:
As demonstrated in #5800, as soon as there is any kind of WAF rule trying to fingerprint headers from the client, merely having UA strings match will instead trigger them. This PR aims at fixing that.

- The request initiators no longer just try to use user agent strings, but supply the missing headers from the real browsers adhering to the conventions as well.
- The `User-Agent` header field now contains special values, namely `chrome`, `firefox`, `edge`, `golang` and an empty string.
  - Leaving the header unset or having it set to `chrome` will apply headers from Google Chrome.
  - `firefox` to set the headers to the current Firefox ESR release.
  - `edge` to set the headers to Microsoft Edge.
  - `golang` to retain the default headers from Go `net/http`.
  - An empty string to disable the UA header altogether.

This PR does not, however, fix the potential issue in Chrome version generation.

Original content:
<details>
As demonstrated in #5800, as soon as there is any kind of WAF rule trying to fingerprint headers from the client, merely having UA strings match will instead trigger them. This PR aims at fixing that by bringing in the missing headers following the browser conventions.

However, this PR is **by no means complete** for the following reasons:
- The JS runtime within the browser is *fundamentally incapable of* initiating real gRPC requests, as such having browsers initiating compliant gRPC requests is *impossible to encounter in the real world*.
- Having the headers emulate Chrome ones, even when the uTLS fingerprint is set to emulate browsers other than Chrome, is ridiculous. If header masquerading is to be done correctly, an empty UA should cause the default headers to follow the one set in uTLS. Or even better, make it configurable when set the `User-Agent` string to following values.
  - An empty string or `!utls` causes the default headers to follow the intended browser set in uTLS.
  - `!chrome`, `!firefox`, `!edge`, `!safari`, etc will set the default headers accordingly.

It's merely a suggestion, but please talk to someone having experience in in-browser fingerprinting before pushing changes like this.

Edit: There's also [this additional comment](https://github.com/XTLS/Xray-core/pull/5802#issuecomment-4062803858) regarding potential pitfalls on Chromium version generation. It's also equally important for GUI clients and sharable links to include a field specifying user agent strings, as this is going to be applied by default, likely causing disruptions.
</details>

### DIFF
--- a/common/utils/browser.go
+++ b/common/utils/browser.go
@@ -106,7 +106,7 @@ func applyMasqueradedHeaders(header http.Header, browser string, variant string)
 		header.Set("User-Agent", FirefoxUA)
 		header["DNT"] = []string{"1"}
 		header.Set("Accept-Language", "en-US,en;q=0.5")
-	case "go":
+	case "golang":
 		// Expose the default net/http header.
 		header.Del("User-Agent")
 		return
@@ -185,7 +185,7 @@ func TryDefaultHeadersWith(header http.Header, variant string) {
 		case "edge":
 			applyMasqueradedHeaders(header, "edge", variant)
 		case "golang":
-			applyMasqueradedHeaders(header, "go", variant)
+			applyMasqueradedHeaders(header, "golang", variant)
 		}
 	}
 }


PoneyClairDeLune:
Added default headers for both Chrome and Firefox. I don't have either Edge or Safari for their headers, but I'd be quite delighted if anyone's capable of adding them.

PoneyClairDeLune:
I've gone forward and fully implemented the brand GREASE algorithm from the Chromium project. Header masquerading should now be pretty close to a real Chromium-based browser.

PoneyClairDeLune:
~~Yeah the commit history alone looks like poetry.~~

PoneyClairDeLune:
This PR should be finished fully within its intended scope, nonetheless I still urge the maintainers to actively ponder on the points raised within this PR.

I swear I had some more points to add as I'm writing this, yet my fish brain seemed to have caught up. I guess apart from asking for people willing to test out how well the browser dialer integrates within `Xray-core`, I'll open a separate issue or pull request whenever I remember it.

PoneyClairDeLune:
The fish brain was defeated.

The result of stacking random values on top of each other, mathematically infinitely approaching a standard normal distribution.

The current algorithm for generating Chrome versions is basically a somewhat simplified version of running the duration against the cumulative distribution function of a scaled normal distribution function. While it is viable in short term, it is likely that variance will be far too great once time catches up. It might be more valid to stack the final few random values against the duration multiplied by the 35-day expectation to decrease the snowballing variation.

useruserdev:
Yo, great find. Can confirm — `chrome` fingerprint gets dropped every time on suspect IPs (CF ranges etc.), while `android` (OkHttp) passes through no problem. Makes total sense now, DPI catches the mismatch between uTLS chrome ClientHello and the bare-bones Go HTTP headers. OkHttp gets away with it because its headers are already minimal, so there's nothing to mismatch.

Literally been stuck on `android` as the only working fingerprint for months lol. Would be huge to finally have proper header masquerading so `chrome`/`firefox` actually work as intended. Looking forward to this getting merged 🤝

PoneyClairDeLune:
@useruserdev Can you test on Windows, Linux or in Termux to confirm if headers alone solve your problem?

useruserdev:
Can test on WSL Ubuntu. Currently `android` is the only fingerprint that passes on CF-proxied IPs - `chrome`, `safari`, `ios` all get dropped. Real browsers (Chrome, Safari from actual devices) pass fine on the same IPs though, so it's clearly the header mismatch getting caught. Will build from your branch and report back.

Edit: Shold I just build from pr-chrome-headers branch as-is, or is there anything specific you want me to configure?

PoneyClairDeLune:
@useruserdev You can download the executable artifacts from the automated job runs to test it out, like [`linux-64.zip`](https://github.com/XTLS/Xray-core/actions/runs/23129910831/artifacts/5937848036).

RPRX:
~~想过这些不过挺麻烦的但似乎已经被你完成了~~，感谢贡献！对照 https://github.com/XTLS/Xray-core/pull/5658 发现只有 gRPC 没改~~不过无所谓~~，可以合并了吗？

另外 Golang h2 的行为也与 Chromium 不完全一致，有空可以研究下

Fangliding:
还有一个问题是 header.Get("User-Agent") == "" 这个不能区分是没设置还是用户手动置空 要用map里判断元素是否存在那个方法判断 我前段时间就想说了

RPRX:
~~原来还支持用户手动置空这么边缘的操作吗，话说 Golang 加默认 UA 前有没有这个判断~~

PoneyClairDeLune:
@RPRX The only reason I didn't change the header in gRPC was that... ~~I wasn't sure if the change was gonna apply with simple removal XD.~~
But yes, apart from gRPC, most others should be ready.

Fangliding:
> ~原来还支持用户手动置空这么边缘的操作吗，话说 Golang 加默认 UA 前有没有这个判断~

没填=填默认值(默认ua)
明确置空=空(不发)
这是标准库行为
xray里不少地方也这样

PoneyClairDeLune:
Correction: It wouldn't be quite ready due to header capitalization formatting applied by `net/http` by default. I'll try with some fixes.

Fangliding:
这个叫 Canonical Header

RPRX:
> 明确置空=空(不发)

~~考虑到有些 GUI 会给用户没填的 json 字段填个 ""，还是先保持现在的行为吧，等真有人需要把它置空的话再改~~

PoneyClairDeLune:
I commented out the lines distinguishing between unset and blank values, so that the behaviour can be easily switched. Also added the `!chrome` (the default value is currently an alias to this), `!firefox`, `!go` (so that the default UA from `net/http` can be still chosen when needed in some edge cases) as special values for coherence with other uTLS fingerprints. Note that even though I tried to bypass the canonical header serialization in `net/http`, it's still present in the resulting connections, so Xray might be forced to soft-fork `net/http` to prevent this from happening.

Reviews are much appreciated. It would be even better if there are other ~~guinea pigs~~ volunteers to help give header values from other safer browsers available under uTLS!

> Considering that some GUIs will fill empty string values for json fields that are not filled in by the user, it is better to keep the current behavior and change it later if someone really needs to leave it blank.

@RPRX Is it possible to also tell GUI clients to also provide a field allowing users to customize user agent strings, as well as incorporating it in the sharable link standard? The current lack of it is just plain infuriating.

useruserdev:
image

first one android utls
second chrome from your pr

useruserdev:
Just to clarify after testing, this PR actually fixes a different layer than I'm dealing with. Your PR actually fixes an HTTP header mismatch issue, which is what I'm experiencing after the handshake. This is actually what triggers the detection of reverse proxy or WAFs, as seen in issue #5800. The issue I'm experiencing on our DPI is actually at the TLS ClientHello level. The uTLS chrome, safari  fingerprints themselves aren't specific enough for reverse proxy DPI inspection, regardless of what HTTP headers are sent after.

So, I'm experiencing two different things here:
HTTP Headers mismatch. Your PR actually fixes this issue
uTLS ClientHello does not match actual browsers

Android actually passes both layers because OkHttp is actually simple enough for uTLS to fingerprint accurately, and default HTTP headers from Go don't actually cause a mismatch

Fangliding:
Android那个指纹是个糟糕的tls1.2 可能只是它和tls1.3使用了不同的机制。。

useruserdev:
You were right. Just tested with plain curl (no uTLS at all):

- curl --tlsv1.3 -> hangs after ClientHello, DPI drops it
- curl --tlsv1.2 --tls-max 1.2 -> passes, full handshake completes

Same IP, same SNI. So DPI is specifically targeting TLS 1.3 on suspect IPs and letting TLS 1.2 through

useruserdev:
but real Chrome/Safari also do TLS 1.3 and they pass fine on the same IP, so it's not just blocking all 1.3 right? any idea what the DPI is actually catching then?

PoneyClairDeLune:
Speaking of, is there any reviewer available?

Fangliding:
image
 
这些被反复复制的部分也可以被封装进函数 而且上面本来之前弄的允许置空的逻辑都注释了?

PoneyClairDeLune:
@Fangliding I commented it out per observation of RPRX for easy behaviour switch. Guess it's time to get the special values in a method too.

Fangliding:
> > 明确置空=空(不发)
> 
> ~考虑到有些 GUI 会给用户没填的 json 字段填个 ""，还是先保持现在的行为吧，等真有人需要把它置空的话再改~

据我所知没有gui给客户填这个 这是个明确的切片性质的字段 `[]`/`null` 和 `["user-agent":""]` 显然不同 不能老是考虑迁就这个点就把行为做这么糟糕

PoneyClairDeLune:
Moved the special value handler to a common global method, so the following behaviour applies:
- `"!chrome"` or unset: Applies Chromium headers.
- `"!firefox"`: Applies Firefox headers.
- `"!go"`: Applies the default Go `net/http` headers.
- `""`: No UA headers are applied at all.

I don't have the sample to do `"!edge"`, `"!android"`, `"!safari"` or `"!ios"` right now, but it would be nice if there's anyone willing to add them. ~~Or even better, send me the samples and let me add them for the extra street credit XD~~

RPRX:
~~有个顾虑是如果装得太像，CDN 可能会弹人机验证页面，可能只影响 nav 吧~~

UA 那个，gRPC 目前还是单独字段设置 UA，~~WS 似乎以前也支持但被我删了~~，其它的改成允许设置空 UA 吧

common/utils/browser.go
PoneyClairDeLune: I meant "cadence", not a certain heartbutt.

PoneyClairDeLune:
Apart from gRPC, user agent fields in other web-based transports already had null values allowed yesterday. I did add special values in the UA field in gRPC, but it's entirely not recommended.

The question currently is, would GUI clients like v2rayN(G), OneXray etc actually allow setting the UA field without writing raw config files? And the lack of such field in the sharable link standard would also be encouraging the incoherence problem, especially when the uTLS fingerprint is not set to `chrome`.

Edit: Oh, and I forgot to mention, but I just added `!edge` to the arsenal.

PoneyClairDeLune:
~~So... Is this pull request good enough to get merged?~~

RPRX:
~~本来问题不大的，改出问题了~~

1. 全局默认 Chrome UA 本来就是因 gRPC 的问题而提上日程，被你改没了，还是改回来吧，有问题的话再说
2. 目前的代码看起来没有区分用户是否指定 UA 为空（至少 XHTTP 支持分享这个），gRPC 除外
3. 和 fingerprint 的配置方式对齐一下，比如 chrome 就填 chrome，不需要 `!`，go 改为 golang，~~虽然指纹配置名是 unsafe~~

PoneyClairDeLune:
> The current code doesn't seem to differentiate between whether the user specifies an empty User Agent (at least XHTTP supports sharing this), except for gRPC.

It does in an implicit way. Current code detects if the header hasn't been set so that Chrome can be used as default, and empty strings used to prevent UA from being set entirely will be set one step before the special value handling, so nothing is needed for that specific use case. No idea how to deal with null UA for gRPC though.

Hopefully the exclamation mark removal wouldn't get Go to scream at me again.

Fangliding:
> ~本来问题不大的，改出问题了~
> 
> 1. 全局默认 Chrome UA 本来就是因 gRPC 的问题而提上日程，被你改没了，还是改回来吧，有问题的话再说
> 2. 目前的代码看起来没有区分用户是否指定 UA 为空（至少 XHTTP 支持分享这个），gRPC 除外
> 3. 和 fingerprint 的配置方式对齐一下，比如 chrome 就填 chrome，不需要 `!`，go 改为 golang，~虽然指纹配置名是 unsafe~

主要是从真实性的角度来看chrome不可能去发出一个grpc请求 这样的流量绝对是诡异的 只是从面向CF编程的角度这样可以解决问题

Fangliding:
UseDefaultHeadersWith和ApplyDefaultHeaders应该没必要拆开 Use 就是判断一下用户是否输入UA是否存在以及根据里面的预设值选择header组合调用ApplyDefaultHeaders 对用户没法自定义的外部请求UA本来就是空的就是走的第一条自动进chrome分支 都可以用一个UseDefaultHeadersWith

RPRX:
~~毕竟 gRPC 传输层从一开始就是面向 cf 编程，而为了讨好 GFW 还得弄 Chrome CH 指纹，所以~~

`if len(header.Values("User-Agent")) < 1 {` 这个判断方式还是有问题，用户指定为空的话也会命中 @PoneyClairDeLune

Fangliding:
空的话不应该 len 出来是1进下面的switch最后哪个case都没中就正常返回了吗（不动header）

RPRX:
~~原来这个是取成员数而不是取值吗，那没事了~~

按照 @Fangliding 说的合并一下函数吧

PoneyClairDeLune:
> There is still a problem with this criterion. If the user specifies an empty string, it will still be satisfied.

@RPRX Empty strings will not satisfy that criterion, as that will make the total length of the value slice to be 1, with the empty string as the sole member of that slice.

> There should be no need to separate UseDefaultHeadersWith and ApplyDefaultHeaders.

@Fangliding I split the two methods because it's not just the transports that are applying the headers, places like the DoH resolver and the observatory are also using that. `UseDefaultHeadersWith` hence is solely used to handle the transports, while `ApplyDefaultHeaders` is used everywhere else. Not so sure about merging the two methods together is a good idea.

Fangliding:
DOH 的 len(header.Values("User-Agent")) 肯定是0的 又没有人去set它 我只是觉得保留两个含义相近命名的导出函数容易造成混淆

RPRX:
~~确实是，我 review 的时候也觉得有点混乱，还是合并下吧~~

PoneyClairDeLune:
Every single instance trying to apply the masqueraded headers now go through `utils.HandleTransportUASettings()`, and `applyMasqueradedHeaders()` is now internal to get around spaghetti code. Guess this would be good enough?

RPRX:
@PoneyClairDeLune 看一下 review，然后没别的要改的话我就合了

PoneyClairDeLune:
@RPRX Should be all good now once the checks are passed.

RPRX:
合并中，@PoneyClairDeLune 更新一下 PR 正文描述

PoneyClairDeLune:
The description of the PR was updated.

PR #5804, opened 2026-03-14
Improve proxy switching reliability

drovosek229:
## Summary

This PR improves `leastload` proxy switching smoothness when used with the burst observatory, while keeping existing behavior backward compatible by default.

The main goals are:
- reduce unnecessary switching/churn
- avoid promoting a proxy too early on a lucky first sample
- keep failover fast on real failures
- improve burst health snapshot quality without changing the public observatory schema

### `leastload`
- add optional `minSamples` setting
  - when `> 0`, burst candidates with too few samples cannot displace the current healthy winner
  - disabled by default (`0`)
- add optional `softFailGrace` setting
  - when `> 0`, the current selected outbound may survive a limited number of consecutive soft-fail observation cycles
  - applies only when the outbound is still `Alive=true` but fails `maxRTT` or `tolerance`
  - hard failures still switch immediately
  - disabled by default (`0`)
- keep existing sticky behavior when the current winner remains qualified
- keep existing cold-start fallback behavior
- tighten ranking for burst data
  - still sorts primarily by deviation-cost, then average RTT
  - now prefers lower fail ratio before raw fail count
  - then prefers higher sample count

### Burst observatory
- spread scheduled probes across per-sample slots inside the batch window instead of distributing all probes uniformly across the full window
- publish meaningful `last_try_time` and `last_seen_time` in burst observation snapshots
- keep deterministic immutable snapshots

### Config / compatibility
- extend `StrategyLeastLoadConfig` with:
  - `minSamples`
  - `softFailGrace`
- add matching JSON parsing in router config
- keep defaults fully backward compatible
- no new top-level config blocks
- no changes to legacy observatory behavior
- no changes to `random`, `roundrobin`, or `leastping`

## Testing

Passed:
- `go test ./app/router`
- `go test ./infra/conf`
- `go test ./app/observatory/burst`

Additional notes:
- local `geoip.dat` / `geosite.dat` assets were downloaded only for test execution and are not part of this PR

## Example

```json
"strategy": {
  "type": "leastload",
  "settings": {
    "expected": 1,
    "maxRTT": "1800ms",
    "tolerance": 0.25,
    "minSamples": 2,
    "softFailGrace": 1
  }
}
```


Fangliding:
这个模块除了bug修复外暂时不想加新功能了（现在的字段已经让人难以理解了） 而且有点小问题(

PR #5805, opened 2026-03-14
Allow sockopt.dialerProxy to resolve balancer tags

drovosek229:
Title: Allow sockopt.dialerProxy to resolve balancer tags

## Summary

This change makes `sockopt.dialerProxy` accept a balancer tag in addition to a direct outbound tag.

When `dialerProxy` is set, Xray now:

- Tries to resolve it as a direct outbound tag first
- Falls back to resolving it as a routing balancer tag
- Uses the selected outbound handler for the redirect path

This keeps existing direct-outbound behavior unchanged while allowing `sockopt` to follow the same balancer selection flow as a normal outbound proxy reference.

## Changes

- Added a small routing-side balancer selector interface
- Implemented balancer tag resolution in the router
- Wired the selector into `transport/internet` system dialer initialization
- Updated `dialerProxy` resolution to use the selected outbound tag before redirecting
- Added tests for:
  - successful balancer-to-outbound resolution
  - balancer error propagation

## Testing

- `go test ./core`
- `go test ./transport/internet`
- `go test ./app/router -run '^$'`

### DIFF
--- a/app/router/balancing.go
+++ b/app/router/balancing.go
@@ -134,6 +134,15 @@ func (b *Balancer) SelectOutbounds() ([]string, error) {
 	return tags, nil
 }
 
+// PickBalancerOutbound implements routing.BalancerSelector.
+func (r *Router) PickBalancerOutbound(tag string) (string, bool, error) {
+	if b, ok := r.balancers[tag]; ok {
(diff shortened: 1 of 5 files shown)


Fangliding:
Stop sending AI pr

Issue #5868, opened 2026-03-30
When using Hysteria inbound, connected users are not tracked in online stats


nwahs1337:
### 完整性要求

- [x] 我读完了 issue 模板中的所有注释，确保填写符合要求。
- [x] 我保证阅读了文档，了解所有我编写的配置文件项的含义，而不是大量堆砌看似有用的选项或默认值。
- [x] 我提供了完整的配置文件和日志，而不是出于自己的判断只给出截取的部分。
- [x] 我搜索了 issues, 没有发现已提出的类似问题。
- [x] 问题在 Release 最新的版本上可以成功复现

### 描述

<pre><code>
When using Hysteria inbound, connected users are not tracked in online stats (`user>>>...>>>online`) if inbound traffic counters are enabled.

Observed behavior:
- User is actually connected via Hysteria and passes traffic.
- Online/IP queries return empty for that user.
- The same environment works for VLESS/Trojan.

Impact:
- `GetStatsOnline` / `GetStatsOnlineIpList` for Hysteria users can return not found/empty.
- Downstream systems (e.g. panels) cannot detect Hysteria online users correctly.

Code-path analysis points to loss of `User()` method due to stats wrapper:

1) Incoming connection is wrapped by `CounterConnection` when inbound stats are enabled:
https://github.com/XTLS/Xray-core/blob/d2758a023cd7f4174a5a5fa4ff66e487d4342ba0/app/proxyman/inbound/worker.go#L105-L110

2) `CounterConnection` does not proxy `User()`:
https://github.com/XTLS/Xray-core/blob/d2758a023cd7f4174a5a5fa4ff66e487d4342ba0/transport/internet/stat/connection.go#L13-L44

3) Hysteria server extracts user from wrapped `conn` via type assertion:
https://github.com/XTLS/Xray-core/blob/d2758a023cd7f4174a5a5fa4ff66e487d4342ba0/proxy/hysteria/server.go#L88-L93

4) Dispatcher adds `user>>>...>>>online` only when `session.Inbound.User != nil`:
https://github.com/XTLS/Xray-core/blob/d2758a023cd7f4174a5a5fa4ff66e487d4342ba0/app/dispatcher/default.go#L205-L229

So, for Hysteria with inbound counters enabled, `inbound.User` may stay nil and online map is never populated.

### 重现方式

1. Enable system inbound stats in policy:
   - `statsInboundUplink: true`
   - `statsInboundDownlink: true`
[...]

PR #5870, opened 2026-03-31
Hysteria inbound: Unwrap stats conn before extracting user

hexband:
proxy/hysteria/server.go extracts the user from conn before unwrapping stats wrappers.

If the connection is wrapped, conn.(User) may fail even though the underlying Hysteria connection implements User() *protocol.MemoryUser.

As a result, inbound.User may remain nil and AccessMessage.Email stays empty.

Fix

Unwrap the stats connection first and then extract the user from the underlying connection.

Related to #5868


LjhAUMEM:
Nice fix; I forgot to enable stats:{} when testing the router.

PR #5871, opened 2026-03-31
feat(xdns): add resolver-based DNS tunneling with multi-resolver fan-out; fix sendLoop starvation

nnemirovsky:
## Summary

Two related changes to XDNS finalmask:

### 1. Resolver-based DNS tunneling (new feature)

Add optional `resolvers` config field. When set, the client sends DNS queries through public resolvers (1.1.1.1, 8.8.8.8, etc.) instead of connecting directly to the server on port 53.

- Each resolver gets its own UDP socket and receive goroutine for parallel I/O
- Round-robin distribution across resolvers
- Fully backward compatible: omitting `resolvers` preserves current direct-connection behavior
- Server needs no changes (already acts as authoritative DNS, uses embedded clientID for demux)

Config example:
```json
"finalmask": {
  "udp": [{
    "type": "xdns",
    "settings": {
      "domain": "t.example.com",
      "resolvers": ["1.1.1.1", "8.8.8.8", "9.9.9.9"]
    }
  }]
}
```

**Why:** Direct connection to server:53 is easily blocked by IP. Routing through public DNS resolvers makes traffic look like normal DNS queries and enables multi-resolver multiplexing for better throughput.

### 2. Fix sendLoop starvation under mKCP load (bug fix)

The XDNS server sendLoop failed to deliver downstream data when mKCP generated retransmission floods. HTTPS connections through the tunnel would time out because:

- The `nextRec = <-c.ch` case in the inner select caused the sendLoop to abandon responses whenever a new query arrived, resulting in empty responses under load
- The 1-second `maxResponseDelay` was too slow, causing the sendLoop to fall behind
- The write queue (capacity 512) overflowed and silently dropped data
- The mKCP 30s connection timeout was too short for DNS tunnel latency

Changes:
- Remove query preemption from sendLoop inner select
- Drain excess records before processing to skip stale queries
- Reduce response wait from 1s to 50ms
- Increase server write queue from 512 to 4096
- Increase mKCP connection timeout from 30s to 120s

## Test plan

- Unit tests for resolver address parsing (`TestParseResolverAddr`)
- Integration test with mock resolver (`TestResolverModeRoundTrip`)
- End-to-end round-trip test (`TestDirectModeRoundTrip`)
- Verified HTTPS works through XDNS tunnel on localhost (both direct and resolver mode)

### DIFF
--- a/transport/internet/finalmask/xdns/server.go
+++ b/transport/internet/finalmask/xdns/server.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/xtls/xray-core/common/errors"
@@ -15,9 +16,8 @@ import (
 )
 
 const (
-	idleTimeout      = 10 * time.Second
-	responseTTL      = 60
-	maxResponseDelay = 1 * time.Second
+	idleTimeout = 10 * time.Second
+	responseTTL = 60
 )
 
 var (
@@ -58,7 +58,7 @@ type xdnsConnServer struct {
 	readQueue     chan *packet
 	writeQueueMap map[string]*queue
 
-	closed bool
+	closed atomic.Bool
 	mutex  sync.Mutex
 }
 
@@ -90,7 +90,7 @@ func (c *xdnsConnServer) clean() {
 		c.mutex.Lock()
 		defer c.mutex.Unlock()
 
-		if c.closed {
+		if c.closed.Load() {
 			return true
 		}
 
@@ -116,14 +116,14 @@ func (c *xdnsConnServer) clean() {
 }
 
 func (c *xdnsConnServer) ensureQueue(addr net.Addr) *queue {
-	if c.closed {
+	if c.closed.Load() {
 		return nil
 	}
 
 	q, ok := c.writeQueueMap[addr.String()]
 	if !ok {
 		q = &queue{
-			queue: make(chan []byte, 512),
+			queue: make(chan []byte, 4096),
 			stash: make(chan []byte, 1),
 		}
 		c.writeQueueMap[addr.String()] = q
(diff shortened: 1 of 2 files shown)


RPRX:
这个不是本来就是在 outbound address 填 1.1.1.1 这类 public DNS 地址吗？

@LjhAUMEM

nnemirovsky:
@RPRX, you're right that setting a public DNS address like `1.1.1.1` in the outbound address would route queries through a resolver. The DNS delegation (NS record) handles forwarding to the authoritative server automatically.

The `resolvers` field adds **multi-resolver fan-out within a single mKCP session**. Each resolver gets its own UDP socket and receive goroutine. Queries are distributed round-robin across all configured resolvers simultaneously. This increases throughput because each resolver is an independent bandwidth channel.

This can't be achieved with multiple `vnext` entries because Xray picks one address per connection. A single mKCP session would use only one resolver for its entire lifetime.

The sendLoop fix in the second commit is independent and addresses a real bug where downstream data delivery stalls under mKCP retransmission load.

RPRX:
大概懂了你是想实现 https://github.com/XTLS/Xray-core/issues/4846#issuecomment-3041680786 这种多路径加速，这对于 XDNS 来说确实挺有用，~~而且还是 mKCP，这下成 mekya 了~~

有没有兴趣顺便实现 https://github.com/XTLS/Xray-core/issues/4846#issuecomment-4150329444

Fangliding:
image
纯AI迷惑改动。。

RPRX:
我发现这些 AI PR 虽然无法被直接合并但有时候想法是有可取之处的，@LjhAUMEM 你看下吧

nnemirovsky:
@RPRX @Fangliding @LjhAUMEM, check this out please #5872. I cleaned up some AI artifacts and fixed what @RPRX asked for

PR #5902, opened 2026-04-10
mux: Fix OOM crash on UDP keep frames

### LINKED ISSUE #5901: mux: server crashes with OOM panic on UDP keep frames — race on b.UDP + signed-int bypass in isDomainTooLong
## What happens

On a busy server handling a lot of UDP-over-mux traffic (mostly DNS, some QUIC) the process occasionally aborts with:

```
runtime: out of memory: cannot allocate 11596739538348670976-byte block (140247040 in use)
fatal error: out of memory

runtime.stringtoslicebyte
github.com/xtls/xray-core/common/protocol.(*addressParser).writeAddress
    common/protocol/address.go:251
github.com/xtls/xray-core/common/protocol.portFirstAddressParser.WriteAddressPort
    common/protocol/address.go:110
github.com/xtls/xray-core/common/mux.FrameMetadata.WriteTo
    common/mux/frame.go:105
github.com/xtls/xray-core/common/mux.writeMetaWithFrame
github.com/xtls/xray-core/common/mux.(*Writer).writeData
github.com/xtls/xray-core/common/mux.(*Writer).WriteMultiBuffer
github.com/xtls/xray-core/common/buf.copyInternal
github.com/xtls/xray-core/common/buf.Copy
github.com/xtls/xray-core/common/mux.handle
github.com/xtls/xray-core/common/mux.(*ServerWorker).handleStatusNew.gowrap1
    common/mux/server.go:270
```

Real memory in use is only ~140 MB. The 11.6 EB allocation request is garbage — Go just read 8 bytes that happened to contain that value and asked the runtime to allocate it.

Tested on `v26.2.6`. The relevant code paths are unchanged on `master` as of today.

## Why it happens

Two small bugs that combine into a process-killing crash. Neither is enough on its own.

### Bug 1 — `common/mux/frame.go:103-106`: race on `b.UDP` + ignored error

```go
} else if b.UDP != nil {
    b.WriteByte(byte(TargetNetworkUDP))
    addrParser.WriteAddressPort(b, b.UDP.Address, b.UDP.Port)  // err discarded
}
```

`b.UDP` is a `*net.Destination` that the upstream UDP reader hands to the buffer (`writeMetaWithFrame` does `frame.UDP = data[0].UDP`). For UDP-over-mux every packet may go to a different address, so the producer keeps updating the same `Destination` for each packet it dispatches.

While the producer is mutating that struct, this code reads `b.UDP.Address` — a 16-byte interface (`itab` + `data`). Reading a two-word interface is **not atomic**. Under contention the reader sees a torn value: e.g. an `itab` for `DomainAddress` paired with a `data` pointer left over from the previous mutation. Following that stale pointer reads a `string` header out of unrelated memory, and you get a `(ptr, len)` pair where both fields are garbage.

The error from `WriteAddressPort` is also silently dropped — note that the corresponding TCP/`SessionStatusNew` branch a few lines above does check `if err := addrParser.WriteAddressPort(...); err != nil`. Looks like a copy-paste oversight.

### Bug 2 — `common/protocol/headers.go:93`: signed-int comparison

```go
func isDomainTooLong(domain string) bool {
    return len(domain) > 256
}
```

`len()` returns a signed `int`. After the torn read above, `len(domain)` can have its high bit set, i.e. it's a **negative** number. `negative > 256` is `false`, so the guard does nothing. `[]byte(domain)` then reaches `runtime.stringtoslicebyte`, which interprets the length as `uint`, asks for ~9 EB, and the runtime aborts.

Without bug 2 the process would survive (it would just write a malformed frame, since bug 1 swallows the error). Without bug 1 the corrupted address would never be reached.

## Why only UDP

The `SessionStatusNew` branch reads the address from `f.Target`, which is **a value field** of `FrameMetadata`. It is copied once when the sub-connection opens and never mutated. No race possible.

The `b.UDP` branch only fires for UDP keep frames, where each packet has its own destination and the address is read **through a pointer into a struct that the producer keeps writing to**. That's the only place in the file where this is possible.

## Reproducer

Standalone Go program. It does not race anything — instead it constructs the exact same corrupted `string` shape that a torn 16-byte interface read produces (real `Data` pointer, `Len = -1`), wraps it in a real `net.Destination`, and calls the real `mux.FrameMetadata.WriteTo`. The crash is bit-for-bit identical to the production stack, including the line numbers.

<details>
<summary><code>go.mod</code> and <code>main.go</code></summary>

`go.mod`:

```
module xray-mux-repro

go 1.25.7

require github.com/xtls/xray-core v0.0.0

replace github.com/xtls/xray-core => ../Xray-core
```

(Point the `replace` at a local checkout of `v26.2.6` or `master`.)

`main.go`:

```go
package main

import (
	"fmt"
	"os"
	"reflect"
	"unsafe"

	"github.com/xtls/xray-core/common/buf"
	"github.com/xtls/xray-core/common/mux"
	"github.com/xtls/xray-core/common/net"
)

func main() {
	// Real backing bytes — Data is a valid pointer, so we hit OOM, not a segfault.
	payload := []byte("a.example.com")

	// Build a string header with a negative Len. This is the exact shape a
	// torn 16-byte interface read of *b.UDP.Address can leave behind when
	// the upstream UDP reader concurrently mutates the Destination it lent
	// to the buffer.
	var s string
	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
	sh.Data = uintptr(unsafe.Pointer(&payload[0]))
	sh.Len = -1

	dest := net.Destination{
		Network: net.Network_UDP,
		Address: net.DomainAddress(s),
		Port:    net.Port(53),
	}

	b := buf.New()
	b.UDP = &dest

	meta := mux.FrameMetadata{
		SessionID:     1,
		SessionStatus: mux.SessionStatusKeep, // takes the `else if b.UDP != nil` branch
	}

	fmt.Fprintln(os.Stderr, "calling FrameMetadata.WriteTo with corrupted UDP destination...")
	if err := meta.WriteTo(b); err != nil {
		fmt.Fprintf(os.Stderr, "WriteTo returned error (patched build): %v\n", err)
		return
	}
	fmt.Fprintln(os.Stderr, "WriteTo returned nil")
}
```

</details>

Run against vanilla `v26.2.6`:

```
$ go run .
calling FrameMetadata.WriteTo with corrupted UDP destination...
fatal error: out of memory

runtime.stringtoslicebyte(..., {..., 0xffffffffffffffff})
    runtime/string.go:230
github.com/xtls/xray-core/common/protocol.(*addressParser).writeAddress
    common/protocol/address.go:251
github.com/xtls/xray-core/common/protocol.portFirstAddressParser.WriteAddressPort
    common/protocol/address.go:110
github.com/xtls/xray-core/common/mux.FrameMetadata.WriteTo
    common/mux/frame.go:105
```

Same line numbers, same call chain as production.

## Patch

Defense in depth — both fixes are tiny and independent, but you really want both. With both applied the reproducer cleanly returns `Super long domain is not supported: ...` instead of crashing.

<details>
<summary>Unified diff</summary>

```diff
diff --git a/common/protocol/headers.go b/common/protocol/headers.go
--- a/common/protocol/headers.go
+++ b/common/protocol/headers.go
@@ -91,5 +91,12 @@ func (sc *SecurityConfig) GetSecurityType() SecurityType {
 }

 func isDomainTooLong(domain string) bool {
-	return len(domain) > 256
+	// Use unsigned comparison: len() returns a signed int, and a torn read
+	// of a string header (e.g. via a racy 16-byte interface read of a
+	// *DomainAddress) can leave the high bit set. With the previous signed
+	// `> 256` check that case slipped through and reached
+	// runtime.stringtoslicebyte with a uint length of ~9 EB, crashing the
+	// process with "fatal error: out of memory". The wire format also only
+	// has one byte for the length, so 255 is the real upper bound.
+	return uint(len(domain)) > 255
 }
```

```diff
diff --git a/common/mux/frame.go b/common/mux/frame.go
--- a/common/mux/frame.go
+++ b/common/mux/frame.go
@@ -100,9 +100,22 @@ func (f FrameMetadata) WriteTo(b *buf.Buffer) error {
 		} else if b.UDP != nil { // make sure it's user's proxy request
 			b.Write(f.GlobalID[:]) // no need to check whether it's empty
 		}
-	} else if b.UDP != nil {
-		b.WriteByte(byte(TargetNetworkUDP))
-		addrParser.WriteAddressPort(b, b.UDP.Address, b.UDP.Port)
+	} else if udp := b.UDP; udp != nil {
+		// Snapshot the destination by value before serializing it. b.UDP is
+		// a *net.Destination borrowed from the upstream UDP reader (set by
+		// writeMetaWithFrame as `frame.UDP = data[0].UDP`), and reading
+		// udp.Address — a non-atomic two-word interface — races with the
+		// producer mutating the same Destination for its next packet. The
+		// race produces a torn DomainAddress that crashes
+		// addressParser.writeAddress with a multi-exabyte allocation.
+		// Copying the struct is still racy at the field level, but it
+		// closes the window between the type-byte write and the address
+		// write, and it lets us actually return the error below.
+		snap := *udp
+		common.Must(b.WriteByte(byte(TargetNetworkUDP)))
+		if err := addrParser.WriteAddressPort(b, snap.Address, snap.Port); err != nil {
+			return err
+		}
 	}

 	len1 := b.Len()
```

</details>

## Note on the underlying race

The patches above stop the crash and stop the malformed frame, but they don't fix the *root* of the data race — a producer that keeps mutating a `*net.Destination` after handing the pointer to a buffer is fundamentally unsafe, and the proper long-term fix is to make UDP producers either allocate a fresh `Destination` per packet or store it by value on `buf.Buffer`. I haven't tried to track down the specific producer that triggers it; happy to dig further if you'd like.

herenickname:
Fixes #5901.

`FrameMetadata.WriteTo` on the UDP keep-frame path dereferences `b.UDP.Address` (a non-atomic two-word interface) while the upstream UDP reader is free to mutate the same `*net.Destination` for its next packet. The resulting torn read can produce a `DomainAddress` whose backing string header has a garbage `(ptr, len)` pair.

`isDomainTooLong` then fails to catch it because it uses a signed `len(domain) > 256` comparison: a negative torn length is `< 256`, slips through, and reaches `runtime.stringtoslicebyte` with a `uint` length of ~9 EB, aborting the process with `fatal error: out of memory`.

The error from `WriteAddressPort` on this branch was also discarded, so even the existing guard would have produced a malformed frame instead of a clean failure.

This PR fixes both halves:

- **`common/protocol/headers.go`** — switch to `uint(len(domain)) > 255` so any negative or absurd length is rejected, and tighten the bound to the actual one-byte wire-format limit.
- **`common/mux/frame.go`** — snapshot `*b.UDP` into a local before writing the address bytes (closes the window between the network-type byte and the address bytes), and actually return `WriteAddressPort`'s error instead of dropping it.

Full root-cause writeup, production stack trace, and a deterministic standalone reproducer are in #5901. With this PR applied the reproducer cleanly returns `Super long domain is not supported: ...` instead of crashing.

## Note on the underlying race

This patch stops the crash and stops the malformed frame, but it does not fix the *root* of the data race — a producer that keeps mutating a `*net.Destination` after handing the pointer to a buffer is fundamentally unsafe, and the proper long-term fix is to make UDP producers either allocate a fresh `Destination` per packet or store it by value on `buf.Buffer`. I haven't tracked down the specific producer that triggers it; happy to follow up in a separate PR if you'd like.

### DIFF
--- a/common/mux/frame.go
+++ b/common/mux/frame.go
@@ -100,9 +100,22 @@ func (f FrameMetadata) WriteTo(b *buf.Buffer) error {
 		} else if b.UDP != nil { // make sure it's user's proxy request
 			b.Write(f.GlobalID[:]) // no need to check whether it's empty
 		}
-	} else if b.UDP != nil {
-		b.WriteByte(byte(TargetNetworkUDP))
-		addrParser.WriteAddressPort(b, b.UDP.Address, b.UDP.Port)
+	} else if udp := b.UDP; udp != nil {
+		// Snapshot the destination by value before serializing it. b.UDP is
+		// a *net.Destination borrowed from the upstream UDP reader (set by
+		// writeMetaWithFrame as `frame.UDP = data[0].UDP`), and reading
+		// udp.Address — a non-atomic two-word interface — races with the
+		// producer mutating the same Destination for its next packet. The
+		// race produces a torn DomainAddress that crashes
+		// addressParser.writeAddress with a multi-exabyte allocation.
+		// Copying the struct is still racy at the field level, but it
+		// closes the window between the type-byte write and the address
+		// write, and it lets us actually return the error below.
+		snap := *udp
+		common.Must(b.WriteByte(byte(TargetNetworkUDP)))
+		if err := addrParser.WriteAddressPort(b, snap.Address, snap.Port); err != nil {
+			return err
+		}
 	}
 
 	len1 := b.Len()
--- a/common/protocol/headers.go
+++ b/common/protocol/headers.go
@@ -91,5 +91,12 @@ func (sc *SecurityConfig) GetSecurityType() SecurityType {
 }
 
 func isDomainTooLong(domain string) bool {
-	return len(domain) > 256
+	// Use unsigned comparison: len() returns a signed int, and a torn read
+	// of a string header (e.g. via a racy 16-byte interface read of a
+	// *DomainAddress) can leave the high bit set. With the previous signed
+	// `> 256` check that case slipped through and reached
+	// runtime.stringtoslicebyte with a uint length of ~9 EB, crashing the
+	// process with "fatal error: out of memory". The wire format also only
+	// has one byte for the length, so 255 is the real upper bound.
+	return uint(len(domain)) > 255
 }


Fangliding:
除了你手动使用unsafe构造的非法数据并触发panic 我没在核心发现任何有尝试在赋值后修改UDP携带数据的位置 b.UDP 应该是一个不变量 如果你发现有对其的覆盖操作应该修复这点 至少我翻了一下核心没有发现 如果你发现可以指出

PR #5419, opened 2025-12-14
Browser dialer: High-performance pass-through rewrite

PoneyClairDeLune:
This PR is currently a draft, just #5334 re-opened because I fucked the Git commit history of the original PR up, as well as having the oversight of choosing the wrong source branch to use in pull requests.

This is a complete rewrite of the browser dialer to utilize pass-through without any processing on the browser side, reaching its theoretical maximum performance. A control socket will be used to instruct the browser dialer to initiate relaying on a dedicated separate channel.

The JS side, [Appat](https://github.com/team-cloudchaser/appat), is currently being built as an implementation-agnostic solution. I'll post a built version of it and unmark this as draft once it's done.

Issue #3840, opened 2024-09-21
[Feature suggestion & draft] Redesigned browser dialer


PoneyClairDeLune:
A continuation of #3832. Aimed at maximizing throughput and minimizing resource consumption on the browser side, and also potentially allowing arbitrary headers with the browser dialer.

## Prerequisites
This draft is designed with the following assumptions.

* The web server implementation in Go supports routable HTTP servers with streaming and WebSocket support.
* The implementation of maps in Go is 💫 *blazingly fast* 💫.
* The user has a recent installation of a modern browser (Chromium, Firefox or Safari).

## Design
Still not fleshed out. Critiques are welcomed!

#### `WS /control`
The control plane. This is the only place where any form of processing can happen with all possible latest web features. Xray can command the browser dialer on what and how to connect through it. If request bodies from Meek-like transports are not streamed, this should also carry the body of the request.

##### Message structure
| Byte length  | Required? | Use |
| ------------ | --------- | --- |
| 4 (`uint32`) | Y         | Length of payload. MSB to eliminate uncertainty of endianness with a somewhat fast implementation, or LSB to maximize speed. Should always be 0 for WebSocket connections and streamed web requests. |
| variable     | Y         | JSON-encoded command. |
| variable     | N         | Raw payload. Currently only used by non-streamed requests to post to remote. |

##### Commands
```js
{
	"m": "WS", // Method, can be "WS", "GET" or "POST"
	"p": "", // A string or an array of strings for expected **WebSocket** protocols
	"r": "", // Remote URL
	"i": 65535, // A numerical (positive integer) or a web-safe random string ID to distinguish connections
	"h": {} // A map of non-restricted custom headers to send to the server, cannot be used for WebSocket
}
```

#### `WS /data?id=<socketId>`
For Chrome 124 and later. Where contents of WebSocket connections are passed through without any processing.

[`WebSocketStream` - MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream)

#### `GET/POST /data?id=<socketId>`
For Chrome 105 and later. Where contents of all web-compliant connections are passed through without any processing. Has the exact same use as above, only that for each duplex connection, a get-post pair is created due to browsers refusing to support HTTP/2 cleartext.

[Send `ReadableStream` in request body - caniuse.com](https://caniuse.com/mdn-api_request_request_request_body_readablestream)

### Browser-side behaviour
| Criteria          | WebSocket behaviour            |
[...]

PR #5916, opened 2026-04-11
Global HTTP headers' masquerading: Add "curl"; Improve version generators

PoneyClairDeLune:
This PR added version generators for cURL, Firefox and Safari, as well as realistic headers for each. The RNG has also been migrated to become global.

The newer version generators still use randomized distributions, however it's now more deterministic, with a single setback scaled by a natural tailed curve, simulating common version distribution graphs.

I do suggest that the Chrome version generator should use the exact same approach to avoid version distribution avalanche caused by normal distribution. The current implementation of the Chrome version generator does not correlate well with the actual browser version distribution.

### DIFF
--- a/common/utils/browser.go
+++ b/common/utils/browser.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"hash/fnv"
+	"math"
 	"math/rand"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


RPRX:
~~Chrome 那个是我临时搞的，你觉得这个更好的话也换成这个吧~~

PoneyClairDeLune:
Just added Chrome to the party, all scaled with exponentials.

RPRX:
~~这个你检查过符合各个浏览器的发版规律吧，当初我搞 Chrome 那个时发现它偶尔会“休息”一个月，当初考虑到了这种情况~~

RPRX:
https://github.com/XTLS/Xray-core/pull/5658#issuecomment-3857083963 25-45 天就是这么来的

PoneyClairDeLune:
@RPRX I did a statistical counting from Chrome 127 all the way to 147. The average release duration counting all the delays is 31.15 days, although I might as well increase the average cycle to 32 just in case.
```
{28, 28, 28, 42, 35, 28, 28, 28, 42, 28, 28, 28, 28, 28, 21, 63, 28, 28, 28, 28}
```

RPRX:
你要取 avg 的话极容易出现“未发布的版本”，https://github.com/XTLS/Xray-core/pull/5658#issuecomment-3857159154 这里也说了可以每年修正下

RPRX:
就是说 Chrome 还是按 avg=35 来算吧，其它浏览器也是改成 avg 加几天

PoneyClairDeLune:
@RPRX Not really, however that will rely on the fixed setback to do the work. The current maximum is 63 days out of the 31.25 day average, a fixed setback of at least 32 day will currently sidestep the problem. Although I admit that having a single version ahead of normal stable release schedule will not be that suspicious, the exponential curve could amplify that.

Although there could be a better modelling curve for Chrome than the exponential curve. I probably need to look at some more curves to see which offers a skewed bell.

Edit: Replied too late.

PoneyClairDeLune:
So 35 days for Chrome and 30 for Firefox.

The current version generator of Safari (macOS and iOS seem to be the same, however Safari on iOS sends headers unpredictably) is heavily reliant on calendar instead of duration like the real deal, and it is pretty conservative based on most edge case observations.

RPRX:
~~保守点好，可以合并了吗~~

PoneyClairDeLune:
I think it is now ready for merging.

RPRX:
比起这个~~假的~~，真的 https://github.com/XTLS/Xray-core/pull/5419 https://github.com/XTLS/Xray-core/issues/3840 有进展吗

还有个问题，XHTTP 的 packet-up 模式在任何浏览器上都能工作，Chrome 理论上还支持 stream-up，现在的代码是支持的吗

PoneyClairDeLune:
@RPRX Streamed upload is a "yes" in theory, however it wasn't properly tested before updates to HTTP-based transports landed, so I have no idea if simply merging the changes to the current codebase will work.
Appat (name of the stream-only dialer) right now can interface and function with other programs without any error, but if support for non-Chromium browsers (browsers that don't support streaming) is required, it will not work until I can add fallback modes to those browsers.

Fangliding:
仍然无法解决最大6个ws连接的问题

RPRX:
> 仍然无法解决最大6个ws连接的问题

~~浏览器的 WS 连接数限制不是最大 256 吗~~

Fangliding:
per host

RPRX:
> per host

就是 per host，至少 Chromium 是这样，~~不然初版 browser dialer 不会那么奢侈的，而且似乎换个端口还重新计算~~

Fangliding:
@RPRX 
我还想到了一个问题 chrome 是会更新的 就像 tls fingerprint 自动推进的新版本号但是无新标头 直接是 unique fingerprint 了 有点聪明反被聪明误

RPRX:
@Fangliding 无论如何动态 Chrome headers 等肯定要有的啊，~~CDN 已经开始 gank 不那么 Chrome 的了 https://github.com/XTLS/Xray-core/issues/4996#issuecomment-3855274627~~

Fangliding:
写死的也很chrome了 只是自动推ua里版本号的问题

RPRX:
~~headers 相对稳定吧，现在外部 TLS 指纹还不一定选 Chrome 呢，这不是更不匹配，只能说尽力而为，Browser Dialer 兜底~~

PR #5920, opened 2026-04-12
header-custom finalmask: Add programmable handshake templates and runtime core

fatyzzz:
## Summary
- extend `finalmask/header-custom` with programmable handshake templates for both TCP and UDP
- add deterministic `capture`, `reuse`, and `transform` fields together with flow-scoped state and runtime-derived metadata
- wire the template model into the existing `header-custom` runtime while keeping existing `packet` / `rand` behavior working

## Why
`header-custom` currently supports fixed bytes and random bytes only. That is enough for static wrappers, but not for handshakes that need captured fields, reused values, deterministic derived fields, or request/response correlation.

This PR keeps the template/config layer and the corresponding runtime core together so the feature is self-contained.

## Scope
- expand `header-custom` config with captured, reused, and transformed fields
- support bounded deterministic transforms: `concat`, `slice`, `xor16`, `xor32`, `be16`, and `be32`
- add flow-scoped state and runtime-derived metadata support for `header-custom`
- wire the new template semantics into the TCP and UDP `header-custom` runtime
- keep backward compatibility with existing `packet` / `rand` configs

## Non-Goals
- no loops
- no branching
- no protocol-specific helpers
- no detached UDP standalone mode in this PR

## Test Plan
- [x] `go test ./infra/conf -run 'Test(HeaderCustomUDPBuild|HeaderCustomTCPBuildRejectsMixedItemKinds|HeaderCustomUDPBuildRejectsInvalidVariableNames|HeaderCustomUDPBuildRejectsExprWithoutArgs)' -count=1`
- [x] `go test ./transport/internet/finalmask/header/custom -run 'Test(Evaluator.*|Metadata.*|State.*|DSLTCP.*|DSLUDP.*)' -count=1`
- [x] `go test ./transport/internet/finalmask -run 'Test(ConnReadWrite|PacketConnReadWrite|TCPcustomStaticHandshakeRoundTrip|TCPcustomClientRejectsMismatchedServerSequence|UDPcustomStaticHeaderWireShape|UDPcustomServerRejectsMismatchedStaticHeader)' -count=1`

### DIFF
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -1265,14 +1265,14 @@ var (
 )
 
 type TCPItem struct {
-	Delay     Int32Range      `json:"delay"`
-	Rand      int32           `json:"rand"`
-	RandRange *Int32Range     `json:"randRange"`
-	Save      string          `json:"save"`
-	Type      string          `json:"type"`
-	Var       string          `json:"var"`
-	Expr      *CustomExpr     `json:"expr"`
-	Packet    json.RawMessage `json:"packet"`
+	Delay     Int32Range       `json:"delay"`
+	Rand      int32            `json:"rand"`
+	RandRange *Int32Range      `json:"randRange"`
+	Capture   string           `json:"capture"`
+	Type      string           `json:"type"`
+	Reuse     string           `json:"reuse"`
+	Transform *CustomTransform `json:"transform"`
+	Packet    json.RawMessage  `json:"packet"`
 }
 
 type HeaderCustomTCP struct {
@@ -1284,21 +1284,21 @@ type HeaderCustomTCP struct {
 func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 	for _, value := range c.Clients {
 		for _, item := range value {
-			if err := validateCustomItemSpec(item.Save, item.Packet, item.Rand, item.Var, item.Expr); err != nil {
+			if err := validateCustomItemSpec(item.Capture, item.Packet, item.Rand, item.Reuse, item.Transform); err != nil {
 				return nil, err
 			}
 		}
 	}
 	for _, value := range c.Servers {
 		for _, item := range value {
-			if err := validateCustomItemSpec(item.Save, item.Packet, item.Rand, item.Var, item.Expr); err != nil {
+			if err := validateCustomItemSpec(item.Capture, item.Packet, item.Rand, item.Reuse, item.Transform); err != nil {
 				return nil, err
 			}
 		}
 	}
 	for _, value := range c.Errors {
 		for _, item := range value {
-			if err := validateCustomItemSpec(item.Save, item.Packet, item.Rand, item.Var, item.Expr); err != nil {
+			if err := validateCustomItemSpec(item.Capture, item.Packet, item.Rand, item.Reuse, item.Transform); err != nil {
 				return nil, err
 			}
 		}
@@ -1320,7 +1320,7 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 			if item.Packet, err = PraseByteSlice(item.Packet, item.Type); err != nil {
 				return nil, err
 			}
-			expr, err := buildCustomExpr(item.Expr)
+			transform, err := buildCustomTransform(item.Transform)
 			if err != nil {
 				return nil, err
 			}
@@ -1331,9 +1331,9 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 				RandMin:  item.RandRange.From,
 				RandMax:  item.RandRange.To,
 				Packet:   item.Packet,
-				Save:     item.Save,
-				Var:      item.Var,
-				Expr:     expr,
+				Save:     item.Capture,
+				Var:      item.Reuse,
+				Expr:     transform,
 			})
 		}
 	}
@@ -1352,7 +1352,7 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 			if item.Packet, err = PraseByteSlice(item.Packet, item.Type); err != nil {
 				return nil, err
 			}
-			expr, err := buildCustomExpr(item.Expr)
+			transform, err := buildCustomTransform(item.Transform)
 			if err != nil {
 				return nil, err
 			}
@@ -1363,9 +1363,9 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 				RandMin:  item.RandRange.From,
 				RandMax:  item.RandRange.To,
 				Packet:   item.Packet,
-				Save:     item.Save,
-				Var:      item.Var,
-				Expr:     expr,
+				Save:     item.Capture,
+				Var:      item.Reuse,
+				Expr:     transform,
 			})
 		}
 	}
@@ -1384,7 +1384,7 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 			if item.Packet, err = PraseByteSlice(item.Packet, item.Type); err != nil {
 				return nil, err
 			}
-			expr, err := buildCustomExpr(item.Expr)
+			transform, err := buildCustomTransform(item.Transform)
 			if err != nil {
 				return nil, err
 			}
@@ -1395,9 +1395,9 @@ func (c *HeaderCustomTCP) Build() (proto.Message, error) {
 				RandMin:  item.RandRange.From,
 				RandMax:  item.RandRange.To,
 				Packet:   item.Packet,
-				Save:     item.Save,
-				Var:      item.Var,
-				Expr:     expr,
+				Save:     item.Capture,
+				Var:      item.Reuse,
+				Expr:     transform,
 			})
 		}
 	}
@@ -1504,27 +1504,27 @@ func (c *NoiseMask) Build() (proto.Message, error) {
 }
 
 type UDPItem struct {
-	Rand      int32           `json:"rand"`
-	RandRange *Int32Range     `json:"randRange"`
-	Save      string          `json:"save"`
-	Type      string          `json:"type"`
-	Var       string          `json:"var"`
-	Expr      *CustomExpr     `json:"expr"`
-	Packet    json.RawMessage `json:"packet"`
+	Rand      int32            `json:"rand"`
+	RandRange *Int32Range      `json:"randRange"`
+	Capture   string           `json:"capture"`
+	Type      string           `json:"type"`
+	Reuse     string           `json:"reuse"`
+	Transform *CustomTransform `json:"transform"`
+	Packet    json.RawMessage  `json:"packet"`
 }
 
-type CustomExpr struct {
-	Op   string          `json:"op"`
-	Args []CustomExprArg `json:"args"`
+type CustomTransform struct {
+	Op   string               `json:"op"`
+	Args []CustomTransformArg `json:"args"`
 }
 
-type CustomExprArg struct {
-	Type     string          `json:"type"`
-	Bytes    json.RawMessage `json:"bytes"`
-	U64      *uint64         `json:"u64"`
-	Var      string          `json:"var"`
-	Metadata string          `json:"metadata"`
-	Expr     *CustomExpr     `json:"expr"`
+type CustomTransformArg struct {
+	Type      string           `json:"type"`
+	Bytes     json.RawMessage  `json:"bytes"`
+	U64       *uint64          `json:"u64"`
+	Reuse     string           `json:"reuse"`
+	Metadata  string           `json:"metadata"`
+	Transform *CustomTransform `json:"transform"`
 }
 
 func validateCustomVarName(name string) error {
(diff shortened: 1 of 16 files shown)


fatyzzz:
Thanks. This PR intentionally focuses on the schema/evaluator layer only, so I avoided adding transport-specific runtime examples to keep the review scope narrow.

The corresponding runtime-oriented config usage is split into the follow-up stacked PRs:
- metadata / flow state / runtime integration
- UDP standalone mode for detached handshake packets

If it helps the review, I can also add a small generic `header-custom` JSON snippet here that demonstrates the new item shapes (`save`, `var`, `expr`, etc.) without tying it to any protocol-specific example.

Fangliding:
算了这太复杂了 最后加一坨都没人知道怎么用 感觉没有现实意义

fatyzzz:
> 算了这太复杂了 最后加一坨都没人知道怎么用 感觉没有现实意义

 But Xray-core was never meant to be the simplest solution in the first place.
  It has always been a tool with a lot of advanced, engineering-oriented capabilities.

  And if we are talking about censorship resistance, then I think this is actually a meaningful step.
  The value is precisely that different people can shape their own traffic patterns instead of everyone reusing the same fixed template.

  Not every feature has to be for everyone.
  In this case, the point is to give capable users more room to build their own behavior, and that itself has practical value in an anti-censorship setting.

RPRX:
可以接受但设计需要简单通俗些，让大多数人知道该怎么用

fatyzzz:
> 可以接受但设计需要简单通俗些，让大多数人知道该怎么用

I think there are two different ways to simplify this, and they have very different trade-offs.

If the concern is mainly about UX and how the feature is presented, then I can simplify the surface without really losing capability:
present it as reusable packet templates with captured/reused fields and a small set of deterministic transforms, instead of making it feel like a mini-language.

But if the expectation is to actually remove syntax/operations from the design itself, then we would start losing generality.
That would make the feature easier to explain, but it would also reduce its usefulness for more varied protocol shapes.

So I want to clarify which direction you mean by “make it simpler”:
1. simpler presentation / simpler user-facing model, while keeping the underlying capability
2. or a real reduction of the syntax and supported operations

Those two lead to very different designs.

RPRX:
1

刚刚才看到 @iambabyninja 也发邮件说了这个，~~算是“完全可编程协议”的青春版吧，毕竟那个更复杂我都不确定好不好实现~~

目前可以先把代理和加密的部分完全交给 VLESS Encryption random，然后 finalmask 做外部流量包装和整形，弄成任何东西

~~至于 Flow，它在 VLESS Encryption 内部~~

fatyzzz:
> 1
> 
> 刚刚才看到 @iambabyninja 也发邮件说了这个，~算是“完全可编程协议”的青春版吧，毕竟那个更复杂我都不确定好不好实现~
> 
> 目前可以先把代理和加密的部分完全交给 VLESS Encryption random，然后 finalmask 做外部流量包装和整形，弄成任何东西
> 
> ~至于 Flow，它在 VLESS Encryption 内部~

Yes, I agree. I’ll think about how to make the design simpler and easier to understand, while keeping the underlying flexibility.

fatyzzz:
I agree with the concern about making the design easier to understand.

I have reworked the user-facing config surface in this PR to make it read more like programmable handshake packet templates, instead of a mini-language:

- kept the existing `packet` / `rand` fields unchanged
- renamed the new field tools to clearer template-oriented terms:
  - `save` -> `capture`
  - `var` -> `reuse`
  - `expr` -> `transform`
- applied the same terminology to nested transform arguments
- updated the focused config tests accordingly

The underlying capability is intentionally unchanged:
- no functionality was removed
- bounded deterministic transforms are still supported
- captured/reused fields are still supported
- the later runtime PRs can still build on the same model

So the goal of this update is to simplify the UX and mental model, without reducing the flexibility needed for programmable handshake shaping.

If you think the public model should be simplified even further, I can continue iterating in that direction.

LjhAUMEM:
这前后也不一致啊，5919 infra 只是新增 save var expr，到 5920 瞬间变成 capture reuse transform
image
image

~说实话不是很想来审这种 vibe 啊~

~另外增加表达式操作感觉不如加个 fallback 有用~

fatyzzz:
> ~说实话不是很想来审这种 vibe 啊~
> 
> ~另外增加表达式操作感觉不如加个 fallback 有用~

The rename was intentional, not arbitrary.

The earlier feedback was that the design should be simpler and easier to understand, so I adjusted the user-facing terminology to make it read less like a mini-DSL and more like packet-template fields.

I agree that changing naming during review adds churn, so I should stabilize the vocabulary more carefully.

But the goal of the rename was to improve UX, not to change the underlying model.

As for fallback, I agree it is useful, but it solves a different problem.
Fallback helps when matching fails; transforms help construct and correlate handshake fields in the first place.
So I do not think they are interchangeable.

LjhAUMEM:
> The rename was intentional, not arbitrary.

Okay, I noticed that this reuse seed wasn't captured before. What is the expected behavior?

image

fatyzzz:
> What is the expected behavior?

Good catch.

In the current implementation, unresolved reuse is not rejected at config-build time. It is only rejected later at runtime evaluation, where it fails with an unknown variable error.

So this test was only meant to validate the parser/config shape, not to imply that an uncaptured field can actually be reused successfully.

That said, I agree this is a poor example for a user-facing review context, because it makes the model look more confusing than it should. I’ll adjust the test case to use a properly captured field instead.

Also, this is my first PR in Xray-core, and I’m glad to have the opportunity to communicate with the maintainers and contributors here while iterating on the design.

Fangliding:
> 可以接受但设计需要简单通俗些，让大多数人知道该怎么用

我自己搓个格式用标准语言还得写一堆函数和变量呢 用这抽象玩意得多难写啊

fatyzzz:
> 用这抽象玩意得多难写啊

Sure, writing this is harder than writing a normal program directly.

But the goal here is not to replace a real programming language.

The goal is to provide a bounded in-config mechanism for programmable handshake shaping inside Xray itself.

So I do not expect it to be “easy” in the same sense as a simple config.
The real question is whether this bounded model is useful enough for the cases it targets.

RPRX:
~~尽量不要 vibe 吧，除非人工校正过~~

fatyzzz:
> ~尽量不要 vibe 吧，除非人工校正过~

yeah, sorry about that

i pushed it once it was working and already usable for me, and only later realized i had not thought enough about the UX side

i’m probably not the pickiest user myself :)

will try to keep it less vibe and more polished from here

fatyzzz:
I restacked the series and rebased it onto current `main`.

There were two reasons for this:
1. the previous split between the template/config layer and the runtime/state layer was not the best review boundary
2. the branch had also drifted behind current `main`

So this PR now represents the self-contained `header-custom` core:
- template/config surface
- evaluator
- state / metadata
- TCP and UDP runtime integration

The remaining follow-up PR is now only for detached UDP standalone mode.

After the restack, I reran the focused tests and local live smoke checks for both TCP and UDP paths.

RPRX:
这部分目前我用不到，也不太想烧脑细胞来确定怎么设计才是“最优”，所以我的态度类似对于 XHTTP 那些新参数，谁行谁上

具体而言，谁用谁设计，只要不影响到已有功能就行，我会合并，文档请自行 PR，未来若需要 breaking 也可以接受

算是实现了半个“完全可编程协议”的饼，也避免了用户自行搞出不够安全的加密，从这个角度来说这种方式更好

不过这只是 header-custom，还需要一个可编程 transport，或者也放 finalmask 里然后加个 [pre](https://github.com/XTLS/Xray-core/pull/5850#issuecomment-4154647029)（生效在 tls/reality 内部）

---

@patterniha 最近在整些新活导致 Project X 伊朗群 [Project XHTTP](t.me/projectXhttp) 这两天涌入了上万人，那些花活想加的话也可以 PR

fatyzzz:
> 这部分目前我用不到，也不太想烧脑细胞来确定怎么设计才是“最优”，所以我的态度类似对于 XHTTP 那些新参数，谁行谁上

Thanks, that makes sense.

I’m glad this direction looks acceptable to you.

I’ll keep the follow-up work scoped to the parts I actually use, and I’ll keep it from affecting existing behavior unless there is a strong reason to change it.

I’ll send the documentation as a separate PR.

For now, the remaining follow-up on my side is the detached UDP standalone part. If I continue in the programmable transport / finalmask-pre direction later, I’ll keep those changes separate and narrowly scoped
as well.

RPRX:
这刚加的功能，一段时间内用户也不多，如果想出了更好的架构就大胆 break 配置，没必要保持兼容

“detached UDP standalone part”你指的是 noise 吧？~~如果不方便改现有 finalmask 的话可以单独出个 handshake~~

话说现在 header-custom TCP 已经实现了不带任何上层数据的纯 handshake 吧？@Fangliding 文档中咋有俩 header-custom

Fangliding:
一个是tcp的一个是udp的啊

LjhAUMEM:
headerReadAddrAware 这个很迷啊，headerManagerConn 里 writeto 的 addr 是无意义的，而且也没有 conn 有实现 headerReadAddrAware 这个 interface，~看错了 udp header custom 实现了~，但是也没必要，addr 对于 header conn 没有什么用

image

fatyzzz:
> headerReadAddrAware 这个很迷啊，headerManagerConn 里 writeto 的 addr 是无意义的，而且也没有 conn 有实现 headerReadAddrAware 这个 interface，~看错了 udp header custom 实现了~，但是也没必要，addr 对于 header conn 没有什么用

The `addr` here is not completely meaningless. It is mainly used by UDP `header-custom`.

On the read path, `headerManagerConn` gets the peer address from the underlying `PacketConn.ReadFrom` and passes it into `header-custom` through `SetReadAddr`, so matching, save/reuse, and state updates can be scoped per peer.

On the write path, the `addr` also goes into the eval context and state key. For a purely static header it may indeed be irrelevant, but for UDP `header-custom` with runtime metadata and per-peer saved state it is still useful.

So `headerReadAddrAware` is not meant for every header conn. It is only there for peer-aware UDP `header-custom`.

LjhAUMEM:
~原来如此，表达式还和 addr 绑定了吗，那行吧~

patterniha:
I recently noticed that the https://github.com/ValdikSS/goodbyeDPI project has all 3 of my implemented methods and even more.

I implemented:
--wrong-seq
--fake-with-sni
--wrong-chksum
--set-ttl

Although I haven't tested it yet, the explanation given is exactly the same as my methods.

RPRX:
@patterniha 有兴趣的话可以 PR 几个 finalmask

MoRanYue:
Actually, a lot of applications were trapped into a "using configuration to program" pit, which is always leading to bad experience and limited usage cases, containing the both disadvantages compare to direct programming.

LjhAUMEM:
> Actually, a lot of applications were trapped into a "using configuration to program" pit, which is always leading to bad experience and limited usage cases, containing the both disadvantages compare to direct programming.

然后还不是可编程，只是给 header 加上了科学计算器的功能，为此还多了两个只服务于 udp header custom 的 interface 来污染 headerManagerConn，成功让我现在碰都不想碰 fm

MoRanYue:
> 然后还不是可编程，只是给 header 加上了科学计算器的功能，为此还多了两个只服务于 udp header custom 的 interface 来污染 headerManagerConn，成功让我现在碰都不想碰 fm

（因此这里才表示“此类设计同时具备了‘灵活度低’与‘编写体验差’的缺点”。这里不知道为何许多开发者偏好“使用配置文件来编程”，性能差、可读性差、灵活度低，需要灵活便会使配置变得更臃肿。根本没有思考过此类设计是弊大于利的，这里很不喜欢这样。真正的“可编程”应当是引入Lua API或创建库API，其它开发者编译动态链接库来实现“插件”之目的，而不是耦合入配置文件。）

RPRX:
> 真正的“可编程”应当是引入Lua API或创建库API，其它开发者编译动态链接库来实现“插件”之目的，而不是耦合入配置文件。

前者有人提过但尚不确定会使 Xray 体积膨胀多少，后者在手机 APP 上不太能过审且有闭源问题

~~其实我本来是想引入个动态解析 Go 代码的库，且可以调用 Xray-core 内的函数，不过同样不确定会使 Xray 体积膨胀多少~~

再研究研究吧，下半年的事情了，有好的建议也可以提出

MoRanYue:
（这里认为，引入“动态解析Go代码”的复杂性 可能相比Lua等本身便是动态类型语言的复杂性高，而2种方式最终增加的二进制文件大小可能相差无几，毕竟基于脚本的插件系统均需要1个语言解释器来执行它。

动态链接库的确容易出现安全问题，设计API需要极其小心。emm，或许基于WebAssembly的插件系统会是更好的选择，跨平台兼容性很好，性能高且权限更易于限制；只是编译需要带有WASM运行时，增加几十MB的大小。

是否有其它更好的方法呢……？）

PR #5922, opened 2026-04-12
buffer writing optimized

ZeN220:
in `common/buf/writer.go` there a was block code like this:
```go
if !w.buffered {
    if writer, ok := w.writer.(io.Writer); ok {
       return writer.Write(b)
    }
}
```
Since the class in `transport/pipe/writer.go` didn't implement the `Write` method, this branch was being skipped resulting in a minimum buffer allication of 8 KB. i have added implementation for this method and now the allocation is reduced to 2 KB (if a payload size allows). this parameter is regulated by the `common/bytespool` package

i would also to broadly consider reducing buffer size wherever possible. after a quick review of the project's codebase, i noticed that 8 KB is allocated in most places on `buf.Size`. does such a larger allocation make sense in all of the cases?

### DIFF
--- a/transport/pipe/writer.go
+++ b/transport/pipe/writer.go
@@ -9,6 +9,33 @@ type Writer struct {
 	pipe *pipe
 }
 
+// Write implements io.Writer.
+func (w *Writer) Write(p []byte) (int, error) {
+	if len(p) == 0 {
+		return 0, nil
+	}
+
+	total := len(p)
+	mb := make(buf.MultiBuffer, 0, (len(p)+buf.Size-1)/buf.Size)
+	for len(p) > 0 {
+		chunkSize := len(p)
+		if chunkSize > buf.Size {
+			chunkSize = buf.Size
+		}
+
+		chunk := buf.NewWithSize(int32(chunkSize))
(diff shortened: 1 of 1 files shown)


wyx2685:
打出一个问号

ZeN220:
> 打出一个问号

excuse me? for what purpose and where should this be done?

Fangliding:
BufferedWriter 没有和pipe进行大流量交互的地方 这个writer只是为**原生**io.Writer留的一个fastpath而已 就算是个 pipe.Writer 它也没法节省什么东西 一样要进行一次数据拷贝 更别说pipe本身就有512kb的缓存

至于 8kb 的单块buffer 它对内存使用的影响没有你想象的那么大 毕竟它只是个临时的搬运工可以被高强度复用 不会导致内存剧烈膨胀或者大量分配造成gc压力 标准库许多buffer大小在 4kb~32kb间 你可能看到过 v2fly是2048 但是像singbox的一个buffer甚至高达32kb 8kb算是一个折中数值 core 大多数地方不强依赖8kb的值 你可以在合理范围内尝试自己改动编译观察影响

PR #5924, opened 2026-04-13
DomainMatcher: Reduce runtime memory usage and startup peak memory on iOS

### LINKED ISSUE #4422: iOS packet tunnel provider limit
Original discussion: https://github.com/XTLS/Xray-core/discussions/4398

Good afternoon, dear members of our community!

Specifically, we know for certain that in the current versions of iOS, we have the following memory limitations:

| Provider        | Limit (MiB) |
|-----------------|-------------|
| packet tunnel   | 50          |
| app proxy       | 15          |
| filter control  | 50          |
| filter data     | 50          |
| DNS proxy       | 15          |
| app push        | 24          |

In our case, Xray is a network process limited to 50MB. It reads geo files and then operates with the read content.

Unfortunately, with the current implementation of geofiles in our core, where all categories are in one large file, we often encounter the inability to even simply start the tunnel on an Apple device.
In particular, with the new geolocation files that we received from our partners, the file sizes have become significantly larger, which means that we cannot load them into memory.

A simple example (Loyalsoldier geosite):
image
None of the first four categories, if set in the routing rules, will allow the tunnel to start.

This also applies to geo files with CIDR (Loyalsoldier geoip)
Simple example:
image

@RPRX  , considering the memory constraints, perhaps we should explore refactoring the geofile handling code.  We could draw inspiration from neighboring teams and implement a system that utilizes multiple smaller files instead of one monolithic file. This approach could effectively resolve the memory issues on Apple devices.  What you think about this?

Meo597:
Closes #4422

### DIFF
--- a/common/geodata/domain_matcher.go
+++ b/common/geodata/domain_matcher.go
@@ -2,7 +2,9 @@ package geodata
 
 import (
 	"context"
+	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/xtls/xray-core/common/errors"
 	"github.com/xtls/xray-core/common/geodata/strmatcher"
@@ -13,7 +15,14 @@ type DomainMatcher interface {
 	MatchAny(input string) bool
 }
 
-func buildDomainMatcher(rules []*DomainRule) (DomainMatcher, error) {
+type DomainMatcherFactory interface {
+	BuildMatcher(rules []*DomainRule) (DomainMatcher, error)
+}
+
+type MphDomainMatcherFactory struct{}
+
+// BuildMatcher implements DomainMatcherFactory.
+func (f *MphDomainMatcherFactory) BuildMatcher(rules []*DomainRule) (DomainMatcher, error) {
 	g := strmatcher.NewMphValueMatcher()
 	for i, r := range rules {
 		switch v := r.Value.(type) {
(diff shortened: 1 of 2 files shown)


Meo597:
绝大多数配置中 dns 和 router 的 domain rule 会有很多交集
现在它们能共用一份内存以降低运行时占用
image

其实仍有进一步优化空间让 geosite:category-ads-all 能用起来，这玩意会吃掉 40mb
以后再说，现在应该够用了

t-e-s-tweb:
@Meo597 request to add this android too here "case "ios":".  I tested it with android added and it reduces the startup time from 3-5 secs to nearly instantaneous and routing works fine.

Meo597:
@t-e-s-tweb 你测过我就放心多了！

android 内存也紧张吗？
这里的 case "ios" 影响的只是启动时的峰值内存，代价是匹配性能低数百纳秒。

t-e-s-tweb:
> @t-e-s-twebI feel much more at ease now that you've tested it!
> 
> Is Android also experiencing memory constraints? The "ios" case here only affects peak memory usage at startup, at the cost of a few hundred nanoseconds of slower matching performance.

No, memory isn't the problem. But the app takes ages to start up when you're using "category-ads-all," even with just a 3MB geosite and the Hagezi list. Since the app gets started and stopped a bunch of times, way more often than a router would, all those extra seconds really stack up.

Meo597:
明白你的意思了
android 用的是默认 mphmatcher
它在移动端可能需要编译几秒

如果从缩短启动时间的角度，那么 android 也可以加

Meo597:
iOS 内存还可以进一步往下压
cn !cn ads-all 加一起用不到 10mb

但鉴于一次重构弄出三个 issue 尿不尽一样
还是以后再说吧

PR #5955, opened 2026-04-16
fix: Fix splithttp uploadQueue off-by-one bug in `Read()` where poppi…

OctopusTakopi:
Fix splithttp uploadQueue off-by-one bug in `Read()` where popping a packet https://github.com/XTLS/Xray-core/blob/c5edc122b70ec56e24ea8038e727da6f823e34be/transport/internet/splithttp/upload_queue.go#L107 reduced the heap length, causing the queue to bypass the maxPackets check.


Fangliding:
~~不过多一个包+1而已影响不大~~

OctopusTakopi:
With only N+1 packets, heap maxes at N+1, post-pop check sees N,  never fires → blocks on empty channel.

Issue #5944, opened 2026-04-15
iOS high memory usage/memory leak


nbghere:
### Integrity requirements

- [x] I have read all the comments in the issue template and ensured that this issue meet the requirements.
- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

Memory leaks are persistent when using certain iOS apps. The pprof profiles are attached.

[Metrics.zip](https://github.com/user-attachments/files/26747302/Metrics.zip)

### Reproduction Method

none

### Client config

```
{
  "dns" : {
    "queryStrategy" : "UseIPv4"
  },
  "inbounds" : [
    {
      "listen" : "127.0.0.1",
      "port" : 10808,
      "protocol" : "socks",
      "settings" : {
        "auth" : "noauth",
        "udp" : true
      },
      "sniffing" : {
        "destOverride" : [
          "quic",
          "http",
          "tls"
        ],
        "enabled" : true,
        "excludedDomains" : [

        ],
        "metadataOnly" : false,
        "routeOnly" : true
      },
      "tag" : "socks-in"
    },
    {
      "listen" : "127.0.0.1",
      "port" : 10809,
      "protocol" : "http",
      "settings" : {
        "auth" : "noauth",
        "udp" : true
      },
      "sniffing" : {
        "destOverride" : [
          "quic",
          "http",
          "tls"
        ],
        "enabled" : true,
        "excludedDomains" : [

        ],
        "metadataOnly" : false,
        "routeOnly" : true
      },
      "tag" : "http"
    },
    {
      "listen" : "127.0.0.1",
      "port" : 10820,
      "protocol" : "socks",
      "settings" : {
        "auth" : "noauth",
        "udp" : true
      },
      "sniffing" : {
        "destOverride" : [
          "quic",
          "http",
          "tls"
        ],
        "enabled" : true,
        "excludedDomains" : [

        ],
        "metadataOnly" : false,
        "routeOnly" : true
      },
      "tag" : "socks-direct"
    }
  ],
  "outbounds" : [
    {
      "protocol" : "vless",
      "settings" : {
        "testpre" : null,
        "testseed" : null,
        "vnext" : [
          {
            "address" : "example.com",
            "port" : 443,
[...]

PR #5957, opened 2026-04-16
splithttp: fix resource leaks on client reconnects

alemhops:
Just the leak/release/race subset of #5944. Small and localized

- `DialerClient` now has a `Close()` method that calls `CloseIdleConnections` on the underlying http2/http3 transport. `GetXmuxClient` calls it before pruning a client from the pool (`mux.go`). Before, the dead transport sat there holding its sockets for up to `ConnIdleTimeout` (~5 min) while a new one was being spun up right next to it, which is the overlap pattern the issue describes.

- `DefaultDialerClient.closed` was a plain bool written from the `client.Do` goroutine and read from `IsClosed()` with no sync (`client.go:34`). `go build -race` catches it. Moved to `atomic.Bool`.

- Two upload paths that were leaking pooled `MultiBuffer`s: `remainder` in the packet-up upload goroutine (`dialer.go:525`) is now released via a `defer` (before it was silently overwritten on the next outer-loop read), and the tail in `uploadWriter.Write` (`dialer.go:610`) when `WriteMultiBuffer` errors mid-loop.

- H1 retry loop in `PostPacket` (`client.go:155`) closes the failed `H1Conn` before grabbing a new one from the pool. Before, the dead pooled socket just leaked.

- `Dial` (`dialer.go:443`) has a rollback `defer`. If stream-down opens and stream-up fails after, `OpenUsage.Add(1)` gets undone and the already-running stream-down goroutine is torn down. Before, every partial-dial failure leaked a stream goroutine and two refcounts.

Things I spotted but left out of this PR, can do follow-ups if useful:

- `SetDeadline` / `SetReadDeadline` / `SetWriteDeadline` in `connection.go:51-64` are `return nil` stubs. Fixing properly means replacing `io.Pipe` with something that actually supports deadlines.
- `context.WithoutCancel` in `OpenStream` / `PostPacket`, plus the upload goroutine spawned per chunk in packet-up (`dialer.go:560`) is never awaited or cancelled.
- `http2.Transport` flow-control windows. `x/net/http2` doesn't let you set per-stream / per-connection receive windows from outside `Transport` cleanly, needs discussion on approach.
- `PostPacket` body-placement wraps `MultiBuffer` in `io.NopCloser`, which leaks on error (NopCloser.Close is a no-op).
- `globalDialerMap` never removes entries.

### DIFF
--- a/transport/internet/splithttp/browser_client.go
+++ b/transport/internet/splithttp/browser_client.go
@@ -21,6 +21,12 @@ func (c *BrowserDialerClient) IsClosed() bool {
 	panic("not implemented yet")
 }
 
+// Close is a no-op for the browser dialer client; the browser itself
+// owns the underlying connection lifecycle.
+func (c *BrowserDialerClient) Close() error {
+	return nil
+}
+
 func (c *BrowserDialerClient) OpenStream(ctx context.Context, url string, sessionId string, body io.Reader, uploadOnly bool) (io.ReadCloser, net.Addr, net.Addr, error) {
 	if body != nil {
 		return nil, nil, nil, errors.New("bidirectional streaming for browser dialer not implemented yet")
--- a/transport/internet/splithttp/client.go
+++ b/transport/internet/splithttp/client.go
@@ -8,17 +8,24 @@ import (
 	"net/http"
 	"net/http/httptrace"
 	"sync"
+	"sync/atomic"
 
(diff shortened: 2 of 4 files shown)


Fangliding:
AI仍然无法弄清楚 CloseIdleConnections() 不是 Close()

Issue #2596, opened 2023-09-27
[feature request] Support reload xray process


ko1ebayev:
Ni Hao xray developers! Id like to reload xray process when updating my config, so can you please support something like SIGHUP signal or reload functionality in upcoming releases?

chika0801:
This is a good suggestion. Do you want the ability to use the systemctl reload xray command to reload the config file without restarting the Xray process? Maybe wait for a fated person to add the code.

ko1ebayev:
I think it will be easier and more useful to support cross-platform reload process. I also don’t have a very good idea of ​​what kind of implementations there might be, but I see some kind of separate command for updating the config on the fly, something like `xray update --config=/path/to/json`. Maybe I will make a pull-request with this feature later

PR #5994, opened 2026-04-21
Implement SIGHUP reloading mechanism

Waujito:
Adds a reload mechanism using SIGHUP signal, allowing to restart without exiting the actual process. Issue XTLS/Xray-core#2596 was closed with no explanation as non-planned, but this logic may be really useful in some cases.


Fangliding:
不是这么弄的 这样可能导致很多问题而且不是reload

PR #5995, opened 2026-04-22
wireguard: make netBind.Close idempotent

abolix:
WireGuard outbound cleanup could panic when `netBind.Close()` was invoked more than once (including concurrent/deferred paths), due to an unguarded `close(closedCh)`. There was also a lifecycle gap where `closedCh` could be nil before `Open()`, which is unsafe for goroutines selecting on it.

- **Close path hardening**
  - Added `closeOnce sync.Once` to `netBind`.
  - Updated `Close()` to close `closedCh` via `closeOnce.Do(...)`, with nil guard, making repeated/concurrent `Close()` calls safe.

- **Channel lifecycle correctness**
  - Initialized `closedCh` when constructing `netBind` in `processWireGuard` so it exists before any goroutine may select on it.
  - Updated `Open()` to only initialize `closedCh` if it is nil, preserving constructor initialization and avoiding unnecessary re-creation.

- **Regression coverage**
  - Added `proxy/wireguard/bind_test.go` with a focused test that calls `Close()` repeatedly and concurrently, asserting no panic.

```go
func (bind *netBind) Close() error {
	errors.LogDebug(context.Background(), "bind closed")
	bind.closeOnce.Do(func() {
		if bind.closedCh != nil {
			close(bind.closedCh)
		}
	})
	return nil
}
```

### DIFF
--- a/proxy/wireguard/bind.go
+++ b/proxy/wireguard/bind.go
@@ -6,6 +6,7 @@ import (
 	"net/netip"
 	"runtime"
 	"strconv"
+	"sync"
 
 	"golang.zx2c4.com/wireguard/conn"
 	"golang.zx2c4.com/wireguard/device"
@@ -30,6 +31,7 @@ type netBind struct {
 	workers   int
 	readQueue chan *netReadInfo
 	closedCh  chan struct{}
+	closeOnce sync.Once
 }
 
 // SetMark implements conn.Bind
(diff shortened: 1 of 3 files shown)


Fangliding:
此类pr需要有真实触发路径(
之前此类问题修过几次 后来没有相关报告了 倾向于问题不大

abolix:
I do have a wireguard connection that can't connect with xray-core but works over the actual wireguard. thats the reason for this PR

Fangliding:
这怎么可能和这个pr相关呢 这是避免重复close造成panic而不是无法连接 你看过这个pr么?

PR #6021, opened 2026-04-27
FakeDNS: Fix race condition and panic when closing

divert-dev:
added missing mutex locks and nil checks so it doesn't crash if methods are called after Close(). 
changed mutex from pointer to value type. also fixed HolderMulti.Close to collect all errors.
added a couple of tests for the close behavior.

### DIFF
--- a/app/dns/fakedns/fake.go
+++ b/app/dns/fakedns/fake.go
@@ -17,7 +17,7 @@ import (
 type Holder struct {
 	domainToIP cache.Lru
 	ipRange    *net.IPNet
-	mu         *sync.Mutex
+	mu         sync.Mutex
 
 	config *FakeDnsPool
 }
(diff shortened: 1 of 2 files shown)


Fangliding:
底下那个LRU有lock的你们为啥对这个玩意race如此耿耿于怀

divert-dev:
yeah i know lru is thread safe, but that's not the issue here.
caught a panic on a very specific edge case: core closes -> late packet arrives -> fakedns tries to process it -> uses already nulled variables like ipRange -> panic: nil pointer dereference.
the lock is just to protect ipRange and other fields from being accessed while Close() is setting them to nil, otherwise it crashes the app.

Fangliding:
你可以反馈panic报告 我会修的（已经弄了） 不要直接糊个AI PR上来

PR #6047, opened 2026-04-30
Add authFailureResponse option to hide SOCKS/HTTP proxy on auth failure

xtclovver:
## Why

This targets active probing from the user's own device.

Some jurisdictions ship on-device scanners (apps or sideloaded probes, mostly it's may be Russian apps) that enumerate listeners on `127.0.0.1` and classify them. Any app with the `INTERNET` permission can do the same with two `connect()` calls:

- Send `\x05\x01\x00`. Any `\x05\xXX` reply identifies a SOCKS5 listener (RFC 1928 §3).
- Send a bare `CONNECT`. A `407 Proxy Authentication Required` with `Proxy-Authenticate: Basic realm="proxy"` identifies an HTTP proxy (RFC 7235 §3.2). The auth scheme and realm leak in the same response.

Password protection on the inbound doesn't help. Both responses are mandated by the RFCs and emitted before any auth exchange. The classification completes with zero traffic ever proxied.

The standard mitigation (binding to a UNIX socket or sitting behind a reverse proxy) isn't available on mobile. Xray clients like Happ and v2rayNG listen on local TCP because that is the only IPC the platform exposes to system VPN service consumers. Hardening the listener itself is the only option.

## What

A new optional `authFailureResponse` field on SOCKS and HTTP inbound settings controls what the server returns when the client offers no acceptable auth method or supplies wrong credentials:

| value | behavior |
|---|---|
| `reject` (default) | current RFC 1928 / RFC 7235 responses, full backwards compatibility |
| `drop` | close the connection without writing anything |
| `http400` | reply with `HTTP/1.1 400 Bad Request\r\nServer: nginx\r\n…` so the port looks like a generic web service |

The setting is honored at both auth-failure points in SOCKS (no matching method, bad password) and on the HTTP path. It also propagates into the embedded HTTP fallback that the SOCKS inbound uses for non-SOCKS traffic, so a probe sending an HTTP request to a SOCKS port can't bypass the camouflage.

`drop` and `http400` are explicit stealth hardening modes. For SOCKS5 password auth, they intentionally avoid sending the normal `0x05 0x02` method-selection byte before credential validation, because that byte alone fingerprints the listener as SOCKS5. As a result, these modes are not fully compatible with clients that wait for method selection before sending username/password. Such clients should keep using the default `reject` mode. Clients that need `drop` or `http400` with SOCKS5 password auth must send credentials preemptively/optimistically.

## Example config

```json
{
  "protocol": "socks",
  "port": 1080,
  "settings": {
    "auth": "password",
    "accounts": [{"user": "u", "pass": "p"}],
    "authFailureResponse": "http400"
  }
}
```

`authFailureResponse` accepts `"reject"`, `"drop"`, or `"http400"`. Omitted or empty means `reject`. Unknown values are rejected at config load with an explicit error.

## Backwards compatibility

Default `reject` emits the same RFC byte sequences the previous code did. Existing configs and clients behave identically.

The non-default `drop` and `http400` modes are opt-in hardening modes and may require client support for preemptive SOCKS5 username/password auth. They should not be used for generic compatibility-sensitive SOCKS5 password deployments unless the client behavior is known.

## Changes

- `proxy/socks/config.proto`, `proxy/http/config.proto`: new `AuthFailureBehavior` enum and field on `ServerConfig` (regenerated `.pb.go`).
- `infra/conf/socks.go`, `infra/conf/http.go`: JSON field `authFailureResponse` with string→enum mapping. Unknown values return a config error.
- `proxy/socks/protocol.go`: `writeSocks5AuthFailure` dispatcher replaces unconditional failure writes (`0x05 0xFF`, `0x01 0xFF`) and SOCKS4 auth-required rejection.
- `proxy/http/server.go`: `writeHTTPAuthFailure` dispatcher replaces the unconditional 407 write.
- `proxy/socks/server.go`: forwards `AuthFailureBehavior` into the embedded HTTP fallback `ServerConfig` through an explicit enum mapper. Does not rely on numeric equality between the two protobuf enums.

## Test plan

- `go build ./...`
- `go vet ./proxy/socks/... ./proxy/http/...`
- Existing tests in `proxy/socks` still pass.
- New unit tests for `writeSocks5AuthFailure` and `writeHTTPAuthFailure` covering all three behaviors.
- New unit tests for SOCKS4 auth-required responses and SOCKS5 bad-password behavior in stealth modes.
- Integration smoke test on a real binary, all three behaviors, both protocols:
  - `reject` produces the original RFC bytes.
  - `drop` closes the TCP connection with zero bytes written.
  - `http400` produces a generic 400 Bad Request with no proxy headers.
- Happy path verified:
  - `reject` keeps the standard SOCKS5 password flow.
  - `drop` / `http400` work with preemptive SOCKS5 username/password auth.
  - HTTP proxy auth with valid credentials still returns `HTTP/1.1 200 Connection established`.
- HTTP fallback path on the SOCKS inbound respects the setting. A non-SOCKS request to a SOCKS port no longer leaks 407 when `http400` or `drop` is configured.

### DIFF
--- a/infra/conf/http.go
+++ b/infra/conf/http.go
@@ -23,15 +23,27 @@ func (v *HTTPAccount) Build() *http.Account {
 }
 
 type HTTPServerConfig struct {
-	Accounts    []*HTTPAccount `json:"accounts"`
-	Transparent bool           `json:"allowTransparent"`
-	UserLevel   uint32         `json:"userLevel"`
+	Accounts            []*HTTPAccount `json:"accounts"`
+	Transparent         bool           `json:"allowTransparent"`
+	UserLevel           uint32         `json:"userLevel"`
+	AuthFailureResponse string         `json:"authFailureResponse"`
 }
 
 func (c *HTTPServerConfig) Build() (proto.Message, error) {
 	config := &http.ServerConfig{
 		AllowTransparent: c.Transparent,
 		UserLevel:        c.UserLevel,
+	}
+
+	switch c.AuthFailureResponse {
+	case "", "reject":
+		config.AuthFailureBehavior = http.AuthFailureBehavior_REJECT
+	case "drop":
+		config.AuthFailureBehavior = http.AuthFailureBehavior_DROP
+	case "http400":
+		config.AuthFailureBehavior = http.AuthFailureBehavior_HTTP400
+	default:
+		return nil, errors.New(`unknown http "authFailureResponse": `, c.AuthFailureResponse, ` (expected "reject", "drop", or "http400")`)
 	}
 
 	if len(c.Accounts) > 0 {
--- a/infra/conf/socks.go
+++ b/infra/conf/socks.go
@@ -28,11 +28,12 @@ const (
 )
 
 type SocksServerConfig struct {
-	AuthMethod string          `json:"auth"`
-	Accounts   []*SocksAccount `json:"accounts"`
-	UDP        bool            `json:"udp"`
-	Host       *Address        `json:"ip"`
-	UserLevel  uint32          `json:"userLevel"`
+	AuthMethod          string          `json:"auth"`
+	Accounts            []*SocksAccount `json:"accounts"`
+	UDP                 bool            `json:"udp"`
+	Host                *Address        `json:"ip"`
+	UserLevel           uint32          `json:"userLevel"`
+	AuthFailureResponse string          `json:"authFailureResponse"`
 }
 
 func (v *SocksServerConfig) Build() (proto.Message, error) {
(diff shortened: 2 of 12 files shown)


RPRX:
虽然疑似 AI 但似乎还行，@Fangliding 你看看

Fangliding:
完全没必要 还破坏了RFC 很多客户端遇到这个直接不认识了 都这一步非标客户端了 要安全本地改个vless/trojan倒数据都行
我就说 本地回环地址 127.0.0.0/8 2^24*65535 的空间 随便找个地方落都能解决问题 为什么还要死抠这个验证 socks5防主动探测都来了

PR #6055, opened 2026-05-01
Config: Parallel `for` for inbounds' `clients`

kastov:
**Parallelize `VLessInboundConfig.Build()`** — per-client work (`json.Unmarshal`, UUID parse, `proto.Marshal` into `TypedMessage`) is independent and now runs in `runtime.GOMAXPROCS(0)` worker goroutines via `errgroup`. Output order preserved (each worker writes to its own slice index).

On a 5M-user config (single VLESS inbound):

| Before | After |
|-------:|------:|
| 60+ seconds | **~22 seconds** |

Together with #6053 startup time drops even further on remote configs.

### DIFF
--- a/common/task/parallel.go
+++ b/common/task/parallel.go
@@ -1,0 +1,43 @@
+package task
+
+import (
+	"runtime"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// ParallelForN runs fn(0..n-1) in parallel across runtime.GOMAXPROCS(0) worker
+// goroutines. Indices are partitioned into contiguous chunks so the number of
+// spawned goroutines stays bounded regardless of n.
+//
+// fn must be safe to call concurrently from different goroutines (each call
+// receives its own unique index). Output collected by writing to indexed slots
+// in a pre-allocated slice is a common safe pattern.
+//
+// Returns the first non-nil error reported by fn; other workers may still be
+// finishing briefly afterwards.
+func ParallelForN(n int, fn func(i int) error) error {
+	if n <= 0 {
+		return nil
+	}
+	workers := max(runtime.GOMAXPROCS(0), 1)
+	workers = min(workers, n)
+	chunk := (n + workers - 1) / workers
+	var eg errgroup.Group
+	for w := range workers {
+		start := w * chunk
+		end := min(start+chunk, n)
+		if start >= end {
+			break
+		}
+		eg.Go(func() error {
+			for i := start; i < end; i++ {
+				if err := fn(i); err != nil {
+					return err
+				}
+			}
+			return nil
+		})
+	}
+	return eg.Wait()
+}
--- a/common/task/parallel_test.go
+++ b/common/task/parallel_test.go
@@ -1,0 +1,50 @@
+package task_test
+
+import (
+	"errors"
+	"sync/atomic"
+	"testing"
+
+	"github.com/xtls/xray-core/common"
+	. "github.com/xtls/xray-core/common/task"
+)
+
+func TestParallelForN_Empty(t *testing.T) {
+	called := false
+	err := ParallelForN(0, func(i int) error {
+		called = true
+		return nil
+	})
+	common.Must(err)
+	if called {
+		t.Fatal("fn should not be called when n=0")
+	}
+}
+
+func TestParallelForN_AllIndicesCovered(t *testing.T) {
+	const N = 10000
+	var seen [N]int32
+	err := ParallelForN(N, func(i int) error {
+		atomic.AddInt32(&seen[i], 1)
+		return nil
+	})
+	common.Must(err)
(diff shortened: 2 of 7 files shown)


Fangliding:
6053还能考虑 这个就太奇怪了

kastov:
What?

Fangliding:
你们就不能考虑用API往core里加或者fork一份调Validator接口再接你们的user数据库么 什么都要塞到核心里来

kastov:
So, in your opinion, adding 5 million users via the API will be faster than with the initial load from config? That's complete nonsense.

22 seconds is an excellent result. In the current release version, it takes an eternity, over a minute.

RPRX:
~~虽然收了赞助不过客观来说还好~~，VLESS inbound 当初用 sync.Map 比较偷懒，读多写少的话直接用读写锁确实性能更好（[否](https://github.com/XTLS/Xray-core/pull/6055#issuecomment-4362212097)

build 配置文件那个是把单核 for 改成多核，不过只改 VLESS 的话有点脏，应当把这个能力抽象为一个函数然后每个入站都改

RPRX:
那个函数可以放 common utils 啥的，以后任何有这种 for 的地方都可以直接调用以优化性能

Fangliding:
这么改非常奇怪 就挤出了那么十几秒几十秒的启动时间（相比改json解析器） 就为了那 "5 million users" 完全超出了本来的设计范畴 管理这么多用户本来就弄不了要上数据库的 这还在这力大砖飞

RPRX:
~~没事，说不定 millions of users 的俄罗斯机场可以成为新的赞助商，不过俄罗斯一亿多人，一个机场有五百万用户太有实力了~~

~~每月几刀来算的话入账都几千万刀了，搞得我都心动了，所以是哪个机场，有没有考虑每月分亿点给 Xray-core~~

Fangliding:
哎 赞助
要加就加吧 别搞别的入站

RPRX:
这是一项通用能力啊，直接 for 确实是单核，现在能用多核来并行处理了，别的入站改一下也就套个函数而已

kastov:
Function placed in /common/task

Trojan, VMESS is also updated

RPRX:
@kastov Hy 和 SS 也改一下吧

kastov:
@RPRX added for SS, SS22 (without relay), Hysteria

RPRX:
@kastov 我没看代码，WireGuard 是不是加不了，依稀记得它传了全部配置给外部库来着？

kastov:
~yes, but honestly I don't think it's necessary. As far as I know, almost nobody uses the WireGuard inbound for a large number of users due to various limitations and issues. As i remember, adding/removing Wireguard users via the gRPC API isn't currently possible anyway.~

kastov:
https://github.com/XTLS/Xray-core/blob/7ab0a3ccb7a53c79669b9cadfdb53a8954ec427d/infra/conf/wireguard.go#L80-L90

victor-bertollo:
well, I'd like to ask you about  replacing sync.Map with map + RWMutex. is this a good idea? do we write as often as read or more often? xray-core is not only used in commercial segment. have you tested the performance of creating a new connection?
I used to want to replace sync.Map, but then I realized that it was almost impossible. maybe in the future...

RPRX:
坏了，读极多的情况下貌似 sync.Map 性能更好，因为它是两份 map，并发读的时候就没锁

不过这里改成一个锁管两个 map 的话。。。我再看看，我留那个注释的意思应该是 email 可能写多读少，所以可以用读写锁

@kastov 你能测一下 benchmark 吗

victor-bertollo:
I also looked at the golang source code, and it seems that it would be very impractical to abandon sync.Map even with frequent write operations

https://github.com/golang/go/blob/f1bc06b98d79c91a0b073098862f1465dbc69776/src/internal/sync/hashtriemap.go#L14-L17

kastov:
@RPRX you are right. RWMutex gives us a ~30% startup speedup and dramatically fewer allocations on bulk insert, but sync.Map is faster on concurrent reads.


Benchmark | map+RWMutex | sync.Map | Δ
-- | -- | -- | --
GetByUUID, 100k users, parallel | 67.3 ns/op | 1.96 ns/op | sync.Map 34× faster
GetByUUID, 1M users, parallel | 68.1 ns/op | 3.89 ns/op | 17× faster
GetByEmail, 100k users, parallel | 65.3 ns/op | 2.82 ns/op | 23× faster
GetByEmail, 1M users, parallel | 66.0 ns/op | 5.80 ns/op | 11× faster

Let me revert validator.go to initial state for now.

RPRX:
说起来我想把 inbounds 这些 `clients` 都更名成 `users` 来着，~~叫前者单独看的话总觉得是客户端侧配置，怪怪的~~

PR #6062, opened 2026-05-02
Tunnel inbound: Fix panic when listening on UDS for Xray's internal services (e.g. API)

MakostaDev:
fixed a panic that occurs when using UNIX socket as the listen address in dokodemo-door inbound configuration:
```
panic: strconv.Atoi: parsing "": invalid syntax

goroutine 51 [running]:
github.com/xtls/xray-core/common.Must(...)
        github.com/xtls/xray-core/common/common.go:22
github.com/xtls/xray-core/common.Must2[...](...)
        github.com/xtls/xray-core/common/common.go:30
github.com/xtls/xray-core/proxy/dokodemo.(*DokodemoDoor).Process(0x3dc6317d7140, {0x16255b0, 0x3dc6318780c0}, 0x4, {0x7fa901fb91a0, 0x3dc6314b20b8}, {0x162a220, 0x3dc63159b710})
        github.com/xtls/xray-core/proxy/dokodemo/dokodemo.go:99 +0x11f1
github.com/xtls/xray-core/app/proxyman/inbound.(*dsWorker).callback(0x3dc631874b60, {0x7fa901fb91a0, 0x3dc6314b20b8})
        github.com/xtls/xray-core/app/proxyman/inbound/worker.go:510 +0x484
created by github.com/xtls/xray-core/app/proxyman/inbound.(*dsWorker).Start.func1 in goroutine 50
        github.com/xtls/xray-core/app/proxyman/inbound/worker.go:530 +0x7a
```
when using UNIX socket, the net.SplitHostPort() function returns an empty string for the port variable. at line 99, the code attempts to convert this empty string to an integer using strconv.Atoi(port), which causes the panic.
my example configuration:
```json
{
    "protocol": "dokodemo-door",
    "listen": "/my/path/api.sock,0666",
    "tag": "api"
}
```

### DIFF
--- a/app/commander/commander.go
+++ b/app/commander/commander.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"net"
 	"sync"
+    "strings"
 
 	"github.com/xtls/xray-core/common"
 	"github.com/xtls/xray-core/common/errors"
 	"github.com/xtls/xray-core/common/signal/done"
 	core "github.com/xtls/xray-core/core"
 	"github.com/xtls/xray-core/features/outbound"
+    "github.com/xtls/xray-core/transport/internet"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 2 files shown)


RPRX:
~~那你这流量准备发到哪，靠 Freedom 出站的 redirect 吗~~

MakostaDev:
the traffic is intended to be processed by the internal gRPC API
here is my configuration:
```json
{
    "inbounds": [
        {
            "protocol": "dokodemo-door",
            "listen": "/my/path/api.sock,0666",
            "tag": "api"
        }
    ],
    "api": {
        "tag": "api",
        "services": [
            "LoggerService",
            "RoutingService",
            "HandlerService"
        ]
    },
    "routing": {
        "rules": [
            {
                "inboundTag": ["api"],
                "outboundTag": "api"
            }
        ]
    }
}
```
I tested it and еverything works perfectly on my end with this configuration and the fix

MakostaDev:
While exploring the XRAY-core internals, I noticed that the API listener currently uses the standard net.Listen from golang. Although net.Listen itself supports Unix sockets, I found it more appropriate to use XRAY's internal listener wrapper instead.
I've updated the implementation to use XRAY's internal listener wrapper instead. This wrapper is more versatile as it handles both TCP and UNIX sockets automatically, including file mode settings. This change makes the API listening logic consistent with the rest of the project and enables native UNIX socket support for the API block

RPRX:
> I've updated the implementation to use XRAY's internal listener wrapper instead

处理 @@ 的代码也在里面对吧

RPRX:
看到了，有包含 @@

MakostaDev:
> > I've updated the implementation to use XRAY's internal listener wrapper instead
> 
> 处理 @@ 的代码也在里面对吧

Yes. Btw, unlike net.Listen, this wrapper has supports @@

PR #6087, opened 2026-05-07
Fix net.Conn memory leak in case of Handshake error

excel23:
On error, Handshake returns (nil, err) → conn becomes nil.
defer conn.Close() (declared above) attempts to close nil → panic.
The original connection (raw net.Conn) is lost but not closed → descriptor and memory leak.

What modification does:
The original conn is preserved before calling Handshake.
On error, the original connection is explicitly closed → no leak.
defer conn.Close() (the outer one) will no longer panic because conn did not become nil (the error returns before reassignment).

### DIFF
--- a/proxy/vless/outbound/outbound.go
+++ b/proxy/vless/outbound/outbound.go
@@ -209,10 +209,12 @@ func (h *Handler) Process(ctx context.Context, link *transport.Link, dialer inte
 	errors.LogInfo(ctx, "tunneling request to ", target, " via ", rec.Destination.NetAddr())
 
 	if h.encryption != nil {
-		var err error
-		if conn, err = h.encryption.Handshake(conn); err != nil {
+		newConn, err := h.encryption.Handshake(conn)
+		if err != nil {
+			conn.Close()
 			return errors.New("ML-KEM-768 handshake failed").Base(err).AtInfo()
 		}
+		conn = newConn
 	}
 
 	command := protocol.RequestCommandTCP


Fangliding:
这个pattern在core中被大量使用 不会有问题的 defer 要执行的函数从执行到那一行起就固定了

excel23:
After this modification, the memory leak that I analysed with pprof wasn't repeated. That was the error, I believe.
Could you look up these heap dumps?
[xray_mem.zip](https://github.com/user-attachments/files/27492743/xray_mem.zip)

Fangliding:
提供一下goroutine的 pprof

excel23:
After the error repeated, I can say that the problem occurs when the system is configured to use VLESS Reality with mlkem768 encryption. If the connection is stable, everything works fine. But as soon as you turn off the connection, the memory leak appears. Here are the heap and goroutine dumps.
[xray_base.zip](https://github.com/user-attachments/files/27502491/xray_base.zip)
After modification leak is reduced, but not gone:
[xray_mod.zip](https://github.com/user-attachments/files/27502285/xray_mod.zip)

Fangliding:
try https://github.com/XTLS/Xray-core/commit/1bae46b1c065eeb6b39d50c63697f56993184c5f

excel23:
Yes, it fixes the heap and goroutine leak, even without my modifications. 
I still think there is a minor bug in Handshake error handling in outbound.go, but at least it doesn't cause an uncontrollable leak.

RPRX:
> I still think there is a minor bug in Handshake error handling in outbound.go, but at least it doesn't cause an uncontrollable leak.

你试下使用其它传输层（比如 WS）时有没有问题

AndyChiang888:
> Yes, it fixes the heap and goroutine leak, even without my modifications. I still think there is a minor bug in Handshake error handling in outbound.go, but at least it doesn't cause an uncontrollable leak.

I think this PR is necessary.

PR #6102, opened 2026-05-09
fix: kill active connections when user is removed via gRPC

nightcrawler42:
## Summary

When a user is removed through the gRPC API (`AlterInbound` → `RemoveUser`), existing connections survive because user validation only happens during the initial handshake. After authentication, the connection runs independently with a captured `User` pointer — the validator is never consulted again.

This is a real-world problem for panel operators (Marzban, 3x-ui, etc.) who expect disabled/removed users to be disconnected immediately. Currently, the only way to force-disconnect is to restart the entire xray process, which drops **all** connections.

## Changes

- **New `ConnTracker`** (`proxy/conntrack.go`): per-user connection tracking using atomic sequence IDs and `sync.Mutex`. Stores both `context.CancelFunc` and `io.Closer` (the underlying `net.Conn`) for each tracked connection.

- **`KillAll(email)`**: cancels contexts AND closes TCP connections. Context cancellation alone is insufficient because `buf.Copy` (the core data relay loop in `common/buf/copy.go`) has no context awareness — it only stops on read/write errors. Closing the `net.Conn` forces an immediate TCP RST.

- **All 5 protocol handlers updated**: VLESS, VMess, Trojan, Shadowsocks, Shadowsocks 2022
  - `Track()` called after user authentication in `Process()`
  - `KillAll()` called in `RemoveUser()` after successful removal

- **Shadowsocks 2022 fixes**:
  - Bounds check for user index in `NewConnection`/`NewPacketConnection` (prevents panic when user removed during active handoff)
  - Error propagation for `UpdateUsersWithPasswords` in `RemoveUser` (was silently discarded)

## Why not just cancel the context?

`buf.Copy` loops forever on `reader.ReadMultiBuffer()` with no `ctx.Done()` check. Long-lived streams (e.g. Psiphon tunnels, persistent WebSocket connections) survive context cancellation indefinitely. Force-closing the TCP connection is the only reliable way to break all in-flight relay loops.

## Testing

- Builds cleanly: `CGO_ENABLED=0 go build ./...`
- Tested in production with Marzban panel + multiple nodes
- Verified with Psiphon (long-lived tunneled connections) — connections now terminate within seconds of user removal

### DIFF
--- a/proxy/conntrack.go
+++ b/proxy/conntrack.go
@@ -1,0 +1,83 @@
+package proxy
+
+import (
+	"context"
+	"io"
+	"strings"
+	"sync"
+	"sync/atomic"
+)
+
+type trackedConn struct {
+	cancel context.CancelFunc
+	closer io.Closer // the actual network connection
+}
+
+// ConnTracker tracks active connections per user email and supports
+// killing all connections for a given user on removal.
+// It both cancels the context AND closes the underlying connection
+// to ensure immediate termination even for long-lived streams.
+type ConnTracker struct {
+	mu    sync.Mutex
+	conns map[string]map[uint64]*trackedConn
+	seq   atomic.Uint64
+}
+
+func NewConnTracker() *ConnTracker {
+	return &ConnTracker{
+		conns: make(map[string]map[uint64]*trackedConn),
+	}
+}
+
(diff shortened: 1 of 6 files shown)


Fangliding:
dup #5844

PR #6106, opened 2026-05-10
Burst observatory: Fix time compare, cancel pending ping on instance close or new schedule started

Jolymmiles:
- Compare Interval to 10s
- Serialize check rounds
- Cancel pending pings on StopScheduler


Jolymmiles:
i think we should rerun checks

Jolymmiles:
changed pow 0.5 to sqrt for leastload with cost

Jolymmiles:
@Fangliding could you look pls?

Fangliding:
这里有cache的根本中不了几次 不要再乱找了
有BUG你以后还是开issue我来看就是了

Jolymmiles:
> 这里有cache的根本中不了几次 不要再乱找了 有BUG你以后还是开issue我来看就是了

I make bench tests for pow 0.5 and sqrt and got some benefits, some time ago I catch some issue with clients side where was used least load and this single point what my eyes catch. I guess we can change this mechanism or create new, something like priority load. In Russian we have providers of servers which have limited traffic and unmetered traffic. At the current time when we want to send traffic on working outbound on client via priority, we should use loopback with balancer fallback on my look this not best solution for this issue

RPRX:
@Fangliding 这个 PR 现在是什么状态

Fangliding:
一些小修复 可以用了

RPRX:
@Fangliding 这个 PR 的标题是否需要修改，需要的话你直接改一下吧

orkhv:
After PR #6106, we lost the separate initial Check() on startup. I understand that this removed race conditions, but on a cold start the first observatory result may now only appear within a window of up to interval × sampling. For example, with interval: 60s and sampling: 1, the first ping may be delayed by up to a minute (+ timeout).

Until data is available, leastLoad finds no candidates and go to fallbackTag, or if no fallback is set cannot select an outbound properly and you will need to wait a few minutes for the initial connection. For configs where the balancer is main routing rule for traffic, this breaks a smooth core start.

I'm not a developer, so I can't propose a concrete fix, but this seems like an important

RPRX:
@Fangliding 你看看

Fangliding:
我不是很理解 在处理循环中首次尝试不会等待计时脉冲而是直接发出 这让我觉得这个完全就是复制了一遍的初始探测完全是多余的于是删掉了 除非其它部分依赖两次初始探测(怎么会呢)
image

orkhv:
You're right: the first doCheck in the loop starts immediately on startup (it does not wait for the first ticker pulse). The difference is when the check result becomes available.

The removed Check() was an immediate ping (doCheck(h.ctx, tags, 0, 1) — duration=0, rounds=1), with results typically in ~timeout — practically immediate and guaranteed.

The remaining call is doCheck(subCtx, tags, interval, SamplingCount), where interval = Interval × SamplingCount. Pings are scheduled with a random delay in the range 0…interval.

Now on start, we use only doCheck, which does not guarantee that a result will be obtained immediately. Maybe Check() doesn't need to be restored as a duplicate path. the first scheduled round could simply use duration=0 and without intervals so start gets a result fast, without reintroducing overlapping probes.

Fangliding:
看起来我们别无选择只好把这个独特的check请求加回来了

Meo597:
这里的行为就一直很迷，观测结果保留的不够多
故障转移、恢复的也慢

PR #6129, opened 2026-05-14
VLESS: randomize pre-connections count and intervals

Alaxay8:
This PR addresses several TODOs in the VLESS outbound handler related to background pre-connections.

Currently, pre-connections follow a very rigid and predictable pattern (fixed number of threads, fixed 200ms intervals, fixed 2-minute expiration). This creates a distinct network signature that can be used for traffic analysis and fingerprinting.

#### Changes:
- **Randomized worker count**: The number of spawned goroutines is now randomized in the range `[testpre/2 + 1, testpre]`.
- **Randomized sleep intervals**: Replaced fixed 200ms sleep with a random delay between **100ms and 300ms**.
- **Randomized expiration**: Pre-connections now expire at random intervals between **90s and 150s** (previously fixed at 120s).

#### Technical Details:
- Integrated `github.com/xtls/xray-core/common/dice` for randomization.
- Added jitter to the core pre-connection loop in `proxy/vless/outbound/outbound.go`.

#### Verification:
Verified using a local test suite monitoring dial attempts. With `testpre = 5`, the total number of connection attempts per second showed significant variation (ranging from 16 to 23 across different runs), confirming that the fixed pattern has been successfully broken.

### DIFF
--- a/proxy/vless/outbound/outbound.go
+++ b/proxy/vless/outbound/outbound.go
@@ -18,6 +18,7 @@ import (
 	"github.com/xtls/xray-core/common"
 	"github.com/xtls/xray-core/common/buf"
 	xctx "github.com/xtls/xray-core/common/ctx"
+	"github.com/xtls/xray-core/common/dice"
 	"github.com/xtls/xray-core/common/errors"
 	"github.com/xtls/xray-core/common/mux"
 	"github.com/xtls/xray-core/common/net"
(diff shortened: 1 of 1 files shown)


Fangliding:
这只是一个测试字段 TODO只是标记一下当真正要上的时候会给出完全可以自定义的字段

PR #6166, opened 2026-05-19
Config: avoid REALITY non-443 warning for Unix socket listen

vemneyy:
Avoid printing the REALITY non-443 port warning when an inbound listens on a UNIX Socket


RPRX:
https://github.com/XTLS/BBS/issues/21#issuecomment-4109400301

PR #6169, opened 2026-05-20
XHTTP: Add zstd and lz4 compression

Deneby67:
Adds optional XHTTP payload compression with `zstd` and `lz4`.

  - Adds `compression` config under `xhttpSettings` / `splithttpSettings`
  - Supports `type`, `bufferSize`, and `compressLevel`
  - Compresses only HTTP body / stream body payload data
  - Keeps headers, cookies, padding, session, and seq metadata uncompressed for stealth/compatibility
  - Rejects `packet-up` compression with `uplinkDataPlacement: "header"` or `"cookie"`

  ## Config Example

  ```json
  {
    "xhttpSettings": {
      "compression": {
        "type": "zstd",
        "bufferSize": 262144,
        "compressLevel": 3
      }
    }
  }
```
  Supported type values:

  - none
  - zstd
  - lz4

  ## Notes

  Compression is symmetric and not negotiated on the wire. Client and server must use matching compression settings.

  For packet-up, compression applies only to body payloads. Header/cookie payload placement remains uncompressed and is rejected when compression is enabled to avoid high-entropy payload data appearing in headers or cookies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,9 +9,11 @@ require (
 	github.com/golang/mock v1.7.0-rc.1
 	github.com/google/go-cmp v0.7.0
 	github.com/gorilla/websocket v1.5.3
+	github.com/klauspost/compress v1.17.4
 	github.com/klauspost/cpuid/v2 v2.3.0
 	github.com/miekg/dns v1.1.72
 	github.com/pelletier/go-toml v1.9.5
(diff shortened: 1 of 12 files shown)


Fangliding:
这里面加压缩做什么 都是高熵比特流。。。

PR #6180, opened 2026-05-23
Geodata: Support hash files for asset updates

kirilllivanov:
# Geodata Hash File Support

## Purpose

This change adds optional hash-file support to the scheduled geodata updater. It lets Xray verify downloaded geodata assets before replacing local files, and skip downloading large geofiles when the remote hash shows that the local asset is already up to date.

The feature is useful on mobile devices and metered networks, where avoiding unnecessary `geoip.dat` and `geosite.dat` downloads saves bandwidth and battery when only a small checksum file needs to be fetched.

## What Was Added

- Added geodata asset hash configuration fields: `hash_url`, `hash_file`, and `hash_type` in protobuf, exposed as `hashUrl`, `hashFile`, and `hashType` in user-facing config.
- Added hash verification before downloaded geodata files are swapped into the asset directory.
- Added hash-first update logic:
  - download the remote hash file first;
  - compare it with the local hash file when available;
  - skip geofile download when the digest is unchanged;
  - download the geofile immediately when the local hash file exists and the digest differs;
  - verify the local geofile only when the local hash file is missing or unreadable.
- Added support for missing local hash files on first run. The configured `hashFile` path must be valid, but the file itself may not exist yet.
- Added default hash type handling: when `hashType` is omitted, `sha256` is used.
- Added tests for config parsing, hash verification, skipped downloads, missing hash files, and filesystem path validation.

## Update Logic

For assets without hash settings, behavior stays the same:

1. Download the geodata file to a temporary file.
2. Swap it into the asset directory.
3. Reload geodata.
4. Roll back if reload fails.

For assets with hash settings:

1. Download `hashUrl` to a temporary hash file.
2. Parse the expected digest from that hash file.
3. Compare the downloaded digest with the local `hashFile` when possible.
   - If the local `hashFile` exists and contains the same digest, skip both geofile download and reload.
   - If the local `hashFile` exists but contains a different digest, download the geofile from `url`.
   - If the local `hashFile` is missing or unreadable, verify the current local geofile against the downloaded hash.
     - If the current local geofile already matches the downloaded hash, only stage the new hash file.
     - If the current local geofile is outdated or fails verification, download the geofile from `url`.
4. When a geofile is downloaded, verify it against the downloaded hash file.
   - If verification fails, abort the update, remove temporary files, and keep the existing local files unchanged.
5. Atomically swap staged files into place using the existing transaction flow.
6. Reload geodata and roll back swapped files if reload fails.

## Hash Types

`hashType` is optional. The default value is:

```json
"hashType": "sha256"
```

Supported hash types are based on cryptographically strong hash implementations from the Go standard library:

- `sha224`
- `sha256`
- `sha384`
- `sha512`
- `sha512/224`
- `sha512/256`
- `sha3-224`
- `sha3-256`
- `sha3-384`
- `sha3-512`

Hash type names are normalized, so these forms are accepted:

- `sha512/256`
- `sha-512-256`
- `sha512_256`
- `sha512256`

Weak or non-standard-library hash types such as `md5`, `sha1`, `ripemd160`, and `blake2*` are rejected.

### Existing Behavior Without Hash Verification

```json
{
  "geodata": {
    "cron": "0 4 * * *",
    "outbound": "proxy",
    "assets": [
      {
        "url": "https://example.com/geoip.dat",
        "file": "geoip.dat"
      },
      {
        "url": "https://example.com/geosite.dat",
        "file": "geosite.dat"
      }
    ]
  }
}
```

### Hash Verification With Default SHA-256

`hashType` can be omitted. It defaults to `sha256`.

```json
{
  "geodata": {
    "cron": "0 4 * * *",
    "outbound": "proxy",
    "assets": [
      {
        "url": "https://example.com/geoip.dat",
        "file": "geoip.dat",
        "hashUrl": "https://example.com/geoip.dat.sha256sum",
        "hashFile": "geoip.dat.sha256sum"
      }
    ]
  }
}
```

### Explicit Hash Type

```json
{
  "geodata": {
    "cron": "0 4 * * *",
    "outbound": "proxy",
    "assets": [
      {
        "url": "https://example.com/geosite.dat",
        "file": "geosite.dat",
        "hashUrl": "https://example.com/geosite.dat.sha512sum",
        "hashFile": "geosite.dat.sha512sum",
        "hashType": "sha512"
      }
    ]
  }
}
```

## Compatibility Notes

- Existing geodata updater configurations continue to work unchanged.
- `file` is still required for every asset because it defines the local asset path.
- `hashUrl` and `hashFile` must be provided together.
- `hashType` is optional and defaults to `sha256`.
- The local `hashFile` may be absent on first run; it will be created after a successful update.
- Hash files are stored in the same asset directory rules as geodata files, and invalid paths outside the asset directory are rejected.
- The downloaded hash file may contain a plain hex digest or common checksum-file formats such as:

```text
<hex-digest>
<hex-digest>  geoip.dat
SHA256 (geoip.dat) = <hex-digest>
```

- If the checksum line includes a filename, it must match the configured geodata `file`. This avoids accidentally using a digest for another file when a checksum file contains multiple entries.

### DIFF
--- a/app/geodata/config.pb.go
+++ b/app/geodata/config.pb.go
@@ -25,6 +25,9 @@ type Asset struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	Url           string                 `protobuf:"bytes,1,opt,name=url,proto3" json:"url,omitempty"`
 	File          string                 `protobuf:"bytes,2,opt,name=file,proto3" json:"file,omitempty"`
+	HashUrl       string                 `protobuf:"bytes,3,opt,name=hash_url,json=hashUrl,proto3" json:"hash_url,omitempty"`
+	HashFile      string                 `protobuf:"bytes,4,opt,name=hash_file,json=hashFile,proto3" json:"hash_file,omitempty"`
+	HashType      string                 `protobuf:"bytes,5,opt,name=hash_type,json=hashType,proto3" json:"hash_type,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -69,6 +72,27 @@ func (x *Asset) GetUrl() string {
 func (x *Asset) GetFile() string {
 	if x != nil {
 		return x.File
+	}
+	return ""
+}
+
+func (x *Asset) GetHashUrl() string {
+	if x != nil {
+		return x.HashUrl
+	}
+	return ""
+}
+
+func (x *Asset) GetHashFile() string {
+	if x != nil {
+		return x.HashFile
+	}
+	return ""
+}
+
+func (x *Asset) GetHashType() string {
+	if x != nil {
+		return x.HashType
 	}
 	return ""
 }
@@ -137,10 +161,13 @@ var File_app_geodata_config_proto protoreflect.FileDescriptor
 
 const file_app_geodata_config_proto_rawDesc = "" +
 	"\n" +
-	"\x18app/geodata/config.proto\x12\x10xray.app.geodata\"-\n" +
+	"\x18app/geodata/config.proto\x12\x10xray.app.geodata\"\x82\x01\n" +
 	"\x05Asset\x12\x10\n" +
 	"\x03url\x18\x01 \x01(\tR\x03url\x12\x12\n" +
-	"\x04file\x18\x02 \x01(\tR\x04file\"i\n" +
+	"\x04file\x18\x02 \x01(\tR\x04file\x12\x19\n" +
+	"\bhash_url\x18\x03 \x01(\tR\ahashUrl\x12\x1b\n" +
+	"\thash_file\x18\x04 \x01(\tR\bhashFile\x12\x1b\n" +
+	"\thash_type\x18\x05 \x01(\tR\bhashType\"i\n" +
 	"\x06Config\x12\x12\n" +
 	"\x04cron\x18\x01 \x01(\tR\x04cron\x12\x1a\n" +
 	"\boutbound\x18\x02 \x01(\tR\boutbound\x12/\n" +
--- a/app/geodata/config.proto
+++ b/app/geodata/config.proto
@@ -10,6 +10,12 @@ message Asset {
   string url = 1;
 
   string file = 2;
+
+  string hash_url = 3;
+
+  string hash_file = 4;
+
+  string hash_type = 5;
 }
 
 message Config {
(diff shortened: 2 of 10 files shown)


RPRX:
感觉没必要吧 @Meo597

Meo597:
完全没必要不该这么做

如果想跳过下载应该检查 http header 这才是正确路线
如果说检验文件损坏，上游有可能发布坏的，哈希解决不了问题，不如现在这样失败了全会滚

目前之所以没去检查更新跳过下载，因为这玩意每天都发布，至少 ip 是这样

RPRX:
File-hash 这种东西更适合当你需要通过无法完全信任的下载源比如 HTTP/BT/CDN/Mirror 等途径下载文件后的防篡改校验

下载超大文件时可能这些途径更快，下载后你拿官方 hash 来校验文件没被夹带私货，而 geodata 顶多几十 MB，所以

kirilllivanov:
> 完全没必要不该这么做
> 
> 如果想跳过下载应该检查 http header 这才是正确路线 如果说检验文件损坏，上游有可能发布坏的，哈希解决不了问题，不如现在这样失败了全会滚
> 
> 目前之所以没去检查更新跳过下载，因为这玩意每天都发布，至少 ip 是这样

From the HTTP standard point of view, there is no generic way to prove that two resources have identical file content before downloading the body. Standard validators such as ETag and Last-Modified identify the server’s representation/version of a resource, not necessarily a content hash.

For example, compare these two GitHub release asset URLs:

https://github.com/Loyalsoldier/v2ray-rules-dat/releases/download/202605222252/apple-cn.txt
https://github.com/Loyalsoldier/v2ray-rules-dat/releases/download/202605212258/apple-cn.txt

In this case, the files appeared to be identical by content: both responses had the same Content-Length and the same Azure-specific x-ms-blob-content-md5 header. However, their standard HTTP validators, ETag and Last-Modified, were different because the URLs resolved to different release asset objects.

This means conditional requests with If-None-Match / If-Modified-Since are correct for detecting whether the same HTTP resource has changed, but they cannot reliably detect content equality across different release asset objects without relying on provider-specific headers such as x-ms-blob-content-md5.

kirilllivanov:
This is not meant to continue the PR discussion. I just want to clarify a technical point about checking whether two web resources are identical before downloading them.

Meo597:
引入哈希校验如果只有几十行代码当然 ok
但现在 1000 多行下来，带不来任何显著的收益

最重要的是我没法维护

PR #6185, opened 2026-05-24
Sudoku finalmask: Harden UDP ReadFrom() against invalid packets

### LINKED ISSUE #6184: [Bug] finalmask UDP listener dies on first invalid packet, kernel buffer overflows silently (Hysteria2/XHTTP3, v26.5.9)
### Integrity Requirements

- [x] I read all comments in the issue template and confirm my submission complies.
- [x] I have read the documentation and understand each option in my configs.
- [x] I provide complete configs and logs, not selective excerpts.
- [x] I searched issues and did not find a duplicate (related: #6182, #6088 — different root cause).
- [x] The problem is reproducible on the latest Release version (26.5.9).

### Description

`streamSettings.finalmask` on a UDP-backed listener (XHTTP/3 `packet-up` or Hysteria2) gets permanently killed by the **first invalid UDP packet** received on the socket. After this happens, the listener stops calling `recvfrom()` on the underlying UDP socket. The kernel receive buffer fills (`ss -uanm` shows `r1049088/rb1048576 d<growing>`), and **all subsequent UDP packets — including from legitimate clients — are silently dropped at the kernel level**. The xray process keeps running and accepts new TCP connections, but the affected UDP port is dead until `xray.service` is restarted.

Trigger: any non-decodable UDP packet hitting the port. In production this is constant — internet background-noise UDP scans, NAT keep-alives from misconfigured peers, packets from xray clients on the same version that do **not** apply sudoku-mask to outgoing QUIC traffic. Within seconds of starting the listener, the first stray packet kills it.

Even an "empty" `finalmask` config that contains only `quicParams` (no `udp[]` entry) is enough to activate the broken wrapper and reproduce the bug — the moment the listener is wrapped in `finalmask`'s `udpConn`, it inherits the fatal error-propagation behavior.

### Reproduce

1. Start xray 26.5.9 with the server config below (Hysteria2 + `finalmask.udp[]` with type `sudoku`).
2. From any host on the internet, send one malformed UDP packet to the listener's port (e.g. `echo -n 'X' | nc -u -w1 <server> 4433`).
3. Observe in xray error.log: `[Error] transport/internet/hysteria: failed to serve hysteria > quic: transport closed: unexpected EOF` (or `invalid sudoku hint tuple` for sudoku-mask).
4. Start a legitimate client → its UDP packets arrive at the NIC (visible in `tcpdump -i any udp port 4433`) but xray never logs receipt and never responds.
5. `ss -uanm 'sport = :4433'` shows the socket's receive buffer filled (`r1049088/rb1048576`) and `d<N>` (dropped) counter growing on every client retry.
6. `sudo systemctl restart xray.service` → the very next stray packet kills the listener again.

### Server Config (minimal repro)

```jsonc
{
  "inbounds": [
    {
      "tag": "hy2-repro",
      "listen": "0.0.0.0",
      "port": 4433,
      "protocol": "vless",
      "settings": {
        "clients": [{ "id": "00000000-0000-0000-0000-000000000000", "flow": "xtls-rprx-vision" }],
        "decryption": "mlkem768x25519plus.native.600s.<REDACTED-X25519-PRIVKEY>"
      },
      "streamSettings": {
        "network": "hysteria",
        "security": "tls",
        "tlsSettings": {
          "serverName": "example.com",
          "alpn": ["h3"],
          "certificates": [{
            "certificateFile": "/etc/ssl/example.com.crt",
            "keyFile": "/etc/ssl/example.com.key"
          }]
        },
        "hysteriaSettings": { "version": 2, "password": "<32-byte-hex-secret>" },
        "finalmask": {
          "udp": [{ "type": "sudoku", "settings": { "password": "<32-byte-hex-secret>" } }],
          "quicParams": {
            "congestion": "force-brutal",
            "brutalUp": "50 mbps",
            "brutalDown": "100 mbps",
            "udpHop": { "ports": "30001-40000", "interval": "15-30" }
          }
        }
      }
    }
  ],
  "outbounds": [{ "protocol": "freedom", "tag": "direct" }]
}
```

The same bug reproduces on `network: "xhttp"` + `mode: "packet-up"` + `alpn: ["h3"]` (XHTTP/3 transport). Removing the `finalmask` block entirely makes both transports stable.

### Server Log

```
[Error] transport/internet/hysteria: failed to serve hysteria > quic: transport closed: unexpected EOF
[Error] transport/internet/splithttp: failed to serve HTTP/3 for XHTTP/3 > quic: transport closed: invalid sudoku hint tuple
```

`error.log` produces one such line at startup (sometimes within the first second), then **goes completely silent** even as `tcpdump` confirms UDP packets continue arriving from real clients.

### Socket-level Evidence

```
$ ss -uanm 'sport = :4433'
State       Recv-Q       Send-Q              Local Address:Port
UNCONN      1049088      0                       0.0.0.0:4433
     skmem:(r1049088,rb1048576,t0,tb212992,f3328,w0,o0,bl0,d5275)
                                                              ^^^^^
                                              dropped packets counter
                                              (grows on every legitimate client retry)
```

`r ≈ rb` (receive queue at receive buffer limit) is the definitive signal: the listener has stopped draining the kernel socket.

### Root Cause (hypothesis)

`transport/internet/finalmask/sudoku/conn_udp.go` (and likely sibling files for other UDP `finalmask` types) propagates decode errors as fatal:

```go
n, addr, err = c.conn.ReadFrom(c.readBuf)
hints, decoded, err = decodeBytes(c.tables, &tableIndex, c.readBuf[:n], hints, decoded)
if err != nil {
    return 0, addr, err   // <-- fatal return
}
```

quic-go interprets any `net.PacketConn.ReadFrom` error as a fatal transport-level failure and terminates its `go serve()` goroutine. Since `finalmask`'s `udpConn.ReadFrom` is the only thing draining the kernel socket, the socket buffer then fills indefinitely.

Suggested fix: on `decodeBytes` failure, **`continue`** the read loop (drop the unparseable datagram) instead of returning the error. Only return on actual underlying socket errors (`c.conn.ReadFrom` failures, EOF, etc.). This matches the behavior real-world UDP listeners need: garbage on the wire is normal and must not kill the listener.

### Workaround

Remove `streamSettings.finalmask` entirely from any UDP-backed inbound until fixed:

```bash
sudo jq '(.inbounds[] | select(.streamSettings.network=="hysteria" or .streamSettings.network=="xhttp") | .streamSettings) |= del(.finalmask)' \
  /usr/local/etc/xray/config.json > /tmp/xray_config_new.json
sudo install -m 600 /tmp/xray_config_new.json /usr/local/etc/xray/config.json
sudo systemctl restart xray.service
```

After workaround, Hysteria2 and XHTTP/3 listeners survive arbitrary garbage on the wire.

### Cost of Workaround

`finalmask sudoku` on UDP listeners is sold as anti-probe shaping. Worth noting: xray 26.5.9 **clients** do not appear to apply the matching sudoku-mask to outgoing QUIC packets — so even before this bug, the on-wire QUIC remained plain. The defensive value of sudoku on Hysteria2/XHTTP/3 in 26.5 may need a separate review independent of this bug fix.

### Environment

- Xray Version: 26.5.9 (release binary)
- OS: Ubuntu 22.04 / Debian 12 (verified on both)
- Kernel: 5.15 / 6.1
- Hardware: bare-metal VPS
- Architecture: amd64

### Related

- #6182 — `hysteria2 + transport=hysteria2` UDP port-mapping silently fails. Different symptoms (no error in log, dokodemo/portMap involved), but possibly the same underlying `finalmask` UDP-wrapper class of bugs.
- #6088 — startup-time config rejection for sudoku finalmask (already fixed).

kuishou68:
## Summary

When `udpConn.ReadFrom` fails on an invalid UDP packet, continue reading the next datagram instead of returning a fatal error. This prevents quic-go from terminating its `Accept` goroutine when garbage arrives on the wire.

## Problem

The current implementation in `conn_udp.go` returns a fatal error when `decodeBytes` fails:

```go
decoded, err := pc.decodeBytes(buff[:n])
if err != nil {
    return 0, nil, err  // <-- fatal to quic-go
}
```

quic-go interprets any non-nil error from `ReadFrom` as a fatal transport-level failure and terminates its `Accept` goroutine. Since `udpConn` is the only thing draining the kernel socket, the socket buffer then fills indefinitely.

## Fix

Modified `ReadFrom` to loop and drop unparseable datagrams instead of returning errors:

```go
decoded, err := pc.decodeBytes(buff[:n])
if err != nil {
    continue  // skip bad datagram, keep reading
}
```

This matches how the sibling `salamander` implementation handles decode errors gracefully.

## Validation

- Any non-decodable UDP packet is silently dropped
- The listener continues processing subsequent packets
- The kernel receive buffer no longer fills indefinitely
- Legitimate clients continue to work after garbage packets

Fixes #6184

### DIFF
--- a/transport/internet/finalmask/sudoku/conn_udp.go
+++ b/transport/internet/finalmask/sudoku/conn_udp.go
@@ -39,26 +39,30 @@ func (c *udpConn) ReadFrom(p []byte) (n int, addr net.Addr, err error) {
 	c.readMu.Lock()
 	defer c.readMu.Unlock()
 
-	n, addr, err = c.conn.ReadFrom(c.readBuf)
-	if err != nil {
-		return n, addr, err
-	}
+	for {
+		n, addr, err = c.conn.ReadFrom(c.readBuf)
+		if err != nil {
+			return n, addr, err
+		}
 
-	decoded := make([]byte, 0, n/4+1)
-	hints := make([]byte, 0, 4)
-	tableIndex := 0
-	hints, decoded, err = decodeBytes(c.tables, &tableIndex, c.readBuf[:n], hints, decoded)
-	if err != nil {
-		return 0, addr, err
+		decoded := make([]byte, 0, n/4+1)
+		hints := make([]byte, 0, 4)
+		tableIndex := 0
+		hints, decoded, err = decodeBytes(c.tables, &tableIndex, c.readBuf[:n], hints, decoded)
+		if err != nil {
+			// Drop unparseable datagram and continue reading
+			continue
+		}
+		if len(hints) != 0 {
+			// Incomplete hints, drop and continue
+			continue
+		}
+		if len(p) < len(decoded) {
+			return 0, addr, io.ErrShortBuffer
+		}
+		copy(p, decoded)
+		return len(decoded), addr, nil
 	}
-	if len(hints) != 0 {
-		return 0, addr, io.ErrUnexpectedEOF
-	}
-	if len(p) < len(decoded) {
-		return 0, addr, io.ErrShortBuffer
-	}
-	copy(p, decoded)
-	return len(decoded), addr, nil
 }
 
 func (c *udpConn) WriteTo(p []byte, addr net.Addr) (n int, err error) {
--- a/transport/internet/finalmask/sudoku/conn_udp.go
+++ b/transport/internet/finalmask/sudoku/conn_udp.go
@@ -1,10 +1,12 @@
 package sudoku
 
 import (
-	"io"
+	"context"
 	"net"
 	"sync"
 	"time"
+
+	"github.com/xtls/xray-core/common/errors"
 )
 
 type udpConn struct {
@@ -50,15 +52,16 @@ func (c *udpConn) ReadFrom(p []byte) (n int, addr net.Addr, err error) {
 		tableIndex := 0
 		hints, decoded, err = decodeBytes(c.tables, &tableIndex, c.readBuf[:n], hints, decoded)
 		if err != nil {
-			// Drop unparseable datagram and continue reading
+			errors.LogErrorInner(context.Background(), err, "drop packet from ", addr, " with size ", n)
 			continue
 		}
 		if len(hints) != 0 {
-			// Incomplete hints, drop and continue
+			errors.LogError(context.Background(), "drop packet from ", addr, " with size ", n)
 			continue
 		}
 		if len(p) < len(decoded) {
-			return 0, addr, io.ErrShortBuffer
+			errors.LogError(context.Background(), "drop packet from ", addr, " with size ", n)
+			continue
 		}
 		copy(p, decoded)
 		return len(decoded), addr, nil
@@ -72,16 +75,15 @@ func (c *udpConn) WriteTo(p []byte, addr net.Addr) (n int, err error) {
 	// UDP decoding restarts at table 0 for every datagram, so encoding must do the same.
 	encoded, err := newCodec(c.tables, c.pMin, c.pMax).encode(p)
 	if err != nil {
-		return 0, err
+		errors.LogErrorInner(context.Background(), err, "drop packet to ", addr, " with size ", len(p))
+		return 0, nil
 	}
 
-	nn, err := c.conn.WriteTo(encoded, addr)
+	_, err = c.conn.WriteTo(encoded, addr)
 	if err != nil {
 		return 0, err
 	}
-	if nn != len(encoded) {
-		return 0, io.ErrShortWrite
-	}
+
 	return len(p), nil
 }
 


transport/internet/finalmask/sudoku/conn_udp.go
```diff
@@ -39,26 +39,30 @@ func (c *udpConn) ReadFrom(p []byte) (n int, addr net.Addr, err error) {
...
+		hints, decoded, err = decodeBytes(c.tables, &tableIndex, c.readBuf[:n], hints, decoded)
+		if err != nil {
+			// Drop unparseable datagram and continue reading
+			continue
+		}
+		if len(hints) != 0 {
+			// Incomplete hints, drop and continue
+			continue
+		}
+		if len(p) < len(decoded) {
```
LjhAUMEM, line 62: let's continue on this case

RPRX:
@LjhAUMEM 你看一下

LjhAUMEM:
他返回的 io.ErrShortBuffer 实际上也会让 quic 停止，建议 continue，~但他似乎不看 review~

RPRX:
@LjhAUMEM 你的 maintain 权限能直接改他的仓库吗，还是说得 admin 权限

Fangliding:
自己改了小铅笔来回复制 是麻烦不是不可以(没法git push)

Fangliding:
和仓库写权限在一起的

RPRX:
@LjhAUMEM 如果你能改的话直接改了吧

LjhAUMEM:
刚准备用铅笔发先 writeto 也需要改，我 clone 一下吧

LjhAUMEM (approved):

PR #6187, opened 2026-05-24
Feature: VlessRoute forwarding between incoming and outgoing  VLESS-connections

fanymagnet:
Hi. When we have an intermediate server between the client and the end server, the client loses the ability to influence routing on the end server.

For example, when the end server has direct or WARP output, we can allow the client to choose the behavior on the end server without duplicating the configuration on the intermediate server.

In any case, it seems right to me that we will forward the route identifier for the entire chain of servers, if it is a chain of VLESS servers.


Fangliding:
似乎有些太奇怪了 这可能导致不希望的vlessroute被传递

PR #6191, opened 2026-05-24
Feature: VlessRoute forwarding between incoming and outgoing  VLESS-connections 

fanymagnet:
Related #6187 

Propagate inbound `VlessRoute` to outbound VLESS connections only when the outbound UUID does not already define its own route bits (`outboundVlessRoute == 0`).

This preserves explicit outbound routing configuration while allowing route inheritance in multi-hop VLESS setups without duplicating configuration on intermediate servers. Since existing outbound route values always take priority, the risk of breaking backward compatibility should be minimal.

### DIFF
--- a/proxy/vless/outbound/outbound.go
+++ b/proxy/vless/outbound/outbound.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	gotls "crypto/tls"
 	"encoding/base64"
+	"encoding/binary"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


fanymagnet:
@Fangliding What exactly is the issue? I understand your concern was backward compatibility, but this change does not introduce that risk.

Fangliding:
我只是顺带指出了你之前修改的一个缺陷 不代表不能改/改了就没问题了 如果目的没问题只是技术小缺陷我不会关而是会修好它 你没有考虑的事情还包括携带1的数据可能因为mux被2带出去等等 根本原因是这样在出入站间隐式传递信息 我也很不喜欢这样做 它们应当是解耦的 尽管有些人可能会因为能在他们的细分需求节省很多config字数而喜欢它

PR #6202, opened 2026-05-26
core: fix nil pointer handling on object creation

alrs:
The comment on `core.CreateObject()` says that it will accept a `nil` Xray `*core.Instance`. In my testing this is incorrect, as passing a nil pointer will cause a `panic` when trying to dereference the non-existent `context.Context`.

This adds a check for `nil` in `CreateObject()`, the custom error type `NilInstanceErr` for the condition, and a test to verify that `nil` pointers are handled correctly.

### DIFF
--- a/core/functions.go
+++ b/core/functions.go
@@ -12,12 +12,19 @@ import (
 	"github.com/xtls/xray-core/transport/internet/udp"
 )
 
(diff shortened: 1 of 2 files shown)


Fangliding:
类似的修复需要在core中存在真实触发路径

PR #6231, opened 2026-06-01
leastPing balancer: prefer reliable outbounds over fast-but-flaky ones

Milomilo777:
## Problem

`leastPing` selects purely on `Alive && Delay`. With a `burstObservatory`, `Delay` is derived from the average of only the **successful** probes (`app/observatory/burst/healthping_result.go`, `getStatistics`), and `Alive` is true as long as a single probe in the window succeeded. So an outbound that is being throttled or blocked — succeeding on a few probes and failing the rest — keeps a tiny `Delay` while dropping most of its traffic, and `leastPing` locks onto that fast‑but‑mostly‑dead tunnel instead of a slower but healthy one. This is a common failure mode under DPI‑based throttling.

## Change

Use the reliability data the observatory already collects:

- skip candidates whose recent health‑check failure rate exceeds 50% (`HealthPing.Fail / HealthPing.All`);
- break exact `Delay` ties by the lower latency deviation (the steadier link).

## Compatibility

Fully backward compatible:

- the plain observer leaves `HealthPing` nil → selection stays by `Delay` only;
- if every candidate is flaky, it falls back to the original lowest‑`Delay` pick, so the result is never worse than before.

No config or proto changes. The `leastPing` name and its behavior on healthy networks are unchanged (covered by a regression test).

## Tests

Adds `app/router/strategy_leastping_test.go` — offline, table‑driven tests covering: lowest‑delay‑wins (regression guard), rejecting a fast‑but‑flaky node, nil‑`HealthPing` pure‑delay fallback, deviation tie‑break, all‑flaky graceful fallback, dead/non‑candidate filtering, and the empty case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/router/strategy_leastping.go
+++ b/app/router/strategy_leastping.go
@@ -2,6 +2,7 @@ package router
 
 import (
 	"context"
+	"math"
 
 	"github.com/xtls/xray-core/app/observatory"
 	"github.com/xtls/xray-core/common"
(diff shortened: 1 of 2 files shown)


Fangliding:
哪来的机器人。。。

bytecategory:
> 哪来的机器人。。。

为什么他就不来我的仓库贡献呢,毕竟是Opus4.8 @Milomilo777

PR #6251, opened 2026-06-03
Add configurable XHTTP session ID format

XXcipherX:
### Summary

This PR adds an optional `sessionIdFormat` setting for XHTTP client-side session ID generation.

By default, XHTTP keeps the existing canonical UUID behavior. Users can opt into other opaque token shapes when a deployment environment needs session IDs to fit a different HTTP-visible format.

### Motivation

XHTTP session IDs may appear in HTTP-visible locations depending on `sessionPlacement`, such as path, query, header, or cookie. Some HTTP intermediaries apply strict handling or filtering to specific token shapes.

This change makes the textual shape of the client-generated session ID configurable without changing XHTTP protocol semantics.

### Supported values

| Value | Shape | Purpose |
| --- | --- | --- |
| `uuid` | Canonical UUID string, e.g. `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` | Compatibility mode and default behavior. |
| `random-hex` | 32 lowercase hex characters generated from 16 cryptographically random bytes | Hash/fingerprint-like shape, similar to common cache keys, ETags, or asset fingerprints. |
| `random-base62` | 22 cryptographically random base62 characters, e.g. `aZ8pQ2nLm7VxK0sYdE4rBt` | Compact token/id-like shape, similar to common opaque web identifiers. |

Example:

```json
{
  "sessionIdFormat": "random-base62"
}
```
## Compatibility

This change is backward compatible.

Existing configurations are unaffected because an unset sessionIdFormat keeps the existing runtime behavior. The server side does not require special handling because XHTTP session IDs are already handled as opaque strings.

### DIFF
--- a/infra/conf/transport_internet.go
+++ b/infra/conf/transport_internet.go
@@ -221,6 +221,7 @@ type SplitHTTPConfig struct {
 	UplinkHTTPMethod     string            `json:"uplinkHTTPMethod"`
 	SessionPlacement     string            `json:"sessionPlacement"`
 	SessionKey           string            `json:"sessionKey"`
+	SessionIdFormat      string            `json:"sessionIdFormat"`
 	SeqPlacement         string            `json:"seqPlacement"`
 	SeqKey               string            `json:"seqKey"`
 	UplinkDataPlacement  string            `json:"uplinkDataPlacement"`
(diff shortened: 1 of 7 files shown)


lie-must-die:
Thanks, it worked!

Zover1337:
Thank you
work!

speznaz97:
Thanks! That fixed the 403 error

win64exe:
Thanks, it worked!

DarkWood312:
++ Working

cbrgpl:
Nice, it works

rubybwn:
Thank you it works

Fangliding:
还是我重新弄一版吧

ghost:
once again this is incorrect, it creates a string of FIXED LENGTH

PR #6252, opened 2026-06-03
Fix xhttp user agent selections, headers arent applied when aren't in extra

ghost:
@RPRX

### DIFF
--- a/common/utils/browser.go
+++ b/common/utils/browser.go
@@ -264,24 +264,18 @@ func applyMasqueradedHeaders(header http.Header, browser string, variant string)
 }
 
 func TryDefaultHeadersWith(header http.Header, variant string) {
-	// The global UA special value handler for transports. Used to be called HandleTransportUASettings.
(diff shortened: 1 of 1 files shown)


Fangliding:
这里是为了区分用户未填(执行默认)和填空(取消UA头)

ghost:
So what's wrong with this fix exactly? It does not change behavior, it only allows us to set required user agent via "headers" and nothing else

ghost:
I checked the code before pushing this fix, https://github.com/search?q=repo%3AXTLS%2FXray-core%20TryDefaultHeadersWith&type=code

it does nothing but prevents us from utilizing User-Agent header and nothing else

ghost:
https://github.com/NoHumanNeeded333/Xray-core/commit/d1a6217354de48a62838af72ba15eef00c3e941c#diff-f898f86523782721aa861d5d5f610608420ae29acf3a27eb71fa618590179d19R272-R274

@Fangliding port this too

ghost:
image

PR #6253, opened 2026-06-03
Proper fix for session key obfuscation

ghost:
This is something that should've been done in the very beginning instead of awaiting 1-2 years before it got finally patched simply by blocking any "uuid" looking format (considering you're not even pinning the version octet and variant) on nginx level by CDNs

PR #6260, opened 2026-06-03
xhttp: add configurable session ID format and length

hexband:
## Summary

This PR adds configurable XHTTP session ID generation.

Previously, XHTTP always generated session identifiers as UUID strings. This change introduces:

- `sessionIdFormat`
- `sessionIdLength`

allowing to choose alternative session ID representations and optionally use variable-length identifiers.

## Supported formats

- `uuid` (default)
- `hex`
- `base64url`
- `base32`
- `base58`
- `alnum`

## Example

```json
{
  "sessionIdFormat": "base58",
  "sessionIdLength": "16-24"
}
```

## Motivation

Some CDN providers return HTTP 403 responses when a UUID string is detected anywhere in the HTTP request.

This PR allows to select alternative session ID formats that remain unique while avoiding infrastructure-specific UUID filtering rules.

### UUID (existing behavior)

```json
{
  "sessionIdFormat": "uuid"
}
```

### Base58 with variable length

```json
{
  "sessionIdFormat": "base58",
  "sessionIdLength": "16-32"
}
```

### Base64URL

```json
{
  "sessionIdFormat": "base64url",
  "sessionIdLength": "20-40"
}
```

## Backward compatibility

Default behavior remains unchanged.

If `sessionIdFormat` is not specified, XHTTP continues to generate UUID session identifiers exactly as before.

Existing configurations require no changes.


ghost:
half of this can be covered via custom alphabet option from #6253

PR #6271, opened 2026-06-05
Prevent XHTTP response padding on invalid requests

XXcipherX:
## Summary

This PR changes when XHTTP response padding is added.

Before this change, the server added response padding very early in `ServeHTTP`, before the request was fully validated. Because of that, even invalid requests could get XHTTP-specific padding in the response headers/cookies together with an error status.

This moves the padding step to the successful response paths only, avoiding XHTTP-specific response padding on requests that are rejected anyway.

## Details

Response padding is still added for normal successful XHTTP traffic:

- `stream-up`
- `packet-up`
- `stream-down`
- `stream-one`

It is no longer added to error responses such as invalid padding, unsupported method, oversized upload, or failed upload queue operations.

`OPTIONS` responses also no longer get response padding. They are used for CORS/preflight handling, so they should not need XHTTP padding.

## Compatibility / risk

This should be low risk for existing clients. Valid XHTTP requests keep the same response padding behavior.

The only intended behavior change is for invalid/error responses: they now return the error without adding XHTTP response padding.

No request validation, session handling, upload/download handling, or padding generation logic is changed.

### DIFF
--- a/transport/internet/splithttp/hub.go
+++ b/transport/internet/splithttp/hub.go
@@ -106,24 +106,26 @@ func (h *requestHandler) ServeHTTP(writer http.ResponseWriter, request *http.Req
 	}
 
 	h.config.WriteResponseHeader(writer, request.Method, request.Header)
-	length := int(h.config.GetNormalizedXPaddingBytes().rand())
-	config := XPaddingConfig{Length: length}
+	applyResponsePadding := func() {
+		length := int(h.config.GetNormalizedXPaddingBytes().rand())
+		config := XPaddingConfig{Length: length}
 
-	if h.config.XPaddingObfsMode {
-		config.Placement = XPaddingPlacement{
-			Placement: h.config.XPaddingPlacement,
-			Key:       h.config.XPaddingKey,
-			Header:    h.config.XPaddingHeader,
+		if h.config.XPaddingObfsMode {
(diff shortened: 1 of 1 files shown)


Fangliding:
感觉没有啥必要 毕竟path是秘密的 后面的无效请求多是配置错误而不是需要处理的常见无效请求

XXcipherX:
> 感觉没有啥必要 毕竟path是秘密的 后面的无效请求多是配置错误而不是需要处理的常见无效请求

Fair point. If the path is random enough, this is probably not a big issue.

My concern is that Xray does not enforce that. Some users may still use paths like /, /api/ or /gateway/, and then random scanners can hit the XHTTP handler quite easily.

So this is mostly a small hardening change: valid clients should behave the same, while invalid requests leak fewer XHTTP-specific signs.

Fangliding:
所有 http based 的transport 基本都是如此设计的用path进行简单的鉴权几行就可以避开的事情

XXcipherX:
> 所有 http based 的transport 基本都是如此设计的用path进行简单的鉴权几行就可以避开的事情

Got it. A properly random path should avoid most random probes.

My point was just that if a request does reach the XHTTP handler, it may be cleaner not to include XHTTP-specific padding in rejected responses. But I understand this is more of a small hardening/cleanup change than a required fix.

PR #6277, opened 2026-06-05
 feat(stats): expose per-inbound online users 

younesvatan78:
## Summary

Expose inbound-scoped online users through OnlineMaps and metrics, without adding a new stats API method.

## Motivation

Xray currently tracks online users under `user>>>EMAIL>>>online`, which is useful for aggregate per-user online status. Some management panels need to know which inbound a user is online through when the same logical user/email is configured on multiple inbounds.

## Changes

- Continue registering the existing `user>>>EMAIL>>>online` map.
- Additionally register `inbound>>>TAG>>>user>>>EMAIL>>>online` when the inbound tag is available.
- Export OnlineMaps through metrics using the full OnlineMap name as the key.
- Keep `GetAllOnlineUsers` and `GetUsersStats` limited to legacy `user>>>...>>>online` maps so existing callers are not affected.
- Add focused tests for legacy filtering.

## Testing

```console
GOPATH=/tmp/xray-go GOMODCACHE=/tmp/xray-go/pkg/mod GOCACHE=/tmp/xray-go/cache go test ./app/stats ./app/stats/command ./app/dispatcher ./app/metrics
```

Result:

```console
ok  	github.com/xtls/xray-core/app/stats	(cached)
ok  	github.com/xtls/xray-core/app/stats/command	0.007s
ok  	github.com/xtls/xray-core/app/dispatcher	(cached)
?   	github.com/xtls/xray-core/app/metrics	[no test files]
```


Fangliding:
API已经够繁乱了 请考虑使用 https://xtls.github.io/config/routing.html#webhookobject 跟踪

younesvatan78:
> API已经够繁乱了 请考虑使用 https://xtls.github.io/config/routing.html#webhookobject 跟踪

Thanks, understood.

Webhook can expose `email` + `inboundTag`, but it seems to be a route-match event only. I could not find a connection-end/offline event.

For this use case we need exact current online state per inbound, like the existing `user>>>email>>>online` map, where the IP is removed when the connection context ends. With webhook the panel can only approximate this via lastSeen/TTL, which may be inaccurate for idle or recently disconnected users. Deduplication also seems keyed by email, so it would need to be disabled for multi-inbound users.

If TTL-based activity tracking is the preferred direction, I can rework around webhook. Otherwise this PR keeps existing API behavior unchanged and adds exact inbound-scoped online tracking.

Fangliding:
也许你可以为不同的入站设置不同的user来跟踪它们的情况
通常情况下一台机器只会有一个入站+数个用户 有多个入站且要跨入站管理这些用户的情况似乎比较少见

younesvatan78:
> 也许你可以为不同的入站设置不同的user来跟踪它们的情况 通常情况下一台机器只会有一个入站+数个用户 有多个入站且要跨入站管理这些用户的情况似乎比较少见

Yes, using different user emails per inbound would work technically.

The issue is that some panels manage one logical user across multiple inbounds/protocols/ports. In that model, the same user/email is intentionally reused so traffic, expiry, limits, and user management stay unified. If we duplicate the user per inbound only for online tracking, the panel then has to split one real user into multiple artificial users and merge their traffic/limits/state again.

I agree this is probably less common than the single-inbound setup. That is why I tried to keep the change additive and backward-compatible: existing `user>>>email>>>online`, `GetAllOnlineUsers`, and `GetUsersStats` behavior remains unchanged, while panels that need multi-inbound attribution can query the inbound-scoped online state.

If you prefer not to add a new API method, another possible compromise is only registering the inbound-scoped online maps and letting external tools query them by exact stat name, without adding more API surface.

younesvatan78:
For context, this use case comes from Xray-based panels such as MHSanaei/3x-ui:
https://github.com/MHSanaei/3x-ui

3x-ui can expose one logical client across multiple inbounds/protocols/ports, while keeping expiry, traffic limit, total usage, enable/disable state, and UI management tied to that same client.

So the repeated email across inbounds is intentional at the panel layer, not just a configuration mistake. The problem is only that current online tracking is aggregated by user/email, so the panel cannot know which inbound the client is actually online through.

Fangliding:
> If you prefer not to add a new API method, another possible compromise is only registering the inbound-scoped online maps and letting external tools query them by exact stat name, without adding more API surface.

看起来这是可以考虑的

https://github.com/XTLS/Xray-core/commit/8562dcb19ffbe36d1650017e3cc701d90a79631b 这样是否足够？整个 onlinemap会在 metrics 导出

younesvatan78:
> > If you prefer not to add a new API method, another possible compromise is only registering the inbound-scoped online maps and letting external tools query them by exact stat name, without adding more API surface.
> 
> 看起来这是可以考虑的
> 
> [8562dcb](https://github.com/XTLS/Xray-core/commit/8562dcb19ffbe36d1650017e3cc701d90a79631b) 这样是否足够？整个 onlinemap会在 metrics 导出

Thanks, I reworked it in this direction and opened a smaller PR:
https://github.com/XTLS/Xray-core/pull/6278

It removes the new gRPC stats API method. Now it only adds the inbound-scoped OnlineMap:

`inbound>>>TAG>>>user>>>EMAIL>>>online`

and exposes OnlineMaps through metrics using the full map name as the key.

I also kept `GetAllOnlineUsers` and `GetUsersStats` limited to `user>>>...>>>online`, so existing API behavior should stay unchanged.

younesvatan78:
@Fangliding No rush. Just to clarify, I won’t open or update another PR until the preferred direction is agreed. If the metrics-based approach is acceptable, I can prepare the changes after confirmation.

younesvatan78:
@Fangliding Hi, sorry to bother again. Did you have any chance to think about whether the metrics-based approach is acceptable?

No rush — I just want to make sure I follow the preferred direction before doing anything else.

Fangliding:
看起来你仍然想内置追踪整个user的用量的逻辑 以为我只是想避免暴露API 实际上我想避免的是前面那个
我加的 metrics 的意思是外部软件可以自行导出所有入站的情况然后自行统计 不用反复发出query

younesvatan78:
> 看起来你仍然想内置追踪整个user的用量的逻辑 以为我只是想避免暴露API 实际上我想避免的是前面那个 我加的 metrics 的意思是外部软件可以自行导出所有入站的情况然后自行统计 不用反复发出query

Thanks, I understand now. Sorry, I misunderstood your point.

Would a connection lifecycle webhook be more aligned with your preferred direction?

For example, core would not store new per-inbound user OnlineMaps and would not add any API. It would only emit events such as:

`start`: connectionId, email, inboundTag, source, ts
`end`: connectionId, email, inboundTag, source, ts

Then external software like 3x-ui can maintain its own online state and statistics by itself.

This would avoid putting panel-specific tracking logic into core, while still allowing external software to calculate exact per-inbound online state without TTL guessing.

Fangliding:
我想通过现有 api query 几个入站并处理它们的 onlinemap 已经够了 不用在core中再放helper(能有多少入站呢？)

younesvatan78:
> 我想通过现有 api query 几个入站并处理它们的 onlinemap 已经够了 不用在core中再放helper

Thanks, understood.

Just to make sure I understand correctly: which OnlineMap names should external software query for per-inbound user online state?

Currently I can query `user>>>EMAIL>>>online`, but that is still aggregated by user and not scoped to inbound.

For example, if the same email exists on two inbounds, querying `user>>>EMAIL>>>online` only tells me the user is online somewhere, not which inbound they are online through.

Are you suggesting there is already an existing inbound-scoped OnlineMap name that I should query, or should this be calculated from other existing metrics?

Fangliding:
> 也许你可以为不同的入站设置不同的user来跟踪它们的情况

Fangliding:
我承认这确实比较麻烦 也是当初onlinemap的设计问题 当初这个东西我也很反对但是最终还是合进来了导致维护和后续用户都得为它买单 所以我尽可能减少又堆新的统计系统层层加码

younesvatan78:
> 我承认这确实比较麻烦 也是当初onlinemap的设计问题 当初这个东西我也很反对但是最终还是合进来了导致维护和后续用户都得为它买单 所以我尽可能减少又堆新的统计系统层层加码

Thanks, I understand the concern better now.

I agree it would be bad to keep adding more layers to the stats system. I’ll avoid pushing another OnlineMap/stat helper approach in core and think more about a panel-side design that does not add maintenance burden to Xray.


Thanks for your time and effort, i highly appreciate it.

Fangliding:
webhook我看了一下它不太能通过简单的修改订阅connection end事件

younesvatan78:
> webhook我看了一下它不太能通过简单的修改订阅connection end事件

Thanks for checking the webhook path.

Then maybe the smallest non-invasive solution is still to reuse the existing OnlineMap mechanism, but without adding any new API/helper.

Core already updates `user>>>EMAIL>>>online` at the point where it has `email`, `inboundTag`, source IP, and connection context cleanup.

The minimal change would be only:

- keep existing `user>>>EMAIL>>>online` unchanged
- additionally register `inbound>>>TAG>>>user>>>EMAIL>>>online` when `statsUserOnline` is enabled
- remove it with the same context cleanup
- keep existing APIs filtered to `user>>>...>>>online`, so old behavior does not change
- expose/query it only through raw OnlineMap/metrics

This should be a small additive change, not a new stats system or new API surface. The extra memory is only for active `inboundTag + email + IP` entries, not all configured users.

Would this smaller scope be acceptable?

younesvatan78:
But if you still prefer not to expand/change OnlineMap at all and keep the current behavior as-is, I understand. I just wanted to clarify the smallest scope I had in mind.

younesvatan78:
> webhook我看了一下它不太能通过简单的修改订阅connection end事件

I want to clarify one concern from the panel side.

A similar OnlineMap metrics export to the one proposed in #6278 was added directly in 8562dcb. I appreciate that direction, but metrics export alone still does not solve the actual per-inbound attribution problem, because the only existing online key is still `user>>>EMAIL>>>online`.

Without an inbound-scoped online key, panels have no exact way to know which inbound the user is online through. The alternatives are all messy or inaccurate: TTL-based webhook guessing, traffic inference, or creating artificial per-inbound Xray users and merging them back into one logical panel user for expiry, traffic limits, enable/disable state, and UI.

That is why I’m asking again whether the minimal additive map could be accepted:

`inbound>>>TAG>>>user>>>EMAIL>>>online`

No new API/helper is needed, existing behavior can stay unchanged, and external software can consume it through the existing/raw OnlineMap metrics.

If you reopen the PR or prefer a new one after the direction is agreed, I can adjust the changes to exactly this smaller scope.

I understand the concern about expanding OnlineMap, but this small addition would avoid pushing a lot of complexity and inconsistency into panels.

PR #6302, opened 2026-06-09
Fix DNS parallel query priority group fallback

Medium1992:
## Summary

This PR fixes `enableParallelQuery` so DNS queries follow the same priority/fallback semantics as the grouped DNS selection logic.

Previously, parallel DNS mode started async queries for all selected DNS clients immediately, even when those clients belonged to different policy groups. As a result, lower-priority fallback groups could receive real DNS queries before the higher-priority group had failed.

This was especially visible with configurations that put `fakedns` first and real DNS resolvers in a later group: even when FakeDNS was supposed to answer first, the real resolvers were still queried in parallel.

## Problem

The DNS code already builds dynamic groups based on matching server properties such as:

- `clientIP`
- `skipFallback`
- `queryStrategy`
- `tag`
- `domains`
- `expectedIPs`
- `unexpectedIPs`

The intended behavior is:

1. Race servers inside the current group.
2. If one server in the group succeeds, return that result.
3. Only if the whole group fails, fall back to the next group.

However, the previous implementation called `asyncQueryAll()` for all selected clients before processing groups. This meant grouping only affected which result was accepted, not which DNS servers were actually queried.

## Change

`parallelQuery` now starts async queries group by group:

- all servers inside one group are still queried in parallel;
- the first successful result in the current group is returned;
- the next group is queried only if the current group fully fails.

This preserves parallelism inside each priority group while restoring fallback behavior between groups.

## Why

This makes `enableParallelQuery` respect DNS priority semantics instead of bypassing them at query dispatch time.

It also prevents unintended DNS queries to lower-priority resolvers when a higher-priority group, such as FakeDNS or a domain-specific resolver group, succeeds.

### DIFF
--- a/app/dns/dns.go
+++ b/app/dns/dns.go
@@ -387,50 +387,30 @@ func (s *DNS) parallelQuery(domain string, option dns.IPOption) ([]net.IP, uint3
 	var errs []error
 	clients := s.sortClients(domain)
 
-	resultsChan := asyncQueryAll(domain, option, clients, s.ctx)
+	groups, _ := makeGroups( /*s.ctx,*/ clients)
 
-	groups, groupOf := makeGroups( /*s.ctx,*/ clients)
-	results := make([]*queryResult, len(clients))
-	pending := make([]int, len(groups))
-	for gi, g := range groups {
-		pending[gi] = g.end - g.start + 1
-	}
+	for _, g := range groups {
+		groupClients := clients[g.start : g.end+1]
+		resultsChan := asyncQueryAll(domain, option, groupClients, s.ctx)
+		results := make([]queryResult, len(groupClients))
 
-	nextGroup := 0
-	for range clients {
-		result := <-resultsChan
-		results[result.index] = &result
+		for i := range groupClients {
+			result := <-resultsChan
+			results[result.index] = result
+			if result.err == nil && len(result.ips) > 0 {
+				return result.ips, result.ttl, nil
+			}
 
-		gi := groupOf[result.index]
-		pending[gi]--
-
-		for nextGroup < len(groups) {
-			g := groups[nextGroup]
-
-			// group race, minimum rtt -> return
-			for j := g.start; j <= g.end; j++ {
-				r := results[j]
-				if r != nil && r.err == nil && len(r.ips) > 0 {
-					return r.ips, r.ttl, nil
+			if i == len(groupClients)-1 {
+				for j, r := range results {
+					e := r.err
+					if e == nil {
+						e = dns.ErrEmptyResponse
+					}
+					errors.LogInfoInner(s.ctx, e, "failed to lookup ip for domain ", domain, " at server ", groupClients[j].Name(), " in parallel query mode")
+					errs = append(errs, e)
 				}
 			}
-
-			// current group is incomplete and no one success -> continue pending
-			if pending[nextGroup] > 0 {
-				break
-			}
-
-			// all failed -> log and continue next group
-			for j := g.start; j <= g.end; j++ {
-				r := results[j]
-				e := r.err
-				if e == nil {
-					e = dns.ErrEmptyResponse
-				}
-				errors.LogInfoInner(s.ctx, e, "failed to lookup ip for domain ", domain, " at server ", clients[j].Name(), " in parallel query mode")
-				errs = append(errs, e)
-			}
-			nextGroup++
 		}
 	}
 


Medium1992:
One practical issue with the current behavior is the FakeDNS flow.

With `fakedns` as the first DNS server and real resolvers in a later group:

1. The client sends a DNS query.
2. Xray returns a fake IP quickly.
3. At the same time, real DNS queries are already dispatched to lower-priority resolvers.
4. The client immediately connects to the fake IP.
5. FakeDNS sniffing maps the fake IP back to the original domain.
6. Routing/outbound resolution may then request the real IP for the same domain.

In practice, the client can connect to the fake IP faster than the earlier real DNS queries finish and populate the cache. This may cause resolution work for the same domain to happen again immediately after the fake DNS response.

More importantly, real DNS queries are sent before they are actually needed. For FakeDNS setups, the expected behavior is usually:

- DNS query from the client: return fake IP only;
- connection to fake IP: recover the domain;
- routing/outbound stage: resolve the real IP only if needed, using the configured resolver strategy.

This PR makes that flow deterministic by preserving parallel queries inside the active group, while delaying lower-priority groups until fallback is actually required.

Meo597:
那这样并行还有什么意义

Medium1992:
> 那这样并行还有什么意义

So they run in parallel within their group.
Alternatively, could we add another JSON key to determine the behavior of parallel requests—globally or by group?
Because the current implementation is a bit odd: let’s say there are 20 DNS servers, each with its own domain settings, order, and logic. In parallel mode, a DNS name query goes to all 20, when in fact it should be directed to, say, a single group.
Not to mention that with fakeDNS, the logic ends up being completely wrong.
Configuration example.
```
{
  “dns”: {
    “tag”: “dns-inbound”,
    “hosts”: {
      “dns.google”: [
        “8.8.8.8”,
        “8.8.4.4”
      ],
      “dns.quad9.net”: [
        “9.9.9.9”,
        “149.112.112.112”
      ],
      “cloudflare-dns.com”: [
        “104.16.248.249”,
        “104.16.249.249”
      ]
    },
    “servers”: [
      {
        “tag”: “fakeip”,
        “address”: “fakedns”
      },
      {
        “tag”: “ParallelQuery”,
        “address”: “https://dns.google/dns-query”
      },
      {
        “tag”: “ParallelQuery”,
        “address”: “https://cloudflare-dns.com/dns-query”
      },
      {
        “tag”: “ParallelQuery”,
        “address”: “https://dns.quad9.net/dns-query”
      }
    ],
    “queryStrategy”: “UseIPv4”,
    “enableParallelQuery”: true
  },
  “fakedns”: {
    “ipPool”: “198.18.0.0/15”,
    “poolSize”: 131070
  }
}
```

Theoretically, the logic should be such that only fake results are returned.
However, when resolving a domain at the routing or exit level—for example, in Freedom—a parallel domain resolution occurs. Why send duplicate requests? Why generate so many of them?

Medium1992:
image
image

Meo597:
你贴的是无效配置，因为你忘记了 `dns.domains` 字段

Meo597:
并行并不是你所理解的全局、或者是组内。而是每次都在动态规划
https://xtls.github.io/en/config/dns.html#dns-processing-flow

一个使用场景是
https://xtls.github.io/ru/document/level-1/routing-with-dns.html

Medium1992:
> 你贴的是无效配置，因为你忘记了 `dns.domains` 字段

I tried adding domains with “regex:.” to fakeDNS
The behavior hasn't changed; parallel mode ignores everything—it just sends requests to all DNS servers

Medium1992:
> 并行并不是你所理解的全局、或者是组内。而是每次都在动态规划 https://xtls.github.io/en/config/dns.html#dns-processing-flow
> 
> 一个使用场景是 https://xtls.github.io/ru/document/level-1/routing-with-dns.html

Please consider the option to select the behavior of parallel mode:
- Global
- Group

Global - Current behavior
Group - Behavior follows the DNS flow for groups as defined in the group descriptions

动态分组，被选中的服务器列表中相邻的服务器如果 clientIP skipFallback queryStrategy tag domains expectedIPs unexpectedIPs 完全一样，被视为同一个组。

Medium1992:
Or perhaps support for the `address` array and defining behavior within a single section?

Meo597:
对于反审查软件要治理的是分流和 dns 污染，必须要全局并行查询

组内并行毫无意义，你应该仔细阅读上面发给你的两个链接

Medium1992:
> 对于反审查软件要治理的是分流和 dns 污染，必须要全局并行查询
> 
> 组内并行毫无意义，你应该仔细阅读上面发给你的两个链接

Why doesn't it make sense?
I gave you a specific example where a fake is served and domain requests are also sent to other upstreams—that's incorrect behavior.

Requests should occur after the fake has been replaced with the domain by the sniffer, and during the routing stage or when exiting to an outbound server that doesn't support domain-based routing.

The configuration I provided behaves illogically.

Apparently, you don’t understand me...

Parallel requests are a cool feature, but they’re currently limited to a single behavior that’s precisely illogical and violates the logic of the entire DNS section.

Medium1992:
> 你贴的是无效配置，因为你忘记了 `dns.domains` 字段

```
{
  "dns": {
    "tag": "dns-inbound",
    "hosts": {
      "dns.google": [
        "8.8.8.8",
        "8.8.4.4"
      ],
      "dns.quad9.net": [
        "9.9.9.9",
        "149.112.112.112"
      ],
      "cloudflare-dns.com": [
        "104.16.248.249",
        "104.16.249.249"
      ]
    },
    "servers": [
      {
        "tag": "fakeip",
        "address": "fakedns",
        "domains": ["domain:xray.com"]
      },
      {
        "tag": "ParallelQuery",
        "address": "https://dns.google/dns-query"
      },
      {
        "tag": "ParallelQuery",
        "address": "https://cloudflare-dns.com/dns-query"
      },
      {
        "tag": "ParallelQuery",
        "address": "https://dns.quad9.net/dns-query"
      }
    ],
    "queryStrategy": "UseIPv4",
    "enableParallelQuery": true
  },
  "fakedns": {
    "ipPool": "198.18.0.0/15",
    "poolSize": 131070
  }
}
```

image
image
image

Meo597:
其它服务器没有 domains 字段即为通配所有
你应该用 finalQuery 终结查询链

Medium1992:
> 其它服务器没有 domains 字段即为通配所有 你应该用 finalQuery 终结查询链

I've tried all sorts of configurations.
I can't seem to get the parallel requests to work properly with fakeDNS.

Meo597:
```jsonc
  // xxx...
  {
    "tag": "fakeip",
    "address": "fakedns",
    "domains": ["domain:xray.com"],
    "finalQuery": true
  },
  // xxx...
```

Medium1992:
> ```json-comments
>   // xxx...
>   {
>     "tag": "fakeip",
>     "address": "fakedns",
>     "domains": ["domain:xray.com"],
>     "finalQuery": true
>   },
>   // xxx...
> ```

I've tried every possible combination—either it breaks or it keeps behaving the same way. I swapped the placeholders, but nothing works.

Parallel mode is a cool feature, but it's broken right now. And there's no normal mode.

Meo597:
> but nothing works.
> but it's broken right now. And there's no normal mode.

它一切正常，或许你应该求助更好的模型，或者认真阅读：

- https://xtls.github.io/en/config/dns.html#dns-processing-flow
- https://xtls.github.io/ru/document/level-1/routing-with-dns.html

PR #6307, opened 2026-06-10
XHTTP transport: Do not force trailing path `/` when both sessionID and seq are placed elsewhere

dudkin-2005:
`GetNormalizedPath` always appends a trailing slash to the configured path.
That slash is only meaningful when sessionID and/or seq are carried in the
path: it separates the configured base path from the appended segments and
lets the server slice them back off via `req.URL.Path[len(path):]`.

When both sessionID and seq are placed in query/cookie/header, the path
carries no appended segments, so the forced trailing slash is unnecessary.
It also makes URLs look unnatural — a directory-style trailing slash right
before the query string, or a file path that gets a slash appended after its
extension.

This change makes the trailing slash conditional on path placement being
used for sessionID or seq. The default placement is `path`, so any config
that does not move both fields off the path keeps the exact previous
behaviour; only configs that move both fields off the path are affected.

### Examples

**Both meta fields in query**

```json
"xhttpSettings": {
  "mode": "packet-up",
  "path": "/stream",
  "sessionIDPlacement": "query",
  "seqPlacement": "query"
}
```

Before:

```
GET  /stream/?x_session=<id>
POST /stream/?x_seq=0&x_session=<id>
```

After:

```
GET  /stream?x_session=<id>
POST /stream?x_seq=0&x_session=<id>
```

File-like path — `"path": "/stream/filename.extension"`, both in query:
Before → `/stream/filename.extension/?x_session=<id>` 
After → `/stream/filename.extension?x_session=<id>`

Unchanged: default (path) placement — `"path": "/abc"`:

```
GET  /abc/<sessionID>
POST /abc/<sessionID>/<seq>
```

sessionID and seq are in the path, so the trailing slash is still added — identical to the current behaviour.

Unchanged: root path — `"path": "/"` stays `/?...` in all cases.

### DIFF
--- a/transport/internet/splithttp/config.go
+++ b/transport/internet/splithttp/config.go
@@ -24,8 +24,11 @@ func (c *Config) GetNormalizedPath() string {
 		path = "/" + path
 	}
 
-	if path[len(path)-1] != '/' {
-		path = path + "/"
+	if c.GetNormalizedSessionPlacement() == PlacementPath ||
+		c.GetNormalizedSeqPlacement() == PlacementPath {
+		if path[len(path)-1] != '/' {
+			path = path + "/"
+		}
 	}
 
 	return path
--- a/transport/internet/splithttp/config_test.go
+++ b/transport/internet/splithttp/config_test.go
@@ -3,16 +3,77 @@ package splithttp_test
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	. "github.com/xtls/xray-core/transport/internet/splithttp"
 )
 
 func Test_GetNormalizedPath(t *testing.T) {
-	c := Config{
-		Path: "/?world",
+	tests := []struct {
+		TestName           string
+		Path               string
+		SessionIDPlacement string
+		SeqPlacement       string
+		Expected           string
+	}{
+		{
+			TestName: "default placement keeps trailing slash",
+			Path:     "/sh",
+			Expected: "/sh/",
+		},
+		{
+			TestName: "query string is stripped",
+			Path:     "/?world",
+			Expected: "/",
+		},
+		{
+			TestName:           "both off path drops trailing slash",
+			Path:               "/stream",
+			SessionIDPlacement: "query",
+			SeqPlacement:       "query",
+			Expected:           "/stream",
+		},
+		{
+			TestName:           "both off path keeps file-like path",
+			Path:               "/stream/filename.extension",
+			SessionIDPlacement: "query",
+			SeqPlacement:       "header",
+			Expected:           "/stream/filename.extension",
+		},
+		{
+			TestName:           "seq in path keeps trailing slash",
+			Path:               "/stream",
+			SessionIDPlacement: "query",
+			Expected:           "/stream/",
+		},
+		{
+			TestName:     "session in path keeps trailing slash",
+			Path:         "/stream",
+			SeqPlacement: "cookie",
+			Expected:     "/stream/",
+		},
+		{
+			TestName:           "existing trailing slash preserved",
+			Path:               "/stream/",
+			SessionIDPlacement: "query",
+			SeqPlacement:       "query",
+			Expected:           "/stream/",
+		},
+		{
+			TestName:           "root unchanged",
+			Path:               "/",
+			SessionIDPlacement: "query",
+			SeqPlacement:       "query",
+			Expected:           "/",
+		},
 	}
-
-	path := c.GetNormalizedPath()
-	if path != "/" {
-		t.Error("Unexpected: ", path)
+	for _, test := range tests {
+		t.Run(test.TestName, func(t *testing.T) {
+			c := Config{
+				Path:               test.Path,
+				SessionIDPlacement: test.SessionIDPlacement,
+				SeqPlacement:       test.SeqPlacement,
+			}
+			assert.Equal(t, test.Expected, c.GetNormalizedPath())
+		})
 	}
 }


artur002:
By the way, you can check the patch and include this code snippet in the branch. You do realize this is an inherent shortcoming, right? Xray has always been highly flexible when it comes to traffic configuration. This looks more like an oversight.

ghost:
@RPRX merge

RPRX:
我得评估下新旧版本兼容性问题，~~盲猜是 breaking~~，@dudkin-2005 你测过吗

dudkin-2005:
> 我得评估下新旧版本兼容性问题，~盲猜是 breaking~，@dudkin-2005 你测过吗

Yes, I tested this; an xray update on the server is required because it can't validate the path.

[Info] transport/internet/splithttp: failed to validate path, request:/stream/filename.extension, config:/stream/filename.extension/

If the client uses a version of xray without this commit, and the server does have this commit, then this will work, because the serveHttp function in hub.go checks whether the path of the incoming HTTP request starts with a certain prefix:

```
	if !strings.HasPrefix(request.URL.Path, h.path) {
		errors.LogInfo(context.Background(), "failed to validate path, request:", request.URL.Path, ", config:", h.path)
		writer.WriteHeader(http.StatusNotFound)
		return
	}
```

ghost:
> 我得评估下新旧版本兼容性问题，盲猜是 breaking，@dudkin-2005 你测过吗

you've already committed breaking change with sessionTables, it won't change much, people who need backward compatibility can add / at the end of their path, additionally you can create a sub option in config in extra to disable slash behavior for query mode if you really need that, but i dont see any reasons to, query and path are split into two different independent chunks even by nginx

PandaWorker:
In xhttp, path assembly and parsing was initially poorly designed for flexibility, I would like to hear opinions on why template uris were not used, such as in masque, but with the flexibility of keys.


```
path-template: "/api/v1/users/{session}/posts/{seq}/image.png"
seq-key: seq
seq-placement: path-template
session-key: session
session-placement: path-template
```


```
path-template: "/api/v1/users/{session}/posts{?offset}"
seq-key: offset
seq-placement: path-template
session-key: session
session-placement: path-template
```

nehamoff:
@RPRX merge, on my configs, they added a block on paths with / at the end, and most people encountered this problem. This is a common issue on many CDN services.

artur002:
@RPRX I don't quite get why still force a trailing slash at the end of the path string with no way to disable it.

Right now, using a clean endpoint like **/name.file** transforms the request into **/name.file/**?... The requests never even reach the server.

If the main reason for not merging a fix is the fear of breaking backward compatibility — why not introduce an exclamation mark "**!**" at the end of the path as an opt-in flag? If present, the core strips it and leaves the path completely clean, without appending any slash. Old configs won't be affected at all.

Example: **"path": "/name.file!"** -> emits a clean /name.file over the wire.

Implementation snippet:

``` GO
if path[len(path)-1] != '/' {
    if path[len(path)-1] == '!' {
        path = path[:len(path)-1]
    } else {
        path = path + "/"
    }
```

vadimsky-ctrl:
Adding the concrete impact, since it's a bit stronger than "looks unnatural": on some CDNs the forced trailing slash is the difference between the tunnel working and a hard `403` on every request — minimal repro in #6410. A path ending in `/` is treated as *dynamic* and its query string is scanned by an anti-cache-poisoning WAF (which matches the `sessionId` token), while a path ending in `<name>.<ext>` is treated as a static asset and skipped. Removing the slash gives 25/25 `204` end-to-end vs. a 100% `403` rate with the same config otherwise.

On the backward-compat concern: the only breaking direction is **new client → old server** — the old server normalizes its own path to `…/`, so `HasPrefix(request.Path, config.Path)` then rejects the slash-less request (matches @dudkin-2005's `failed to validate path` log). An **opt-in** form sidesteps that completely, since you'd only enable it on a matched client+server pair while every existing config stays byte-identical — e.g. @artur002's `!` suffix (`"path": "/name.file!"` → clean `/name.file`) or a small `extra` flag. Given how common this CDN block is, an opt-in looks like a cheap unblock.

elder-rus:
Reporting an adjacent case that hits the same function, since sessionPlacement/seqPlacement aren't involved at all here:
GetNormalizedPath() in transport/internet/splithttp/config.go unconditionally forces a trailing / onto whatever precedes the first ? in path. This also fires when path itself contains a literal embedded query string, e.g.:
json"path": ".../index.jpeg=?eHa7cn7SskUMfEHBC5lmnmcuZbpSUWjchoENvLe4AR69U47ZIekyR6TuvPLXMvt"
gopathAndQuery := strings.SplitN(c.Path, "?", 2)
path := pathAndQuery[0]           // "...index.jpeg="
if path[len(path)-1] != '/' {     // last char is '=', not '/'
    path = path + "/"             // → "...index.jpeg=/"
}
Result: the outgoing request becomes
.../index.jpeg=/?eHa7cn7SskUMfEHBC5lmnmcuZbpSUWjchoENvLe4AR69U47ZIekyR6TuvPLXMvt
instead of the intended
.../index.jpeg=?eHa7cn7SskUMfEHBC5lmnmcuZbpSUWjchoENvLe4AR69U47ZIekyR6TuvPLXMvt
Since this PR conditions the forced slash on session/seq using path placement, this case may slip through the current scope — there's no session/seq path segment here at all, the corruption is purely from normalizing whatever sits before the first ?. Might be worth having GetNormalizedPath() only add / when something is actually going to be appended after it, rather than unconditionally normalizing the raw path value itself.

AI generated (sorry bad at eng)

hxehex:
plz merge or I will murder

RPRX:
~~几天没看咋刷了这么多，实际上一开始这里似乎主要是为了少写点代码，只要不影响默认（非 obfs）的 XHTTP，那么会合并~~

RPRX:
~~若要不强制 `/`，对于 stream-up/down 不用判断 seq 在哪，对于 stream-one 不用判断 sessionID 和 seq 在哪，不过炸得更多~~

所以暂时先这样吧，毕竟你们的需求主要是穿透力最强的 packet-up + stream-down 模式

PR #6321, opened 2026-06-13
fix(wireguard): prevent races and double-close panics during shutdown

hossinasaadi:


Fangliding:
wireguard 正在经历大范围重构 #6287 
如果问题在重构分支仍然存在建议在下面指出

hossinasaadi:
still exists.

Crashed: Thread
0x12fc5a4 golang.zx2c4.com/wireguard/device.NewDevice + 196
0x12fc534 golang.zx2c4.com/wireguard/device.NewDevice + 84
0x1315fc0 github.com/xtls/xray-core/proxy/wireguard.(*Handler).init + 432
0x13149ac github.com/xtls/xray-core/proxy/wireguard.(*Handler).Process + 220
0x12dc504 github.com/xtls/xray-core/app/proxyman/outbound.(*Handler).Dispatch + 2116
0xe82df0 github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).routedDispatch + 2224
0xe81890 github.com/xtls/xray-core/app/dispatcher.(*DefaultDispatcher).Dispatch.gowrap1 + 48
0x81a524 runtime.goexit.abi0 + 4

@LjhAUMEM would you please take a look at this when you have a chance?

LjhAUMEM:
@hossinasaadi Any configuration that can be triggered, or is it just a coincidence?

hossinasaadi:
> @hossinasaadi Any configuration that can be triggered, or is it just a coincidence?

its happens in coincidence

LjhAUMEM:
That’s strange. If it’s a “closed” error, how could the call stack still show NewDevice? Do you have the full logs?

hossinasaadi:
@LjhAUMEM sorry for the late reply. I only have the Firebase Crashlytics stack trace, not the full logs.

I do not think this is caused by a specific config. I believe that it is a shutdown race, for example with a short startup timeout like 5s (ping), WireGuard init may still run while the core is already closing. That would explain why Crashlytics still shows `NewDevice` in the stack.

LjhAUMEM:
Can you make sure that you are tracking the latest source code? After the refactoring, there should no longer be an init function, but it appears in your stack trace. Also, only line 196 matches the current source code.

hossinasaadi:
Yes, the version is 26.6.27.

The crash line is different because of compiler optimization.

I believe this is the init function:
https://github.com/XTLS/Xray-core/blob/1aabe7ea78eca88deedc234a3e668895fa9bc08c/proxy/wireguard/client.go#L258

PR #6333, opened 2026-06-17
tls: present client certificate on outbound connections

alicommit-malp:
Configured tlsSettings.certificates were wired only into GetTLSConfig's server-side GetCertificate callback, so an Xray client never offered a certificate when a server requested one, making outbound mutual TLS impossible. copyConfig (crypto/tls.Config -> utls.Config) also dropped the certificate fields, so even a manually set client certificate was lost on the default uTLS/fingerprinted path.

Set config.Certificates and GetClientCertificate in GetTLSConfig when ENCIPHERMENT certificates are configured (client-only; server still uses GetCertificate), and carry the certificates through copyConfig so mutual TLS works on the uTLS path.

### DIFF
--- a/transport/internet/tls/config.go
+++ b/transport/internet/tls/config.go
@@ -411,6 +411,24 @@ func (c *Config) GetTLSConfig(opts ...Option) *tls.Config {
 		config.GetCertificate = getNewGetCertificateFunc(c.BuildCertificates(), c.RejectUnknownSni)
 	}
 
+	// Present a client certificate for mutual TLS on outbound connections.
+	// Configured certificates are otherwise wired only into GetCertificate (the
+	// server-side selection callback), so an Xray client never offers a
+	// certificate when a server requests one. Set both Certificates (so the
+	// standard tls.Client path works and copyConfig can carry them to the
+	// uTLS/fingerprinted path) and GetClientCertificate. Client-only; harmless
+	// server-side, where GetCertificate takes precedence.
+	if built := c.BuildCertificates(); len(built) > 0 {
+		clientCerts := make([]tls.Certificate, 0, len(built))
+		for _, bc := range built {
+			clientCerts = append(clientCerts, *bc)
+		}
+		config.Certificates = clientCerts
+		config.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return &clientCerts[0], nil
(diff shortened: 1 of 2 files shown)


Fangliding:
我最近正在重构tls模块以实现mtls
btw 我最近看到了好几个错误实现mtls的pr 而且它们无一例外都只有client实现 我从没看到server实现 我想你们有类似的目的？

alicommit-malp:
Thanks, and great to hear the TLS module is getting proper mTLS upstream — happy to drop this in favor of that.

On the client-only point: you're right, this is client-side only by design. In our case the server requiring/verifying the client cert is a Caddy edge in front (client_auth require_and_verify), so we only needed Xray to present a client cert on outbound TLS to pass that gate — we don't run an Xray server that verifies peer certs. One thing worth carrying into the refactor: outbound cert presentation has to survive the uTLS/fingerprinted path too — copyConfig currently drops the cert fields, so setting them on the crypto/tls config alone isn't enough on the default path.

Happy to test the new mTLS implementation against our setup when it's ready.

Fangliding:
实际上它已经在 https://github.com/XTLS/Xray-core/tree/mtls 了 只需要在客户端配置证书并且在 usage 写上 "client-cert" 就行了 你可以测试一下

alicommit-malp:
Hi :) 

quick field report on the mTLS work (PR #6333 / mtls branch). I implemented your official client-cert approach — config cert + usage: "client-cert", the getClientCertificate callback honoring
  cri.SupportsCertificate, carried through copyConfig to the uTLS path via the uGetClientCertificate adapter. To stay on Go 1.25, I backported those changes onto the released v1.260206.0 tag rather than building the
  branch HEAD directly, but the code path is the same as yours. It works end-to-end against a real relay. Setup: VLESS + XHTTP outbound → Caddy terminating TLS with client_auth { mode require_and_verify } as the
  verifier. The Xray client presents the per-server client cert, Caddy admits the handshake, and the tunnel comes up. Tested on Windows (amd64) and Linux. 
  
  The key confirmation: client-cert presentation works on the
  fingerprinted uTLS / XHTTP path, not just plain crypto/tls. That copyConfig cert-carry is the easy spot to drop the cert, and your adapter handles it correctly. Happy to share more detail or test specific commits if
  useful. Thanks for landing this.

PR #6334, opened 2026-06-17
Fragment finalmask: Add `lengths` and `delays`

patterniha:
more control on lengths and delays of segments:

```
{
  "type": "fragment",
  "settings": {
    "packets": "tlshello",
    "lengths": ["3-5", "6-8", "10-20"],
    "delays": ["1-2", "5-6", "10-20"],
    "maxSplit": 64
  }
}
```
`"lengths": ["3-5", "6-8", "10-20"]` means: use length 3-5 for first segment, 6-8 for second segment, and 10-20 for others.

`"delays": ["1-2", "5-6", "20-30"]` means: wait 1-2 after sending first segment, 5-6 after second segment, and 20-30 after sending each of the others.

`length` and `delay` options are removed, but for compatibility `"length": "3-5"` convert to` "lengths": ["3-5"]` and `"delay": "1-2"` convert to `"delays": ["1-2"]`.

for "tlshello" fragment special value `"delays": ["0-0"]` (or `["0"]` or `[0]`) means merge all segments and then send.

///

reasons:
 1. two years ago GFW checked the len of first packet and if it was less then 120 then it didn't accept fragments for TLS.
 2. currently `"packets": "tlshello", "lengths": ["1", "10000"]` bypass the GFW dpi, so there is no need to fragment more.
 3. in android, you cannot change "initcwnd" so only 10 segments is sent and then OS waiting for "ack", but during this time some segments is written to OS, but OS merge them and send them in one packet after receiving "ack", so some packets are merged, as a result, DPI may be not bypassed, so you should add extra delay after sending 10 segments.
 ...


RPRX:
如果 `packets` 不是 "tlshello" 而是 "1-3"，lengths 和 delays 是每个包都重新计数还是？

patterniha:
The settings (maxSplit, lengths, delays) are for segments of each packet, the settings will be reapplied for the next packet.

RPRX:
要不你一步到位把 `maxSplit` 也加个 s 算了，可以针对 "1-3" 每个包做最大分割限制

`delays` 的特殊语义扩展到 "1-3" 这种也适用，降低用户心智负担

顺便把 Freedom 的 fragment 删了，话说 Freedom 的 noise 比 Finalmask 的多了啥来着？

patterniha:
> Or you could just add an 's' to it all at once `maxSplit`, and then set a maximum split limit for each packet from "1-3".
> 
> `delays`The special semantic extension to "1-3" also applies, reducing the mental burden on users.
> 

???

~I didn't understand exactly what I should do~.
Anyway, It doesn't matter how long tls-client-hello or http-request are, locally they are received in one packet, so "1-1" and "tlshello" are enough and I don't see a need for  "1-2", "1-3", ...

///


> By the way, I deleted the Freedom fragment. By the way, what does Freedom's noise have that Finalmask has?

Just `applyTo` option.

RPRX:
~~换一个好点的翻译器，算了这个 PR 先这样吧，其它的我再看看~~

PR #6360, opened 2026-06-22
WireGuard inbound: Support dynamic peer management

bitwiresys:
## Problem

The WireGuard inbound only supports static peers defined in the JSON config. Every user change requires a full process restart, which breaks the standard xray API-driven user lifecycle used by panels and management tools (AddUserOperation / RemoveUserOperation via the gRPC management API).

## Solution

This PR adds full `proxy.UserManager` support to the WireGuard inbound `Server`, enabling peers to be added and removed at runtime without restarting xray.

### New file: `proxy/wireguard/account.go`

- **`MemoryAccount`** — runtime peer credentials (public key, pre-shared key, allowed IPs). Implements `protocol.Account` (`Equals` / `ToProto`).
- **`PeerConfig.AsAccount()`** — lets API callers use `"@type": "xray.proxy.wireguard.PeerConfig"` in `AddUserOperation` requests, consistent with how every other inbound protocol works.
- **`buildPeerIPC` / `buildRemovePeerIPC`** — produce WireGuard IPC strings for `device.Device.IpcSet`.
- **`parseFirstAddr`** — extracts a `netip.Addr` from a CIDR or plain address (used for the IP→user reverse lookup map).

### Changes to `proxy/wireguard/server.go`

**New fields on `Server`:**
```go
peers     sync.Map    // email (or pubkey) → *protocol.MemoryUser
peersByIP sync.Map    // netip.Addr → *protocol.MemoryUser
peerCount atomic.Int64
```

**`NewServer`** seeds `peers` / `peersByIP` from the static peer list so that `GetUser` / `GetUsers` work immediately for peers configured at startup.

**`HandleConnection`** tags each session's `Inbound.User` with the matching `*protocol.MemoryUser` when the tunnel source IP is in `peersByIP`. This makes the peer identity visible in access logs and enables per-user traffic stats through the existing `stats.Manager`.

**UserManager methods:**

| Method | Behaviour |
|--------|-----------|
| `AddUser` | Calls `device.IpcSet` to install the peer, then records it in the maps. Falls back to the public key when `Email` is empty. |
| `RemoveUser` | Removes from maps, then calls `IpcSet` with `remove=true`. |
| `GetUser` | Lock-free map lookup by email. |
| `GetUsers` | Snapshot of all tracked peers. |
| `GetUsersCount` | Atomic counter, no map scan. |

### New test files

- **`account_test.go`** — covers `MemoryAccount`, `PeerConfig.AsAccount`, IPC helpers, and `parseFirstAddr`.
[...]

PR #6375, opened 2026-06-24
Hysteria inbound: Support `routing`'s `vlessRoute` as well

nasralbek:
### Summary
- allow UUID-like Hysteria auth values to carry route IDs in the third UUID group / Go slice `id[6:8]`
- match configured Hysteria users while ignoring only those route bytes
- return the matched user with the client-presented auth in its Hysteria account
- set existing `VlessRoute` from that auth so existing `vlessRoute` routing rules can be used
- keep transport and app/router unchanged

### Tests
- go test ./proxy/hysteria/account ./proxy/hysteria
- go test ./... -run "^$"
- live checked with two Xray instances using `vlessRoute`: auth route 1 reached route1 backend and auth route 2 reached route2 backend


RPRX:
@LjhAUMEM 看一下

RPRX:
> bytes 6-7

7-8

nasralbek:
> > bytes 6-7
> 
> 7-8

yes you are right i meant go slice id 6 8 it is uuid bytes 7 and 8 or third uuid group i will change wording to avoid confusion

Fangliding:
改其他的出站是否不合适 而且这个参数还叫vlessRoute

nasralbek:
> 改其他的出站是否不合适 而且这个参数还叫vlessRoute

i changed it to a separate hysteriaRoute condition
vlessRoute is no longer used by hysteria

i tested it with two xray instances and route id 1 and 2 both work

Fangliding:
加一个condition更烂了 污染更多地方。。
我的主要意思是这是在其他人的协议里加非标准功能

Fangliding:
还是迁移到 vless+xhttp3（有hysteria同款阻塞控制）或者 vless+hysterica transport 吧 然后直接 vless route 就行了 不然写这么多组合功能不是白写了
Trojan XTLS 的教训告诉我们强行把 vless 功能依葫芦画瓢搬到其他协议上不会有好结果的

RPRX:
相当于只是服务端改动，还行吧，算是一个小 trick，路由那个选项就叫 `vlessRoute`，别改

~~刚收到两个迷惑的 Security Advisories 似乎是不喜欢 VLESS 默认拿那两个字节当 route，实在不行的话可改成 0000 显式启用~~

LjhAUMEM:
瞄了两眼，就是个 str 转 uuid 动我 account 和 transport 干啥，在 server 的 process 拿到 user 后操作不行？

nasralbek:
> 瞄了两眼，就是个 str 转 uuid 动我 account 和 transport 干啥，在 server 的 process 拿到 user 后操作不行？

server process only gets matched user
it does not know the auth sent by client

if i use user account auth then it is server config auth
so route id is always from config not from client

account change lets auth variants match same user
transport change only passes client sent auth to server process

LjhAUMEM:
> it does not know the auth sent by client

不不不，原始认证信息是有的，把 Account 转一下，参考 #6360

nasralbek:
> > it does not know the auth sent by client
> 
> 不不不，原始认证信息是有的，把 Account 转一下，参考 #6360

got it thanks

i refactored it
transport changes are removed

server process now reads client presented auth from user Account
validator returns matched user with Account auth set to that client auth

LjhAUMEM:
@nasralbek app/router 改动删了，使用 vless route

nasralbek:
updated removed app/router changes and now use existing `vlessRoute`

nasralbek:
> @nasralbek app/router 改动删了，使用 vless route

i added hysteriaRoute because previous feedback said own name may be more logical

i removed app/router changes and use existing vlessRoute

LjhAUMEM:
vless route 好像没有空值的说法，要使用只能服务端自己确保所有用户都是 uuid

@nasralbek 先把那两个 test 文件删了，还有一些细节给我吧，我是否可以直接 push 到你的 branch

nasralbek:
removed the two test files

yes you can push to my branch

LjhAUMEM (approved):

nasralbek:
@LjhAUMEM i noticed one issue after checking the final implementation

vless route logic ignores uuid bytes 6-7 during user validation and then uses bytes 6-7 from the client sent uuid as the route id

current hysteria implementation looks like it validates auth by exact match and uses the route id from the configured server auth

so this works in vless
server uuid 00000000-0000-0000-8000-000000000000
client uuid 00000000-0000-0001-8000-000000000000
route 1

but for hysteria the same idea does not work because client auth 0001 is not accepted when server auth is 0000

was this difference intentional or should hysteria follow vless behavior here

LjhAUMEM:
纠正一下，是 7 8，看来认证不比对这两个，让我思考一下

RPRX:
~~存储和对比前肯定都要先把那两个字节置空啊，没测过就 LGTM 吗，人才~~

LjhAUMEM:
确实是没使用过vlessroute，似乎就是为了少写user，hy2把认证器传到传输层还是有点不自然

但hy2这样路由速度应该是比user name路由要慢，转uuid会有损失

PR #6395, opened 2026-06-28
Hysteria inbound: Fix dynamic UUID not accepted

LjhAUMEM:
https://github.com/XTLS/Xray-core/pull/6375#issuecomment-4825923934

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20220118164431-d8423dcdf344
 	github.com/golang/mock v1.7.0-rc.1
 	github.com/google/go-cmp v0.7.0
+	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.3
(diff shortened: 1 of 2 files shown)


LjhAUMEM:
@nasralbek https://github.com/XTLS/Xray-core/actions/runs/28323633800

LjhAUMEM:
这个居然也是合法uuid？`hy2-auth-secret`

nasralbek:
tested this pr and it fixes the issue for me

now hysteria auth behaves like vless route logic
server auth can be 00000000-0000-0000-8000-000000000000
client auth can be 00000000-0000-0001-8000-000000000000
auth is accepted and vlessRoute becomes 1 from client auth

password auth like hy2-auth-secret also stays exact auth because google uuid parser does not treat it as uuid this is good

i also live tested with two routes
client auth 0001 went to route1
client auth 0002 went to route2

one small note maybe not blocker
when Get returns route variant user it creates MemoryAccount only with VR and without Auth
server currently only uses VR there so it works
but maybe keeping Auth as client auth would be more complete

Issue #4621, opened 2025-04-14
Avoid annoying log


megasoheilsh:
### Integrity requirements

- [x] I confirm that I have read the documentation, understand the meaning of all the configuration items I wrote, and did not pile up seemingly useful options or default values.
- [x] I provided the complete config and logs, rather than just providing the truncated parts based on my own judgment.
- [x] I searched issues and did not find any similar issues.
- [x] The problem can be successfully reproduced in the latest Release

### Description

I have been seeing disconnect  connection without any reason , I have tested 10 different internets. Please follow up and find out what the problem is.

### Reproduction Method

ws , xhttp 

### Client config

<details><pre><code>
{
  "log": {
    "loglevel": "warning"
  },
  "inbounds": [
    {
      "tag": "socks",
      "port": 10808,
      "listen": "127.0.0.1",
      "protocol": "socks",
      "sniffing": {
        "enabled": true,
        "destOverride": [
          "http",
          "tls",
          "quic"
        ],
        "routeOnly": false
      },
      "settings": {
        "auth": "noauth",
        "udp": true,
        "allowTransparent": false
      }
    }
  ],
  "outbounds": [
    {
      "tag": "proxy",
      "protocol": "vless",
      "settings": {
        "vnext": [
          {
            "address": "ip",
            "port": 65302,
            "users": [
              {
                "id": "id",
                "email": "t@t.tt",
                "security": "auto",
                "encryption": "none"
              }
            ]
          }
        ]
      },
      "streamSettings": {
        "network": "xhttp",
        "xhttpSettings": {
          "mode": "packet-up"
        }
      }
    }
  ],
  "routing": {
    "domainStrategy": "AsIs",
    "rules": [
    ]
  }
}
</code></pre></details>

### Server config

<details><pre><code>
{
  "inbounds": [
    {
      "listen": null,
      "port": 65302,
      "protocol": "vless",
      "settings": {
        "clients": [
          {
            "email": "email",
            "flow": "",
            "id": "id"
          }
        ],
        "decryption": "none",
        "fallbacks": []
      },
      "sniffing": {
        "destOverride": [
          "http",
          "tls",
          "fakedns",
          "quic"
        ],
        "enabled": true,
        "metadataOnly": false,
        "routeOnly": false
      },
      "streamSettings": {
        "network": "xhttp",
        "security": "none",
        "xhttpSettings": {
          "headers": {},
          "host": "",
          "mode": "packet-up",
          "noSSEHeader": false,
[...]

PR #6418, opened 2026-07-01
XHTTP: Treat peer HTTP/2 stream reset as EOF on the downlink

lArtiquel:
### What

When the HTTP/2 downlink of an XHTTP session is reset by the peer, the response-body read returns an `http2.StreamError`, which propagates to the outbound handler and gets logged as a processing failure:

```
app/proxyman/outbound: failed to process outbound traffic > ... > stream error: stream ID N; INTERNAL_ERROR; received from peer
```

That's just the remote tearing down the stream — a recurring source of the noise reported in #4621 for XHTTP tunnels.

### Change (rev 2 — moved into the XHTTP transport per review)

Per @Fangliding's note that this belongs in the XHTTP layer rather than the generic outbound handler (it's a no-op for gRPC, which never produces `http2.StreamError`), the handling now lives in `splitConn.Read`: an `http2.StreamError` on the downlink is converted to `io.EOF`, so the existing benign-teardown handling — which already ignores `io.EOF` — treats it as a normal close.

- **Scoped to XHTTP.** Only the HTTP/2 downlink (`http2.Transport` in `dialer.go`) yields `http2.StreamError`. The HTTP/1.1 and HTTP/3 paths, and non-XHTTP transports such as gRPC, are untouched.
- **No change to `app/proxyman/outbound`** anymore.
- Uses `errors.As`, so it also matches a wrapped `StreamError`.

### Testing

- `go build` / `go vet` clean for the package; `gofmt` clean.
- Added `TestSplitConnReadTranslatesH2StreamReset` covering peer / local / wrapped `StreamError` → `io.EOF`, plus passthrough of plain `io.EOF` and unrelated errors.
- Verified on production XHTTP tunnels: the `INTERNAL_ERROR; received from peer` lines that previously appeared on every stream teardown are gone; connectivity unaffected.

Refs #4621


RPRX:
@Fangliding 事实核查

Fangliding:
对grpc没用 xhttp可能有用吧 所以更适合xh文件里不适合放这里

lArtiquel:
@Fangliding thanks — you're right that it's a no-op for gRPC, since that path never yields an `http2.StreamError`. I've moved it out of the outbound handler and into the XHTTP transport: `splitConn.Read` now converts a peer HTTP/2 stream reset to `io.EOF`, and the existing `io.EOF` teardown handling takes it from there. Nothing touches `app/proxyman/outbound` anymore, and it's scoped to the HTTP/2 downlink, so HTTP/1.1, HTTP/3 and gRPC are unaffected. Added a unit test and force-pushed — please take another look when you have a chance.

已把处理移到 XHTTP 传输层：`splitConn.Read` 把对端的 h2 stream reset 转成 `io.EOF`，不再改动 outbound handler；只影响 h2 下行，gRPC / HTTP1.1 / HTTP3 均不受影响。

Fangliding:
经过重复审查后 上游其实包含EOF处理 标准库还是没有那么蠢
image

lArtiquel:
Got it — that makes sense. The genuinely benign stream-end is already handled upstream, so this isn't needed. Thanks to you and @RPRX for taking the time to review it. 👍

明白了，谢谢复审 — 上游已经处理良性关闭，这个改动确实没必要。

PR #6424, opened 2026-07-03
Downgrade golang to 1.24.5 for open-harmony

zhangsan946:
As OpenHarmony SIG only has the golang binary up to 1.24.5, port one branch with go version downgraded to the same for open-harmony build support.

To @reviewer, please help to change the base branch to some other name (probably `open-harmony`) as I am lack of the permission to add a new branch.


Fangliding:
？
我们不会为了这个就降版本 部分行为和功能依赖于高版本go 包括1.27更新后还有迁移要做 有特殊需求请自行维护

PR #6427, opened 2026-07-03
Observatory: Support per-outbound probeUrl override

bitwiresys:
### Summary
- Observatory and BurstObservatory probe every selected outbound against one shared `probeUrl`/`destination`. That works for a generic internet-reachability check, but not for a node pool behind a `leastPing` balancer, where you want each outbound checked against itself (e.g. a `/ping` endpoint the node serves), not a third-party URL
- a single shared probe URL is also a single point of failure: if it goes down or gets rate-limited, every outbound in the pool is reported dead at once, regardless of the outbounds' actual state
- adds an optional `probeUrl` field on the outbound handler config (`outbounds[].probeUrl` in JSON). If set, Observatory/BurstObservatory use it instead of their own global probe URL for that one outbound; if unset, behavior is unchanged
- outbound handlers optionally implement a new `outbound.ProbeURLProvider` interface (`ProbeURL() string`) to expose the override, so this doesn't touch the `outbound.Handler`/`outbound.Manager` interfaces

Closes the same need as #4962 (per-group `probeUrl`/`probeInterval`), which was closed because turning `ObservatoryObject` into an array would break existing configs. Attaching `probeUrl` to the outbound instead is purely additive - no existing config's behavior changes.

### Example
```json
"outbounds": [
  { "tag": "node-nl", "protocol": "vless", "probeUrl": "https://node-nl.example/ping", "settings": {} },
  { "tag": "node-kz", "protocol": "vless", "probeUrl": "https://node-kz.example/ping", "settings": {} }
]
```

### Tests
- `go test ./app/observatory/... ./app/proxyman/outbound/... ./infra/conf/...`
- added `TestOutboundDetourConfig_ProbeURL` (`infra/conf/xray_test.go`): json -> `OutboundHandlerConfig` -> protobuf marshal/unmarshal round trip
- added `TestObservatoryPerOutboundProbeURL` (`testing/scenarios/observatory_test.go`): spins up a real xray instance with two outbounds and two local mock ping endpoints via the gRPC `ObservatoryService`, confirms each outbound is probed on its own URL and gets marked dead independently when only its own endpoint goes down
- `go build ./...`, `go vet ./...` clean

### DIFF
--- a/app/observatory/burst/burstobserver.go
+++ b/app/observatory/burst/burstobserver.go
@@ -99,7 +99,7 @@ func New(ctx context.Context, config *Config) (*Observer, error) {
 	if err != nil {
 		return nil, errors.New("Cannot get depended features").Base(err)
 	}
-	hp := NewHealthPing(ctx, dispatcher, config.PingConfig)
+	hp := NewHealthPing(ctx, outboundManager, dispatcher, config.PingConfig)
 	return &Observer{
 		config: config,
 		ctx:    ctx,
--- a/app/observatory/burst/healthping.go
+++ b/app/observatory/burst/healthping.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/xtls/xray-core/common/dice"
 	"github.com/xtls/xray-core/common/errors"
+	"github.com/xtls/xray-core/features/outbound"
 	"github.com/xtls/xray-core/features/routing"
 )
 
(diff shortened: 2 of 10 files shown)


Fangliding:
在outbound加一个专供obs的字段并用一堆方法导出是很不合适的
probe url 通常来源于知名网站 比如 cp.cloudflare.com 或者 google.com 如果你的机器与这些网站连接有问题本身也象征着某种不可用
顺便你可以在服务端使用路由系统将这些请求导入黑洞出站并且设置http响应 探测器会收到http响应并给出结果 我曾经想避免服务端公网状态对延迟的干扰时曾经使用过

PR #6429, opened 2026-07-04
Skip REALITY non-443 warning for local inbounds

XXcipherX:
Currently Xray warns when a REALITY inbound listens on a non-443 port, even if the inbound is bound only to loopback, such as `127.0.0.1:10000` or `localhost:10000`. In this setup the port is not exposed publicly and may be used behind a local L4/TLS frontend.

This change keeps the warning for public/non-local listeners, but skips it when the inbound listen address is loopback or `localhost`.


RPRX:
https://github.com/XTLS/BBS/issues/21#issuecomment-4109400301

PR #6441, opened 2026-07-07
TLS ECH: Update ECH configuration parsing logic to be more robust

jamesliu96:
Key Changes: Robust Configuration String Splitting

1. Swapped out aggressive, un-delimited splitting in favor of `strings.SplitN(c.EchConfigList, "+", 2)`.
This ensures that if the downstream DoH or h2c URL contains query parameters or search paths that inherently use the plus sign (+) (e.g., `https://some-dns.com/dns-query?token=123+xyz`), it will no longer be erroneously truncated.

2. IPv6 and port detection support. Improved the port-appending mechanism for fallback classic UDP servers using standard library `net.SplitHostPort`.
The previous `strings.Contains(udpServerAddr, ":")` check was fundamentally flawed when users inputted explicit IPv6 formats like `udp://[2001:db8::1]`, bypassing the default :53 port injection.

### DIFF
--- a/transport/internet/tls/config.go
+++ b/transport/internet/tls/config.go
@@ -450,7 +450,7 @@ func (c *Config) GetTLSConfig(opts ...Option) *tls.Config {
 		for _, s := range tls.CipherSuites() {
 			id[s.Name] = s.ID
 		}
-		for _, n := range strings.Split(c.CipherSuites, ":") {
+		for n := range strings.SplitSeq(c.CipherSuites, ":") {
 			if id[n] != 0 {
 				config.CipherSuites = append(config.CipherSuites, id[n])
 			}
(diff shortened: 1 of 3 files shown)


RPRX:
@Fangliding ready 时说一声

Fangliding:
就这样吧

RPRX:
@jamesliu96 ~~顺便说一下你这个 commit 邮箱挺怪的，是 `jamesliu96@users.noreply.github.com`，前面少了 `2211002+`~~

jamesliu96:
@RPRX Thanks for the heads-up! According to GitHub's documentation (see [GitHub Docs](https://docs.github.com/en/account-and-profile/reference/email-addresses-reference#your-noreply-email-address)), accounts created before July 18, 2017 use the USERNAME@users.noreply.github.com format without the ID prefix. That's why it looks a bit different.

PR #6450, opened 2026-07-08
TLS client: Support more `cipherSuites` for "unsafe" (golang) `fingerprint` for anti-NIN

patterniha:
Weirdly, if you use python-tls-fingerprint, you can bypass dpi even on the most restrictive ISPs in Iran.
and for now to imitate python-tls-fingerprint, you just need to imitate it's cipherSuites (of course, the fragment also needs to be set up correctly so that cipherSuites is read by GFW but not SNI)

but even python-3.14.6-default-tls uses some cipherSuites that are listed in go-crypto/tls-InsecureCipherSuites.
so to imitate python-tls-fingerprint, you need to be able to use "InsecureCipherSuites" as well.

~The word "InsecureCipherSuites" is a bit confusing. actually, they are not insecure, they are just obsolete and generally not recommended.~

///
to be more precise, they only allowed python-tls-cipherSuites, which means the first 13 cipher should be almost identical to the first 13 in python-cipherSuites (with a few exceptions), for example the 13th-cipher must be "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (0xc027)", otherwise dpi cannot be bypassed.

also, utls does not have python-fingerprint and none of the other fingerprints, even the old ones, have this feature, so we have to use tls with custom-cipherSuites.

### DIFF
--- a/transport/internet/tls/config.go
+++ b/transport/internet/tls/config.go
@@ -450,9 +450,13 @@ func (c *Config) GetTLSConfig(opts ...Option) *tls.Config {
 		for _, s := range tls.CipherSuites() {
 			id[s.Name] = s.ID
 		}
+		for _, s := range tls.InsecureCipherSuites() {
+			id[s.Name] = s.ID
+		}
 		for n := range strings.SplitSeq(c.CipherSuites, ":") {
-			if id[n] != 0 {
-				config.CipherSuites = append(config.CipherSuites, id[n])
+			n = strings.TrimSpace(n)
+			if v, ok := id[n]; ok {
+				config.CipherSuites = append(config.CipherSuites, v)
 			}
 		}
 	}
--- a/transport/internet/tls/tls.go
+++ b/transport/internet/tls/tls.go
@@ -147,6 +147,11 @@ func GeneraticUClient(c net.Conn, config *tls.Config) *utls.UConn {
 }
 
 func copyConfig(c *tls.Config) *utls.Config {
+	curvePreferences := make([]utls.CurveID, 0, len(c.CurvePreferences))
+	for _, curve := range c.CurvePreferences {
+		curvePreferences = append(curvePreferences, utls.CurveID(curve))
+	}
+
 	config := &utls.Config{
 		Rand:                           c.Rand,
 		RootCAs:                        c.RootCAs,
@@ -156,6 +161,10 @@ func copyConfig(c *tls.Config) *utls.Config {
 		KeyLogWriter:                   c.KeyLogWriter,
 		EncryptedClientHelloConfigList: c.EncryptedClientHelloConfigList,
 		NextProtos:                     c.NextProtos,
+		CipherSuites:                   c.CipherSuites,
+		MinVersion:                     c.MinVersion,
+		MaxVersion:                     c.MaxVersion,
+		CurvePreferences:               curvePreferences,
 	}
 	return config
 }
--- a/transport/internet/tls/tls.go
+++ b/transport/internet/tls/tls.go
@@ -147,11 +147,6 @@ func GeneraticUClient(c net.Conn, config *tls.Config) *utls.UConn {
 }
 
 func copyConfig(c *tls.Config) *utls.Config {
-	curvePreferences := make([]utls.CurveID, 0, len(c.CurvePreferences))
-	for _, curve := range c.CurvePreferences {
-		curvePreferences = append(curvePreferences, utls.CurveID(curve))
-	}
-
 	config := &utls.Config{
 		Rand:                           c.Rand,
 		RootCAs:                        c.RootCAs,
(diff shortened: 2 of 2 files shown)


patterniha:
it seems that the order of the cipherSuites is ignored, the only way to preserve the order is to use `HelloCustom` but Xray-core does not support it.

so more work is needed...

patterniha:
or I can add python-fingerprint to utls, but it will takes a while to merge and release the new version.
@RPRX  what is your opinion? ~I wish you had forked utls.~

RPRX:
不要加这个选项，uTLS 支持 custom，你可以在 Xray-core 内实现个 python 指纹

Fangliding:
copyconfig 里新加的四个参数全是无效的
utls不好自己的外面写一个独立指纹 只能按一个指纹生成了clienthello照猫画虎再改
这样八成又活不了多久 不如就单纯开一个gotls允许InsecureCipherSuites算了

RPRX:
~~还是正常 python 指纹吧，要不就给 uTLS PR 一个~~ @patterniha

patterniha:
> The four newly added parameters in copyconfig are all invalid

yes (except probably for  `hellogolang`), anyway, future utls updates may support them for some other fingerprins, so it would be better to add them.

> It's better to just simply enable Gotls and allow InsecureCipherSuites.

the problem is that Gotls (`unsafe`/`hellogolang`) does not preserve cipherSuites order.

> or maybe submit a uTLS PR.

UTLS updates are released very infrequently.

Fangliding:
> yes (except probably for `hellogolang`), anyway, future utls updates may support them for some other fingerprins, so it would be better to add them.

你在想什么 不支持这些是utls原理所致 扩展由spec自行指定所以才无视config hellogolang有效只是让它们落回go原生的client hello构建路径所以才有效

Fangliding:
> the problem is that Gotls (`unsafe`/`hellogolang`) does not preserve cipherSuites order.

这不是自己给自己拆台？就是说你提交了一个完全不起作用的PR

patterniha:
i tested, `minVersion` and `maxVersion` and `cipherSuites` is valid for UTLS-`hellogolang`, so they should be added to UTLS-Settings.

also, I don't know why, but UTLS-`hellogolang` and `unsafe` are not exactly the same, they differ in three tls-extensions: "supported_groups", "signature_algorithms", and "signature_algorithms_cert".
///

after testing several different modes, I realized that for bypassing dpi, it is enough that the last two non-tls13-ciphers (in the first 13 ciphers) to be `TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256` and `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256`.

fortunately, the order of these two ciphers is preserved, so for bypassing dpi, it is enough to set:
```
"fingerprint": "unsafe",
"cipherSuites": "...:TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256:TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"
```

but these two ciphers are listed in "InsecureCipherSuites" and we should be able to use it.
///

so there is no need to fully emulate Python's fingerprint, and this PR solves the problem.

@RPRX  ready to merge.

~my next task is to add `cipherSuites` to v2rayN/v2rayNG UI-Settings.~

RPRX:
其实伊朗 GFW 很快就可以针对这种特殊指纹，且没啥附带伤害，有什么意义？所以我说不如加 python 指纹

另外 copyConfig 里面加一项就行，有必要加四项吗，不要乱搞 uTLS 的浏览器指纹（虽然它可能也不会遵守）

RPRX:
@Meo597 话说 TLS/uTLS alpn 等行为似乎也应该更详细地 doc 出来，还有 core 里的请求均默认 Chrome 指纹、headers 啥的

~~话说我之前让你写一篇文章你没写的是啥来着~~

Fangliding:
> 另外 copyConfig 里面加一项就行，有必要加四项吗，不要乱搞 uTLS 的浏览器指纹（虽然它可能也不会遵守）

那几项对于正常指纹没一项是生效的 他的论据是对hellogolang有效所以要加 要hellogolang不如去用unsafe直接调真正的gotls。。

Fangliding:
utls的行为之前我就已经写过 仔细看看应该就少了个新加ech参数
image

Meo597:
> 话说我之前让你写一篇文章你没写的是啥来着

finalmask，我一直用不到，没摸过呢还

patterniha:
> That's why I suggest adding Python fingerprinting.

I will try to add python-fingerprint to UTLS later (If GFW uses stricter rules), anyway, this is an independent discussion, this PR just allows us to use "InsecureCipherSuites".

> Don't mess with uTLS browser fingerprinting

no, they only affect `hellogolang` (which is 99% (but not 100%) similar to non-UTLS `unsafe`).

RPRX:
“虽然它可能也不会遵守”但是又有写，代码看着就很迷惑，就 ALPN 一个特例就够了，你实在想加的话只给 unsafe 指纹加

~~虽然那代码也会影响到服务端不过还好对于 VLESS 要么是 REALITY 要么是 TLS behind CDN，所以也就影响 Hy2 和 Trojan~~

patterniha:
>  but it's written there, and the code looks confusing. Just one special case, ALPN, is enough

done.

PR #6460, opened 2026-07-09
feat(mux): add 'drop' option for xudpProxyUDP443 to prevent IP leak in TUN mode

TGIR0:
### What does this PR do?
Adds a new `"drop"` option to `xudpProxyUDP443` in the Mux configuration to silently consume UDP 443 (QUIC) packets instead of abruptly rejecting them.

### Why is this needed? (The Bug)
Currently, when `xudpProxyUDP443` is set to `"reject"`, the outbound handler abruptly returns an error and interrupts the reader/writer. 
In VPN / TUN environments (e.g. Windows/Android TUN mode clients), when the OS network stack receives a sudden connection rejection or ICMP port unreachable for UDP, it assumes the TUN interface is broken for that destination. The OS then performs an automatic fallback to the physical network interface (e.g., Wi-Fi / Cellular). 

This OS-level behavior bypasses the proxy entirely, causing the user's real IP address (e.g. in Iran or China) to be leaked for blocked QUIC traffic.

### How does this fix it?
By setting `"xudpProxyUDP443": "drop"`, the UDP traffic is silently consumed and discarded (similar to the `blackhole` proxy mechanism) using `buf.Discard` with an inactivity timeout. 
Because the connection is dropped silently rather than rejected, the OS does not trigger a network fallback. Instead, the browser's QUIC connection simply times out, forcing it to gracefully fall back to TCP 443 (HTTPS), which is then safely routed through the proxy tunnel without any IP leaks.

### Changes:
- `infra/conf/xray.go`: Added `"drop"` to the valid switch cases for `xudpProxyUDP443`.
- `app/proxyman/outbound/handler.go`: Added the `"drop"` case which copies the `link.Reader` to `buf.Discard` using `signal.CancelAfterInactivity` to prevent goroutine leaks.

### DIFF
--- a/app/proxyman/outbound/handler.go
+++ b/app/proxyman/outbound/handler.go
@@ -6,6 +6,7 @@ import (
 	goerrors "errors"
 	"io"
 	"math/big"
+	"time"
 
 	"github.com/xtls/xray-core/common/dice"
 
@@ -18,6 +19,7 @@ import (
 	"github.com/xtls/xray-core/common/net/cnc"
 	"github.com/xtls/xray-core/common/serial"
 	"github.com/xtls/xray-core/common/session"
+	"github.com/xtls/xray-core/common/signal"
 	"github.com/xtls/xray-core/core"
 	"github.com/xtls/xray-core/features/outbound"
 	"github.com/xtls/xray-core/features/policy"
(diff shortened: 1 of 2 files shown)


Fangliding:
哎 又是AI pr乱猜 core里关闭当前udp conn在tun表现为删除该映射 没有你说的 "connection rejection or ICMP port unreachable" 甚至连 “automatic fallback” 都不知道存不存在
恰恰就是因为是静默丢弃的才会让浏览器继续重试最后偶然绕过QUIC嗅探 所以才会有blackhole的延时关闭机制 这里是看端口的所以不需要

PR #6463, opened 2026-07-09
HTTPUpgrade server: Keep accepting after transient errors as RAW does; Add handshake read deadline as WS does

ranthe21:
not ai

On a few busy servers behind a CDN we kept hitting a state where the httpupgrade inbound goes deaf: the process is fine, the port still shows LISTEN, but nothing gets accepted anymore and every new connection times out until xray is restarted. All other inbounds on the same instance keep working. Took a while to hunt down because nothing gets logged when it happens.

Two things in the httpupgrade hub cause it, and they feed each other:

1. `keepAccepting()` returns on any `Accept()` error. Under fd pressure `Accept()` returns "too many open files", the loop exits silently, and since the listening socket is still open the kernel keeps queueing SYNs into a backlog nobody drains. The port looks alive but every connection hangs. The tcp listener already handles this case (log, short sleep on EMFILE, continue), and ws is safe too since net/http's Serve retries temporary errors. httpupgrade is the only listener that gives up.

2. The upgrade handshake has no read deadline, so a connection that never sends a complete request holds a goroutine and an fd forever. On a public server that junk piles up and is exactly what builds the fd pressure for point 1.

This PR copies the tcp listener's accept loop behavior (only stop when the listener is closed) and puts a 4 second read deadline on the handshake, same value the ws listener uses for ReadHeaderTimeout. The deadline is cleared when the connection is handed over.

Seen in production on nodes proxying traffic through Cloudflare and nginx into this inbound, the port went deaf every day or two under load. `go test ./transport/internet/httpupgrade/` passes with the change.

This image is for some servers only with vless-hu-tls-cdn
image

### DIFF
--- a/transport/internet/httpupgrade/hub.go
+++ b/transport/internet/httpupgrade/hub.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"crypto/tls"
+	"io"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


Fangliding:
就这样吧

PR #6467, opened 2026-07-10
Fix(stats): handle concurrent registration races

sujoshua:
## Issue

Fix concurrent registration races in the stats `GetOrRegister*` helpers.

Previously, multiple goroutines could observe that the same counter, online map,
or channel was missing and then attempt to register it concurrently. The losing
goroutine would receive an `already registered` error instead of returning the
object created by the winning goroutine.

### Adopted

Retry the complete get-or-register sequence when registration fails with
`ErrAlreadyRegistered`.

### Alternative solution

The `Manager` interface is marked as stable, so this change keeps the existing
interface unchanged.

If changing the interface is acceptable, a cleaner solution would be to add an
atomic `Ensure*` or `GetOrRegister*` operation to `Manager` and implement the
lookup and creation under a single write lock. The retry approach used here
preserves compatibility with existing `Manager` implementations while fixing
the current race.

### DIFF
--- a/app/stats/stats.go
+++ b/app/stats/stats.go
@@ -40,7 +40,9 @@ func (m *Manager) RegisterCounter(name string) (stats.Counter, error) {
 	defer m.access.Unlock()
 
 	if _, found := m.counters[name]; found {
-		return nil, errors.New("Counter ", name, " already registered.")
+		return nil, errors.New(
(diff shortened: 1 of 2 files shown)


Fangliding:
哎 标准的race 我重新写一个内部方法吧 改的有点多先关了

PR #6473, opened 2026-07-11
Proxyman: Fix shared inbound proxy ownership

sujoshua:
`AlwaysOnInboundHandler` creates a single inbound proxy and shares it with all TCP, UDP, and Unix workers created for that handler.

The worker shutdown paths currently close this shared proxy, while `AlwaysOnInboundHandler.Close` also closes it after closing the workers. As a result, the same proxy may be closed multiple times, with the exact number depending on the configured networks and port ranges.

## Motivation

Inbound proxy implementations are not required to support repeated or concurrent `Close` calls. Even when a particular implementation happens to tolerate repeated closure, the lifetime of one shared proxy should not depend on the number of workers created for an inbound.

This also establishes the ownership boundary needed for a future graceful shutdown path, where `AlwaysOnInboundHandler` can stop worker listeners independently, drain existing connections, and close the shared inbound proxy only after the drain completes.

### DIFF
--- a/app/proxyman/inbound/worker.go
+++ b/app/proxyman/inbound/worker.go
@@ -149,20 +149,10 @@ func (w *tcpWorker) Start() error {
 }
 
 func (w *tcpWorker) Close() error {
-	var errs []interface{}
-	if w.hub != nil {
-		if err := common.Close(w.hub); err != nil {
-			errs = append(errs, err)
-		}
-		if err := common.Close(w.proxy); err != nil {
-			errs = append(errs, err)
-		}
+	if w.hub == nil {
+		return nil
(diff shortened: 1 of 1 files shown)


Fangliding:
现在不会有问题 几个proxy不是没有就是幂等 就算了 下次重构close一起处理了吧

sujoshua:
下次重构是啥时候。我主要 pr 这个是想把优雅关闭做了，让inbound可以安全drain已有连接，最终目标是实现 reload 功能

Fangliding:
别想了 不会有的 出入站倒是可以用api加所以留了close接口

PR #6476, opened 2026-07-11
splithttp: bind sendThrough=origin to the real per-connection local IP

echoowall:
### Problem

With `sendThrough: "origin"` (source-in-source-out), the outbound gateway is derived from `inbound.Local`, which is set from `conn.LocalAddr()` of the accepted connection (`app/proxyman/inbound/worker.go`).

For the **XHTTP (splithttp)** inbound, every accepted connection's `LocalAddr` is set to `h.localAddr`, which is the **listener** address (`l.listener.Addr()`). On a wildcard listener that is the unspecified address (`[::]` or `0.0.0.0`).

Consequences on a multi-IP host:

- `sendThrough: origin` reads the wildcard `[::]` for every connection, so all egress is bound to a single (often IPv6) source address regardless of which local IP the client actually connected to — source-in-source-out silently stops working.
- On hosts whose `[::]` has no route, the outbound fails entirely: the REALITY handshake and VLESS tunnel come up, but no data flows (`network is unreachable`).

TCP inbounds are unaffected because `worker.go` already uses the real `conn.LocalAddr()`; only XHTTP substitutes the listener address.

### Fix

Read the concrete per-connection local address from the request context via `http.LocalAddrContextKey`, which `net/http` populates with the address the client actually connected to. Fall back to the listener address when the key is absent (e.g. HTTP/3, where `net/http` does not set it), preserving current behavior there.

### Verification

Confirmed under HTTP/2 on a wildcard `[::]` listener that `request.Context().Value(http.LocalAddrContextKey)` returns the concrete IP the client connected to (different entry IP → different value), whereas `l.listener.Addr()` is always `[::]`.

End-to-end on a 5-IP host running `sendThrough: origin` + VLESS/REALITY/XHTTP: before the patch all egress bound `[::]` (broken on a host without IPv6 route); after the patch each entry IP egresses via the matching source IP (`.38→.38`, `.49→.49`, …), all reachable.

### DIFF
--- a/transport/internet/splithttp/hub.go
+++ b/transport/internet/splithttp/hub.go
@@ -373,11 +373,23 @@ func (h *requestHandler) ServeHTTP(writer http.ResponseWriter, request *http.Req
 			Reader:         request.Body,
 			ResponseWriter: writer,
 		}
+		// Use the concrete local address this request actually arrived on,
+		// not the listener's wildcard address (e.g. "[::]" / "0.0.0.0"). Go's
+		// net/http stores the per-connection local address in the request
+		// context under LocalAddrContextKey. Without this, sendThrough "origin"
+		// (which derives the outbound gateway from inbound.Local) reads the
+		// wildcard and pins every connection's egress to a single address,
+		// breaking source-in-source-out on multi-IP hosts. Fall back to the
+		// listener address when the key is absent (e.g. HTTP/3).
+		localAddr := h.localAddr
+		if la, ok := request.Context().Value(http.LocalAddrContextKey).(net.Addr); ok && la != nil {
+			localAddr = la
+		}
 		conn := splitConn{
 			writer:     httpSC,
 			reader:     httpSC,
 			remoteAddr: remoteAddr,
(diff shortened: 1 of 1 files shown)


Fangliding:
哎 小修复

RPRX:
@Fangliding 能合吗

Fangliding:
似乎可可
这种东西不知道为什么golang不放request里要放value ctx有点迷惑

RPRX:
其它传输层没这问题吗

RPRX:
@Fangliding

RPRX:
~~风扇旅游去了吗~~

Fangliding:
有时候上午爬起来看一堆乱七八糟的平台消息然后吃东西去了就忘了

至于其他传输 只有从 流/io对 重组出来的conn会丢 ws和httpupgrade都是原生listener返回的原本的conn grpc也是h2 based也会丢 不过应该找一下

Fangliding:
grpc确实有 我改一下吧

echoowall:
哦伤心，没蹭上contributors

PR #6483, opened 2026-07-12
Fix cross-instance system dialer state corruption

eliotcougar:
## Summary

This PR fixes process-global system-dialer state that can be corrupted when an embedding application creates multiple Xray core instances in the same process.

The scope has been deliberately reduced from the original version of this PR. All Observatory update-notification and probe-deadline changes have been removed. The ongoing Observatory work, including APIs intended for outbound probing by third-party embedders, will be presented separately.

This PR now addresses only the cross-instance DNS client and outbound-manager corruption.

## Problem

Some embedding applications currently create more than one `core.Instance` in the same process.

For example, v2rayNG may have:

1. A long-lived core carrying VPN traffic.
2. One or more short-lived cores performing outbound-delay or policy-group tests.

Before this change, construction of every core called `InitSystemDialer`, replacing package-level references to the DNS client and outbound manager.

A temporary test core could therefore overwrite the system-dialer dependencies belonging to the live core. After that temporary core was closed, the live core could continue trying to resolve domains through a closed DNS client or use an outbound manager belonging to the wrong instance.

This can produce partial and inconsistent failures:

- proxied or direct connections may stop working;
- some destinations may continue working while others fail;
- a newly created test core may still pass its own connection tests;
- restarting the live core restores normal operation.

## Root cause

The system dialer stored its DNS client and outbound manager in process-global variables even though those dependencies belong to a particular `core.Instance`.

Internal operations using those globals included:

- system-dialer DNS resolution;
- `dialerProxy` outbound-manager lookup;
- freedom outbound domain resolution;
- freedom UDP packet-writer resolution; and
- proxyman outbound target resolution.

Creating another core replaced the globals for every instance in the process.

## Changes

- Attach a lazy system-dialer dependency provider to each core instance context.
- Add `LookupForIPWithContext` for instance-aware DNS resolution.
- Resolve `dialerProxy` through the outbound manager associated with the current context.
- Propagate instance context through freedom, proxyman, system-dialer, and freedom UDP resolution paths.
- Synchronize the process-wide compatibility fallback.
- Preserve `InitSystemDialer` and `LookupForIP` for callers without an instance context.
- Continue initializing that compatibility fallback during ordinary core construction.
- Add regression coverage demonstrating that closing a temporary instance does not corrupt lookups performed by another instance.
- Add regression coverage for legacy fallback initialization through `core.New`.

## Working principle

Instance-owned internal paths now obtain their DNS client and outbound manager from the current Xray context.

The context contains a lazy provider rather than eagerly captured feature references because contexts are created while an instance may still be under construction. The provider resolves the current DNS client and outbound manager from that specific instance when they are needed.

Starting another core may still update the process-wide compatibility fallback, but it no longer changes the dependencies used by context-aware paths belonging to an existing instance.

Legacy callers without an instance context continue to use the synchronized process-wide fallback.

## Scope and compatibility

- No configuration schema changes.
- Existing `InitSystemDialer` and `LookupForIP` APIs remain available.
- Ordinary single-instance behavior is unchanged.
- This PR does not attempt to define broader lifecycle guarantees for running multiple complete Xray servers in one process.
- It narrowly prevents instance-owned internal dialer paths from being redirected to dependencies belonging to another in-process instance.

## Validation

The affected packages pass:

```text
go test -count=1 \
  ./core \
  ./transport/internet \
  ./app/proxyman/outbound \
  ./proxy/freedom
```

### DIFF
--- a/app/observatory/burst/burstobserver.go
+++ b/app/observatory/burst/burstobserver.go
@@ -3,6 +3,7 @@ package burst
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/xtls/xray-core/app/observatory"
 	"github.com/xtls/xray-core/common"
@@ -21,6 +22,7 @@ type Observer struct {
 
 	statusLock sync.Mutex
 	hp         *HealthPing
+	updates    extension.ObservatoryUpdateDispatcher
 
 	finished *done.Instance
 
@@ -29,6 +31,21 @@ type Observer struct {
 
 func (o *Observer) GetObservation(ctx context.Context) (proto.Message, error) {
 	return &observatory.ObservationResult{Status: o.createResult()}, nil
+}
+
+func (o *Observer) SubscribeObservationUpdates(listener func()) func() {
+	return o.updates.SubscribeObservationUpdates(listener)
+}
+
+func (o *Observer) ObservationProbeDeadline() time.Duration {
+	if o.hp == nil || o.hp.Settings == nil || o.hp.Settings.Timeout <= 0 {
+		return 0
+	}
+	deadline := o.hp.Settings.Timeout
+	if o.hp.Settings.Connectivity != "" {
+		deadline += o.hp.Settings.Timeout
+	}
+	return deadline
 }
 
 func (o *Observer) Check(tag []string) {
@@ -100,12 +117,14 @@ func New(ctx context.Context, config *Config) (*Observer, error) {
 		return nil, errors.New("Cannot get depended features").Base(err)
 	}
 	hp := NewHealthPing(ctx, dispatcher, config.PingConfig)
-	return &Observer{
+	observer := &Observer{
 		config: config,
 		ctx:    ctx,
 		ohm:    outboundManager,
 		hp:     hp,
-	}, nil
+	}
+	hp.onUpdate = observer.updates.NotifyObservationUpdate
+	return observer, nil
 }
 
 func init() {
--- a/app/observatory/burst/healthping.go
+++ b/app/observatory/burst/healthping.go
@@ -34,6 +34,7 @@ type HealthPing struct {
 
 	Settings *HealthPingSettings
 	Results  map[string]*HealthPingRTTS
+	onUpdate func()
 }
 
 // NewHealthPing creates a new HealthPing with settings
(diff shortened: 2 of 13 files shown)


Fangliding:
> This PR fixes process-global system-dialer state that can be corrupted when multiple Xray instances run in the same process,

这不是设计考虑

> Server is an instance of Xray. At any time, there must be at most one Server instance running.

https://github.com/XTLS/Xray-core/blob/50231eaff98ccc31b5cbd247a721c16e97fe5ec1/core/xray.go#L23

eliotcougar:
Thank you for pointing to that comment. I understand it as expressing a design intention that no more than one `core.Instance` should run in a process.

However, the situation addressed by this PR is not hypothetical behavior introduced by the PR. It is how the current v2rayNG and AndroidLibXrayLite code already operate.

In current upstream v2rayNG:

- `CoreVpnService`, which owns the live core, and `CoreTestService`, which performs bulk real-ping tests, both run in the Android process `:RunSoLibV2RayDaemon`.
- The live connection is held as a long-lived `*core.Instance` by `CoreController`.
- Bulk real-ping starts `CoreTestService` without stopping the live connection.
- `RealPingWorkerService` runs tests with configurable concurrency.
- Every test calls AndroidLibXrayLite’s `MeasureOutboundDelay`, which calls `core.New`, `Instance.Start`, and later `Instance.Close`.

Therefore, current v2rayNG can run one or more short-lived `*core.Instance` objects concurrently with the live `*core.Instance` in the same process. These are the same type identified by `ServerType()`.

Xray currently permits this at runtime: `core.New` and `Instance.Start` do not enforce a process-wide singleton. Instead, every instance construction calls `InitSystemDialer`, replacing the package-global DNS client and outbound manager. When a temporary instance is created and then closed, the live instance can consequently be left using dependencies belonging to the closed temporary instance. That is the concrete failure fixed by this PR.

If multiple instances in one process are categorically unsupported, I will try to implement an alternative solution. However, other implementations can also act like this. Leaving the current API and downstream implementation in place means this corruption remains possible.

The patch preserves ordinary single-instance behavior unchanged.

Would you be willing to reopen the PR for this technical discussion?

eliotcougar:
I have reduced the scope of this PR to the cross-instance system-dialer state-corruption fix only. All Observatory-related changes have been removed from the branch. The broader Observatory and one-shot outbound-probing work required for proper implementation of bulk probing without spawning separate per-outbound processes is still ongoing and will be presented as a separate PR.

If appropriate, please reopen this PR for review in its reduced form.

PR #6493, opened 2026-07-14
Sockopt: Add sendProxyProtocol for TCP outbounds

slashfast:
When you chain proxies (inbound → outbound → upstream), the real client IP is lost after the first hop. Every intermediary and the final backend only see the previous node's IP. IP-based logging becomes useless after the first hop — every node in the chain logs the previous node's IP instead of the real client.

acceptProxyProtocol (receiving PROXY headers from upstream) has existed since v1.0.0. sendProxyProtocol (writing them downstream) was never implemented at the transport layer. Freedom added its own proxy protocol support in [3a99520](https://github.com/XTLS/Xray-core/commit/3a995203), but only for direct TCP — VLESS, VMess, Trojan and other protocols were left out.

This adds sockopt.sendProxyProtocol (0|1|2), which writes a HAProxy PROXY protocol v1 or v2 header on every outbound TCP connection, at the transport layer (after dial, before TLS), regardless of the proxy protocol.

Config:

```json
"streamSettings": {
  "sockopt": {
    "sendProxyProtocol": 1
  }
}
```
All five TCP dialers (tcp, ws, grpc, httpupgrade, splithttp) write the header. Freedom outbound blocks the sockopt field and redirects to its own settings.proxyProtocol instead. Freedom's implementation was deduplicated — it now calls the same WriteProxyProtocolHeader as the dialers, removing the go-proxyproto import from freedom.go.

```bash
go test ./transport/internet -run TestWriteOutboundProxyProtocol -count=1
go test ./infra/conf -run TestFreedomOutboundRejects -count=1
```

Running on a [remnanode](https://github.com/remnawave/node) (VLESS+TCP+REALITY, sendProxyProtocol:2) with real users for a month — stable. Client IP visible in the [remnawave panel](https://github.com/remnawave/panel) where only node IP was visible before.

### DIFF
--- a/infra/conf/freedom_test.go
+++ b/infra/conf/freedom_test.go
@@ -1,6 +1,7 @@
 package conf_test
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/xtls/xray-core/common/geodata"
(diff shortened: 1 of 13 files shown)


Fangliding:
可以使用dialer proxy到freedom以发送proxy protocol 毕竟除了freedom xray的出站协议通常设计用来穿过防火墙 携带proxy protocol不太常见 这个实现也不太对

slashfast:
哈哈，确实，我应该更关注项目架构。dialerProxy 到 freedom 就行了，感谢指出。

PR #6494, opened 2026-07-14
Stats: Track online users per inbound tag

slashfast:
Online users are tracked per-user globally (`user>>>{email}>>>online`) but there's no way to see which inbound they came from. Adding a per-inbound key (`inbound>>>{tag}>>>user>>>{email}>>>online`) alongside the existing global one and wiring VisitOnlineMaps into QueryStats makes both queryable.

```bash
go test ./app/dispatcher ./app/stats/command -count=1


Fangliding:
api没有为 per inbound per user 设计 不然各种东西都要加又是一堆麻烦 也许你可以直接在不同入站用不同user

PR #6496, opened 2026-07-15
Add per-user bandwidth limits

Lagnuty:
## Summary

This PR adds optional per-user bandwidth limits to `protocol.User` and applies them in the dispatcher.

New optional user fields:

- `speed_limit_up_mbps` / `speedLimitUpMbps`
- `speed_limit_down_mbps` / `speedLimitDownMbps`

A value of `0` keeps the current unlimited behavior.

## Behavior

- Limits are applied per user email and per direction (`uplink` / `downlink`).
- Concurrent connections for the same user share the same limiter, so the configured value is a per-user aggregate limit rather than a per-connection limit.
- Different users in the same inbound get separate limiters.
- Existing configs are unaffected because the new fields default to zero.

## Implementation notes

- The fields are added to `common/protocol/user.proto` and propagated to `MemoryUser`.
- Dispatcher writers/readers are wrapped with a token-bucket limiter only when the corresponding user limit is non-zero.
- `UnmarshalJSON` accepts camelCase keys as used by JSON config/frontends while keeping the generated protobuf snake_case JSON tags intact.
- `common/protocol/user.pb.go` was regenerated with `protoc 33.5`, `protoc-gen-go v1.36.11`, and `protoc-gen-go-grpc v1.6.0`.

## Tests

- `go test ./common/protocol ./app/dispatcher`
- `go build -o <temp> ./main`

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -182,6 +182,9 @@ func (d *DefaultDispatcher) getLink(ctx context.Context) (*transport.Link, *tran
 		if p.Stats.UserOnline {
 			trackOnlineIP(ctx, d.stats, user.Email, sessionInbound.Source.Address.String())
 		}
+
+		inboundLink.Writer = NewUserRateLimitWriter(ctx, user.Email, "uplink", user.SpeedLimitUpMbps, inboundLink.Writer)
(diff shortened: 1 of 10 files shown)


Fangliding:
不打算加入限速功能

PR #6506, opened 2026-07-16
fix: header/custom UDP mask - orphaned handshake waiter + unsynchronized vars field

hexonal:
## Summary

Two independent concurrency bugs in `transport/internet/finalmask/header/custom/udp.go`.

### 1. Orphaned handshake waiter

`udpCustomStandaloneClientConn.registerWaiter` unconditionally overwrote `c.wait[key]`:

```go
func (c *udpCustomStandaloneClientConn) registerWaiter(key string, vars map[string][]byte) *udpStandaloneWaiter {
	waiter := &udpStandaloneWaiter{...}
	c.mu.Lock()
	c.wait[key] = waiter   // <-- overwrites any existing entry
	c.mu.Unlock()
	return waiter
}
```

When two goroutines both call `WriteTo()` to the same not-yet-established remote `addr` before the handshake completes — the normal shape of many client sessions sharing one outbound UDP socket, e.g. VMess/VLESS UDP-over-one-socket — the second call's waiter silently replaces the first's in the map. The first caller then blocks forever on `<-waiter.done`. Even a connection read error (`failWaiters`) doesn't help, since it only drains whatever's *currently* in the map (the second waiter) — the first, now-orphaned waiter is unreachable and never woken.

**Fix:** make registration get-or-join (`registerOrJoinWaiter`): if a waiter for this key is already in flight, join it instead of creating (and silently losing) a new one, and don't resend the handshake request in that case. Also switched the waiter's completion signal from a single-buffered `chan error` (which can only ever deliver to one of N joiners) to a `close()`-based `chan struct{}` + `err` field, so every concurrent joiner wakes up when the handshake resolves, not just whichever one wins the race on a one-shot channel send.

### 2. Unsynchronized `vars` field

`udpCustomClient`/`udpCustomServer.vars` is written by `Match()` (called from the `ReadFrom` fan-in path) and read by `WriteTo()` (called concurrently by however many client sessions share the outbound `PacketConn`) with no synchronization at all — unlike the adjacent `stateStore`, which already has its own mutex for exactly this reason. Added a mutex guarding this field specifically, with a `currentVars()` accessor for the read side.

## Testing

Added `TestRegisterOrJoinWaiterFansOutToAllConcurrentCallers`: 32 concurrent goroutines register/join on the same key, exactly one is told to send the handshake request, and all 32 wake up once it completes. Passes clean under `-race`. Full existing `header/custom` test suite passes unchanged under `-race`.

### DIFF
--- a/transport/internet/finalmask/header/custom/udp.go
+++ b/transport/internet/finalmask/header/custom/udp.go
@@ -20,7 +20,21 @@ type udpCustomClient struct {
 	merged []byte
 	read   int
 	state  *stateStore
-	vars   map[string][]byte
+	// mu guards vars: Match() writes it from the ReadFrom fan-in loop while
+	// WriteTo (called concurrently by many client sessions sharing this one
+	// outbound PacketConn) reads it as a fallback when no per-addr state
+	// exists yet - the adjacent stateStore already has its own lock for
+	// exactly this reason, but this plain field didn't.
+	mu   sync.Mutex
+	vars map[string][]byte
+}
+
+// currentVars returns a defensive copy of the last-matched vars, safe to
+// call concurrently with Match().
+func (h *udpCustomClient) currentVars() map[string][]byte {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	return cloneVars(h.vars)
 }
 
 func (h *udpCustomClient) Serialize(b []byte) {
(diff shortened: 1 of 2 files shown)


Fangliding:
这狗屎vibe出来的功能 找时间删了吧

PR #6547, opened 2026-07-27
feat(splithttp): add hMinSocketInterval to XmuxConfig

neveraboutpast:
### Description
A new field `hMinSocketInterval` (type `Int32Range`, measured in milliseconds) has been added to the `XmuxConfig` configuration for the `SplitHTTP` transport (also known as `XHTTP`).

This option sets the minimum time interval between establishing new physical TCP/TLS connections (sockets) within XMUX multiplexing. If application logic requests opening multiple physical connections in rapid succession over a short period of time, XMUX pauses for a duration corresponding to a value from the `hMinSocketInterval` range before initiating the next connection.

### Motivation
According to an analysis of behavioral blocks (specifically RKN's "Siberian" heuristic as of June 2026), DPI systems have actively started utilizing external connection timings to detect circumvention tools:
* One of the key behavioral signatures of proxy activity is creating **more than 3 parallel TLS connections to the same SNI with an interval of less than 350-400 ms between them within a 60-second window**.
* Upon matching this pattern (along with a suspicious hosting subnet and a popular browser TLS fingerprint), the connection undergoes degradation ("freezing") for 120 seconds.

Although multiplexing (XMUX) significantly reduces the number of parallel connections, during peak web activity or when opening heavy resources, XMUX may still initiate new physical connections too rapidly.

The `hMinSocketInterval` parameter allows artificially spacing out connection establishment times by a safe margin (e.g., `400-600` ms), reducing the risk of triggering censor behavioral heuristics.

### Technical Implementation Details
1. **`transport/internet/splithttp/config.proto`**: Added field `RangeConfig hMinSocketInterval = 7;` to the `XmuxConfig` message.
2. **`transport/internet/splithttp/config.go`**: Added helper method `GetNormalizedHMinSocketInterval()` for safely retrieving the interval (defaults to range `0-0`).
3. **`infra/conf/transport_method.go`**: Defined field `HMinSocketInterval Int32Range` in the `XmuxConfig` JSON structure.
4. **`transport/internet/splithttp/xmux_manager.go`**:
   * Added fields `minSocketInterval time.Duration`, `lastSocketCreateAt time.Time`, and mutex `mu sync.Mutex` to the `XmuxManager` struct.
   * During manager initialization, the `minSocketInterval` duration is randomized from the specified range.
   * Added logic to track the last socket creation time to ensure the minimum delay is respected for subsequent connections.
5. **`transport/internet/splithttp/dialer.go`**: Moved the `globalDialerAccess.Unlock()` call before `GetXmuxClient` to prevent unnecessarily holding the global dialer map mutex during the interval delay.

### Configuration Example (Client)
To configure, simply add the `hMinSocketInterval` parameter to the `xmux` block inside `xhttpSettings` (or `splithttpSettings`):

```json
{
  "mode": "stream-one",
  "xPaddingBytes": "50-1200",
  "xPaddingObfsMode": true,
  "xmux": {
    "cMaxReuseTimes": "30-90",
    "hKeepAlivePeriod": 10,
    "hMaxRequestTimes": "40-100",
    "hMaxReusableSecs": "30-50",
    "maxConcurrency": "8-16",
    "maxConnections": 0,
    "hMinSocketInterval": "400-600", // socket delay range in milliseconds
  }
}
```


Fangliding:
只有几百毫秒 调高 maxConcurrency 应该就行了吧 主要是东西一多
看了下实现 time.Sleep(sleepDur) 太不合适了

PR #6548, opened 2026-07-27
fix(tun): add default IPv4 gateway and split-horizon routes for Linux

neveraboutpast:
## Description

When `autoSystemRoutingTable` is empty (which is the common case with v2rayN on Linux), the TUN interface was brought up with an IPv4 address but **without any routes**. Traffic never entered the TUN stack, making TUN mode non-functional on Linux.

### Changes

- **Default gateway fallback**: Added `defaultLinuxGateway = "198.18.0.1/15"` — if `gateway` is not configured in the TUN config, Xray-core now assigns this address to the TUN interface, matching the existing macOS behavior (`defaultDarwinGateway = "169.254.10.1/30"`)
- **Split-horizon routes**: When `autoSystemRoutingTable` is empty, Xray-core now adds `0.0.0.0/1` and `128.0.0.0/1` routes through the TUN interface. These two routes cover the entire IPv4 space without replacing the system's existing default gateway, avoiding network loops
- **Cleanup**: Routes and interface address are properly removed on shutdown via the existing `unsetSystemRoutes()` / `unsetInterfaceAddresses()` path

### Why split-horizon?

Using `0.0.0.0/1` + `128.0.0.0/1` instead of `0.0.0.0/0` ensures the system default gateway remains intact. Xray's own outbound connections to upstream servers still use the physical interface, preventing the infinite network loop described in the TUN README.

### Testing

- `go build ./proxy/tun/` — compiles cleanly
- `go test ./proxy/tun/ -v` — all tests pass (including 7 new tests for `selectLinuxGateway()`)
- Verified with v2rayN v7.24.3 on Linux — TUN mode now works correctly

https://github.com/2dust/v2rayN/issues/9755

### DIFF
--- a/proxy/tun/tun_linux.go
+++ b/proxy/tun/tun_linux.go
@@ -17,6 +17,8 @@ import (
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 )
 
+const defaultLinuxGateway = "198.18.0.1/15"
+
 // LinuxTun is an object that handles tun network interface on linux
 // current version is heavily stripped to do nothing more,
 // then create a network interface, to be provided as file descriptor to gVisor ip stack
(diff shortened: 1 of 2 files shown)


Fangliding:
autoSystemRoutingTable 为空就是不启用啊 用户自己写路由表或者 bind interface

PR #6553, opened 2026-07-29
Routing: Add `localOS` that directly matches `runtime.GOOS`

fanymagnet:
Adds `localOS` to routing rules: the rule only applies when Xray itself runs on
one of the listed operating systems. Values are `runtime.GOOS` (`android`, `ios`,
`windows`, `darwin`, `linux`, ...), compared case-insensitively. Since the OS
never changes at runtime, the result is resolved once when the rule is built, so
there is no cost on the hot path.

**Why:** clients commonly fetch a single config from a URL, and the same config
is shared across Android, iOS, Windows and desktop. Some rules are inherently
platform-specific — `process` values are package names on Android, and a `tun`
inbound only exists on mobile — so without a guard they either silently never
match or, in the case of a catch-all, do the wrong thing everywhere else.

In the example below the second rule blocks everything from `tun-in` that is not
an allowed app. Both rules are scoped to Android, so on Windows or macOS the
same shared config simply falls through them instead of blocking all traffic.

```json
{
  "inbounds": [
    {
      "protocol": "tun",
      "tag": "tun-in"
    }
  ],
  "outbounds": [
    {
      "protocol": "freedom",
      "tag": "direct-out"
    },
    {
      "protocol": "loopback",
      "settings": {
        "inboundTag": "allowed-apps-in"
      },
      "tag": "allowed-apps-out"
    },
    {
      "protocol": "blackhole",
      "tag": "block-out"
    }
  ],
  "routing": {
    "rules": [
      {
        "inboundTag": ["tun-in"],
        "process": [
          "com.google.android.youtube",
          "com.android.chrome"
        ],
        "localOS": ["android"],
        "outboundTag": "allowed-apps-out"
      },
      {
        "inboundTag": ["tun-in"],
        "localOS": ["android"],
        "outboundTag": "block-out"
      }
    ]
  }
}

### DIFF
--- a/app/router/condition.go
+++ b/app/router/condition.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"slices"
 	"strings"
 
@@ -392,4 +393,23 @@ func (m *ProcessNameMatcher) Apply(ctx routing.Context) bool {
 		}
 	}
 	return false
+}
+
+// LocalOSMatcher matches the operating system Xray itself is running on. That never
+// changes while Xray is running, so the result is resolved when the rule is built.
+type LocalOSMatcher struct {
+	matched bool
+}
+
+func NewLocalOSMatcher(names []string) *LocalOSMatcher {
+	return &LocalOSMatcher{
+		matched: slices.ContainsFunc(names, func(name string) bool {
+			return strings.EqualFold(name, runtime.GOOS)
+		}),
+	}
+}
+
+// Apply implements Condition.
+func (m *LocalOSMatcher) Apply(_ routing.Context) bool {
+	return m.matched
 }
--- a/app/router/condition_test.go
+++ b/app/router/condition_test.go
@@ -2,7 +2,9 @@ package router_test
 
 import (
 	"path/filepath"
+	"runtime"
 	"strconv"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


patterniha:
This is exactly what I need for "Serverless-for-Iran" (because fragment behaves differently on different OSes), ~I was thinking of adding it.~  Thanks.

vo0ov:
Great idea! This would be a really useful addition for shared configs across different platforms. Hope it gets merged.

RPRX:
感觉 GOOS 在匹配 `darwin`/`ios` 上有点混乱比如 https://github.com/XTLS/Xray-core/pull/6524 ，~~不过似乎只影响开发者，且无关这个 PR 毕竟单值肯定取最精确的~~

RPRX:
~~以后真有需要的话可以水一个 PR 加上 localARCH 等别的东西~~

RPRX:
@Fangliding 写文档时要说明 `localOS` 既然是数组就应该写最精确的，比如 darwin 不含 ios，要同时匹配上两个的话需要都写

Fangliding:
神秘goos逻辑

RPRX:
可能是因为 iOS 不像 Android 上还有可能跑裸 bin 所以 `go build` 那里没分配单独的 `iOS`，但又有 gomobile，~~总之有点混乱~~

~~不过 `// go:build` 那里写 linux 似乎也能作用到 android 来着？~~

Issue #6554, opened 2026-07-29
XHTTP packet-up: idle download stream severed by CDN read-timeout during long uploads — server keepalive (scStreamUpServerSecs) covers stream-up only


twilgate:
### Summary

In `packet-up`, a sustained **upload** leaves the long-lived download GET idle. A fronting CDN / reverse proxy (nginx `proxy_read_timeout` 60s; Cloudflare ~100s) closes a response that carries no bytes for that long. The download stream is opened once, has **no periodic keepalive and no client-side recovery**, so the proxied connection stalls until the whole session is torn down (redial / close-all-connections).

The existing server keepalive from PR #4306 (`scStreamUpServerSecs`) is wired only into the **stream-up** path, so packet-up remains exposed. This is the packet-up counterpart of the exact problem #4306 solved for stream-up. (References below are to `v26.7.28`.)

### Where the keepalive is, and why packet-up misses it

Server keepalive lives in the **upload (POST) handler** and writes padding into the *upload response* — `hub.go`:

```go
scStreamUpServerSecs := h.config.GetNormalizedScStreamUpServerSecs()
hasLegacyRefererCompatMarker := request.Header.Get("Referer") != ""
if (hasLegacyRefererCompatMarker || obfsPaddingAccepted) && scStreamUpServerSecs.To > 0 {
    go func() {
        for {
            _, err := httpSC.Write(bytes.Repeat([]byte{'X'}, int(h.config.GetNormalizedXPaddingBytes().rand())))
            if err != nil { break }
            time.Sleep(time.Duration(scStreamUpServerSecs.rand()) * time.Second)
        }
    }()
}
```

This keeps the **upload** response alive for stream-up, and the client can discard it wholesale (the upload response body carries no application data). It does nothing for packet-up, where:

- upload = many short POSTs (each round-trips quickly, never idle long enough to be cut),
- download = one long-lived GET whose response **is** the proxied data stream, and which goes silent for the whole duration of an upload.

Client opens that download GET exactly once, with no re-open path — `dialer.go`:

```go
} else { // stream-down
    if xmuxClient2 != nil {
        xmuxClient2.LeftRequests.Add(-1)
    }
    conn.reader, conn.remoteAddr, conn.localAddr, err = httpClient2.OpenStream(ctx, requestURL2.String(), sessionId, nil, false)
```

[...]

PR #6561, opened 2026-07-31
Mux.Cool: actually send SessionStatusKeepAlive (0x04)

twilgate:
### What

`SessionStatusKeepAlive (0x04)` has been part of the Mux.Cool frame vocabulary since the protocol was introduced: it is defined in `common/mux/frame.go`, handled by the client (`handleStatueKeepAlive` — the typo itself dates back to v2ray days) and by the server (`handleStatusKeepAlive`). But a `grep -rn SessionStatusKeepAlive` over the tree yields exactly three hits — the definition and the two receivers. **No code path has ever emitted this frame.** An idle Mux connection is therefore idle on the wire.

This PR adds the missing emitter on the server side: while at least one session is active and the downlink has been idle for the keepalive interval, `ServerWorker` writes a minimal 6-byte `0x04` frame (no options, no payload).

### Why

In #6554 (XHTTP packet-up: download stream cut by middleboxes after idle) the suggested remedy was to enable Mux. With the current code that advice cannot work — nothing keeps the connection non-idle. This PR makes it work.

### Compatibility / risk

- **Zero wire risk.** Every existing client already discards a `0x04` frame without `OptionData` silently (`client.go` `handleStatueKeepAlive` returns `nil`). Covered by a test that runs an unmodified `ClientWorker` against the emitting server.
- **Server-only deploy fixes old clients** — no client update needed.
- Frames are only sent while sessions exist and only when the downlink is idle (a wrapper writer tracks the last write time), so active traffic and connection reaping (`CloseIfNoSessionAndIdle`) are unaffected.
- Interval: `xray.mux.server.keepalive` env flag (`XRAY_MUX_SERVER_KEEPALIVE`), seconds; default 15, `0` disables. Happy to flip the default to opt-in if preferred.

Tests: emitter fires on idle, stays silent under active traffic, unmodified client transparency. `go test -race ./common/mux/...` green.

Closes the gap behind #6554.

### 中文摘要

`SessionStatusKeepAlive (0x04)` 自协议诞生起就有定义和接收处理（客户端与服务端都会静默丢弃该帧），但**从未有任何代码发送过它**——空闲的 Mux 连接在线路上就是空闲的，中间设备照样掐断空闲下行（见 #6554，当时建议"开启 Mux"来保活，但现有代码做不到）。本 PR 补上缺失的发送端：服务端在有活跃会话且下行空闲达到间隔时，发送最小 6 字节的 0x04 帧。现有客户端本来就会静默丢弃该帧，因此**零线路兼容风险**，且**仅升级服务端即可修复老客户端**。间隔通过环境变量 `XRAY_MUX_SERVER_KEEPALIVE`（秒）配置，默认 15，0 为关闭。

### DIFF
--- a/common/mux/keepalive_test.go
+++ b/common/mux/keepalive_test.go
@@ -1,0 +1,238 @@
+package mux_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/xtls/xray-core/common"
+	"github.com/xtls/xray-core/common/buf"
+	"github.com/xtls/xray-core/common/mux"
+	"github.com/xtls/xray-core/common/net"
+	"github.com/xtls/xray-core/common/protocol"
+	"github.com/xtls/xray-core/common/session"
+	"github.com/xtls/xray-core/transport"
+)
+
+// readOneFrame reads a single mux frame metadata from the reader, draining
+// the data chunk if the frame carries one.
+func readOneFrame(reader *buf.BufferedReader) (*mux.FrameMetadata, error) {
+	var meta mux.FrameMetadata
+	if err := meta.Unmarshal(reader, false); err != nil {
+		return nil, err
+	}
+	if meta.Option.Has(mux.OptionData) {
+		if err := buf.Copy(mux.NewStreamReader(reader), buf.Discard); err != nil {
+			return nil, err
+		}
+	}
+	return &meta, nil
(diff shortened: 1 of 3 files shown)


Fangliding:
你这种改法不对 而且发送时机需要考虑

PR #6562, opened 2026-07-31
XHTTP: keep the stream-down response alive (completes #4306)

twilgate:
## Summary

This completes the keepalive work started in #4306. That PR keeps the
**stream-up** (upload) response alive so a fronting CDN does not cut it. But the
long-lived **stream-down** GET — the single download response shared by both
`packet-up` and `stream-up` — still has no keepalive. When data flows upward
(bulk upload) the download half legitimately goes silent, and a fronting proxy
with a response idle-read timeout tears it down. The session then hangs and does
not recover.

This is not a new feature; it is the other half of #4306.

## Why this is not a rare corner case

- nginx `proxy_read_timeout` defaults to **60s**.
- Cloudflare's edge idle timeout is **~100s** and is not configurable below
  Enterprise.
- `mode: auto` resolves to `packet-up` for everything that is not REALITY, so
  **every CDN-fronted XHTTP deployment** is exposed, and `stream-up` shares the
  same download GET, so it is exposed too.

Raising the origin's own `proxy_read_timeout` does not help — the cut happens at
the CDN edge, upstream of the origin.

## Proof (the important part)

`Test_XHTTP_StreamDownKeepalive_Repro` stands up a real splithttp server behind a
proxy that cuts the connection when the upstream is silent for `idle`
(modelling the CDN). A stream-down half that stays silent past `idle`:

- **on unpatched `main`**: is torn down — the test fails with `unexpected EOF` at
  the idle cut (verified: RED, cut at 1.50s).
- **with this PR**: survives — padding keepalive records hold the half open and
  the payload arrives (verified: GREEN at 3.00s).

`Test_XHTTP_StreamDownFraming_Compat` covers all four cells of the
old/new-client × old/new-server matrix; every cell delivers the bytes intact.

## Design

Length-prefixed records on the stream-down body, negotiated both ways so the
wire is unchanged unless both peers opt in.

**Record format**

```
record := uvarint(hdr) || payload
hdr    := (len << 1) | kind        // kind: 0 = data, 1 = padding(keepalive)
```

Overhead is ~0.02% on a typical 16–64 KiB record. Keepalive records are
**padding** records with random content (not zero-length), so no periodic 0x00
marker appears in the plaintext for an edge terminating TLS to key on. A
`maxFrameRecord` guard (2 MiB) makes a desync fail loudly instead of silently
corrupting the tunnel.

**Negotiation / backward-compat**

1. A client that set `downFrame` adds a meta marker (`downFrameKey`, default
   `x_df`) to the download GET.
2. The server frames only if the marker is present **and** its own
   `scStreamDownServerSecs > 0`.
3. The server confirms with the `X-Down-Frame: 1` response header; the client
   strips framing only after seeing that confirmation.

| client        | server        | result                                   |
|---------------|---------------|------------------------------------------|
| old           | new (on)      | no marker → no framing → wire as today   |
| new (on)      | old           | no confirm → raw stream                  |
| new (on)      | new (on)      | framing + keepalive                      |
| new (off)     | new (on)      | no marker → wire as today                |

Byte-for-byte identical to today until **both** sides opt in.

**Defaults / scope**

- `scStreamDownServerSecs` defaults to `0` = disabled (opt-in).
- `stream-one` is never framed (no session; REALITY path, not fronted by a CDN).
- Browser dialer never sends the marker.
- H3 uses the same handler and needs no special handling.
- `x_version` is left as-is (deliberately commented out; client-version leak /
  active-probing risk).

## Implementation notes

`httpServerConn.Write` holds its mutex and flushes on every call, so the framer
writes each `hdr || payload` in a **single** underlying `Write` from a pooled
buffer. Otherwise the keepalive goroutine could interleave inside a record, and
on HTTP/1.1 one record would split across two chunks.

## Commits

1. config plumbing (`scStreamDownServerSecs`, `downFrameKey`, `downFrame`)
2. record codec + framer + unit tests
3. server keepalive (hub.go)
4. client negotiation + strip + repro/compat tests

## Alternative, if you prefer

If the confirm header is unwelcome, we can drop it and rely on pure
config-matching (as `noSSEHeader`/`mode` already do). We kept the confirm header
because with plain config-matching a client/server config drift would corrupt
the stream, whereas the confirmation makes a mismatch fall back to the raw
stream safely. Your call.

Related: closes the gap left open after #6554 (please do not reopen — linking for
context).

---

## 摘要（中文）

本 PR 补全 #4306。#4306 让 **stream-up**（上行）响应保持活跃，但长连接的
**stream-down** GET（`packet-up` 与 `stream-up` 共用的那个下行响应）仍然没有
keepalive。批量上传时下行半程会合理地保持静默，前置 CDN 的响应空闲超时
（nginx `proxy_read_timeout` 默认 60s，Cloudflare 边缘 ~100s 且低于企业版不可
调）就会切断它，会话随之挂死。`mode: auto` 对所有非 REALITY 都解析为
`packet-up`，因此每个经 CDN 的 XHTTP 部署都受影响。

方案：在 stream-down 响应体上使用长度前缀记录
（`record := uvarint((len<<1)|kind) || payload`，kind 0=数据、1=填充/保活），
并做双向协商——客户端用元数据标记（`downFrameKey`，默认 `x_df`）声明支持，
服务端仅在自身 `scStreamDownServerSecs > 0` 且收到标记时才分帧，并用
`X-Down-Frame: 1` 响应头确认；客户端只有看到确认才剥离分帧。默认关闭，
只有双方都开启时线路才改变，其余组合与现状逐字节一致。保活记录使用随机填充，
并设 2 MiB 上限，失步时显式报错而非静默损坏。PR 内含复现测试：静默的下行半程
在未打补丁的 main 上被切断（测试变红），打补丁后存活（变绿），另有四格
兼容性矩阵测试。

---

Related: #6554 (original report), #6561 (companion PR reviving the Mux.Cool keepalive path the maintainer suggested there).


Fangliding:
我都说了不会有帧格式 而且AI

PR #6582, opened 2026-08-02
Observatory: cancel health-check probes when they time out

embogomolov:
Two things let a probe outlive the round that started it.

`doCheck` takes a cancellable context and the scheduler cancels the previous round before starting the next one, but the probe client is still built from `h.ctx`. The machinery is there with nothing attached to it.

The probe transport also leaves `TLSHandshakeTimeout` unset, which Go reads as no limit. `http.Client.Timeout` ends the request, but the goroutine inside `addTLS` stays parked in the handshake, holding the dispatched outbound open.

Pass the round context to `newPingClient`, and give the handshake the same timeout as the probe that started it.

The test swaps in a `tagged.Dialer` that blocks until its context ends, runs a round, cancels it, and expects the dial to be cancelled too. It fails on main.

### DIFF
--- a/app/observatory/burst/healthping.go
+++ b/app/observatory/burst/healthping.go
@@ -172,7 +172,7 @@ func (h *HealthPing) doCheck(ctx context.Context, tags []string, duration time.D
 	for _, tag := range tags {
 		handler := tag
 		client := newPingClient(
-			h.ctx,
+			ctx,
 			h.dispatcher,
 			h.Settings.Destination,
 			h.Settings.Timeout,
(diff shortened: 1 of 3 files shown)


Fangliding:
也许应该用其他更直接的方法

embogomolov:
What would the more direct approach be?

This is two lines. `doCheck` already gets a cancellable context and the scheduler already cancels the previous round, but the probe client is built from `h.ctx`, so the cancellation never reaches the dial. The other line sets `TLSHandshakeTimeout`, which is currently unset and therefore unlimited.

PR #6583, opened 2026-08-02
Tagged dialer: release the dispatched outbound when the conn closes

embogomolov:
`DialTaggedOutbound` wraps the pipes from `dispatcher.Dispatch` in a `cnc.Connection`. Closing that connection interrupts the pipes but leaves the dispatch context alone, so the outbound handler keeps running until the caller's context ends.

Every call site in tree passes a context that outlives the connection: `observer.go` derives it from `o.ctx`, `geodata/download.go` uses the downloader's, `burst/ping.go` passes the one `newPingClient` received. None of them ends when the conn closes.

Interrupting the pipes does not reach an outbound that has not started reading them. Against a peer that completes the TCP handshake and then stalls, the dispatched goroutine sits in `tcp/dialer.go:81` under `retry.On` and never touches `link.Reader` or `link.Writer`. I measured it on v26.7.28 with #6582 applied, so that the probe does close its conn: after 150s of health check probes at a 15s interval, 22 of these goroutines were still parked, while the probe side had none.

Derive a cancellable context for the dispatch and cancel it from `ConnectionOnClose`. A conn that nobody closes behaves as it does today.

One correction to what I first wrote here. I said `ConnectionOnClose` had no users. It has six call sites. The nearest one is `redirect` in `transport/internet/dialer.go`, which dispatches with `context.WithoutCancel` and attaches pipe closers instead, so this patch does go against that precedent. Bounding the handshake inside `tcp/dialer.go` would be the other way to reach the same goroutine.

### DIFF
--- a/transport/internet/tagged/taggedimpl/impl.go
+++ b/transport/internet/tagged/taggedimpl/impl.go
@@ -22,8 +22,10 @@ func DialTaggedOutbound(ctx context.Context, dispatcher routing.Dispatcher, dest
 	ctx = session.ContextWithContent(ctx, content)
 	ctx = session.SetForcedOutboundTagToContext(ctx, tag)
 
+	ctx, cancel := context.WithCancel(ctx)
 	r, err := dispatcher.Dispatch(ctx, dest)
 	if err != nil {
+		cancel()
 		return nil, err
 	}
 	var readerOpt cnc.ConnectionOption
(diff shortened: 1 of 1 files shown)


Fangliding:
管道关闭后它应该会被两边的调度器杀死

embogomolov:
That holds for a session that reached the copy loop. `Close` closes the writer and interrupts the reader, so the outbound side gets `io.EOF` from `link.Reader` and `io.ErrClosedPipe` from `link.Writer`, and `Process` returns.

A dispatched outbound does not start there. It dials first:

```
routedDispatch
  proxyman/outbound.(*Handler).Dispatch
  vless/outbound.(*Handler).Process
  retry.(*retryer).On
  (*Handler).Dial -> internet.Dial -> transport dial
```

`vless/outbound` reaches `link.Reader` and `link.Writer` at line 312, well after the dial at the top of `Process`. A goroutine parked in that dial has never read or written the pipes, so interrupting them changes state that nothing is watching.

The dispatch context is the one signal that reaches this phase. `tcp/dialer.go` runs the handshake as `HandshakeContext(ctx)` or `reality.UClient(conn, config, ctx, dest)` and bounds it by nothing else. `net.Dialer.Timeout` in `DefaultSystemDialer.Dial` covers the TCP phase alone, so a peer that finishes the TCP handshake and then goes quiet holds that goroutine for as long as the context lives.

No call site in tree passes a context that ends when the conn closes. `observer.go` derives it from `o.ctx`, `geodata/download.go` from the downloader's. A caller that closes the conn on its own timeout and dials again keeps one parked goroutine per attempt.

Cancelling that context on `Close` is what the patch wires up. A conn that nobody closes behaves as it does today.

PR #6584, opened 2026-08-02
XHTTP: bound the dial phase with a deadline

embogomolov:
`OpenStream` and `PostPacket` build requests with `context.WithoutCancel` so an established stream survives cancellation of the caller's context. That part is intended.

The side effect is that the same uncancellable context reaches `DialTLSContext`, and `http2.Transport` has no dial timeout of its own. Connection setup runs with no deadline at all. A server that completes the TCP handshake and then goes quiet during TLS parks the dial goroutine in `clientConnPool.getClientConn` with nothing left to wake it.

Give the establishment phase its own deadline. `ctxInner` is not referenced once a connection is returned, so `defer cancel()` cannot disturb an established stream.

The value is a package constant matching the default `policy.Timeouts.Handshake`. Reading the policy would need a `policyManager`, and `createHTTPClient` cannot reach one: it runs inside the `XmuxManager` factory closure, and its clients are cached in `globalDialerMap` and reused, so capturing a request context there would pin the first caller's context to every reuse.

For comparison, `grpc/dial.go` marks its nested dial with `session.ContextWithTimeoutOnly`, detaching it from the parent while keeping an idle timeout. XHTTP detached it from the parent and from any timeout.

### DIFF
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -46,6 +46,10 @@ var (
 	globalDialerAccess sync.Mutex
 )
 
+// dialHandshakeTimeout bounds the connection-establishment phase of an XHTTP
+// dial. Matches the default policy handshake timeout.
+const dialHandshakeTimeout = 60 * time.Second
+
 func getHTTPClient(ctx context.Context, dest net.Destination, streamSettings *internet.MemoryStreamConfig) (DialerClient, *XmuxClient) {
 	realityConfig := reality.ConfigFromStreamSettings(streamSettings)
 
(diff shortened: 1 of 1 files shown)


Fangliding:
这种小超时问题太多了 没有太多实际错误就不接受了 服务端一直不回话是用户配置的服务端的问题 如果是连接整个死掉了会被tcp keepalive探测到

embogomolov:
Keepalive will not catch this. `system_dialer.go` sets `KeepAliveConfig{Enable: true, Idle: 45s, Interval: 45s, Count: -1}`. With `Count: -1` the probes never give up, so nothing closes the connection no matter how long the peer stays quiet. And the peer is not quiet at the TCP layer: it completed the handshake and acks the probes. It just never sends the TLS response.

`net.Dialer.Timeout` already bounds the TCP phase at 16s. Nothing bounds what comes after it: `reality.UClient` and `tls.UConn.HandshakeContext` both take `ctxInner`, which arrives uncancellable from `context.WithoutCancel`, and `getClientConn` waits on `<-call.done` with no select on the context.

That makes it a goroutine leak rather than a slow dial. A stalled handshake parks a goroutine in `getClientConn` with no timer and no cancellable context to wake it, so it stays for the life of the process, holding its `Dispatch` and its socket. Any caller that redials on a schedule accumulates them without limit. The burst observatory is one such caller, and it is enough to exhaust memory on a small device.

The server side is worth separating out. A plain HTTPS request to the same address and port answers in under a second, so the endpoint is not down. Only the XHTTP handshake stalls. Whatever the server is doing wrong, the client should not leak a goroutine per attempt because of it.

A fixed constant is the weakest part of this. `policy.Timeouts.Handshake` would be the right source, but reaching it needs a `policyManager`, and `createHTTPClient` has no path to one: its clients live in `globalDialerMap` and get reused across requests.

PR #6586, opened 2026-08-03
SS2022 outbound: call deferred context cancel to avoid leak

rxzsu:
## What does this PR do?

Fixes a context leak flagged by `go vet` in the SS2022 outbound ([link](proxy/shadowsocks_2022/outbound.go)).

When `session.TimeoutOnlyFromContext` is set, `context.WithCancel(context.Background())` was created and the returned `cancel` function was discarded, so the context could never be cancelled and leaked until GC.

## How to test

```bash
go build ./...
go vet ./proxy/shadowsocks_2022/...
go test ./proxy/...
```

No behavioral change: the cancel fires only when `Process` returns, which is exactly when the returned context is no longer needed.


Fangliding:
TimeoutOnlyFromContext 这个很多地方都在用 还需要研究一下

PR #6614, opened 2026-08-12
VLESS inbound: Support pluggable external user validators (`memory`/`redis`/`rest`)

bitwiresys:
Adds a pluggable `Validator` interface to the VLESS inbound so user lookup can be delegated to an external store instead of always living in the inbound's in-process `clients` list. Three backends are included: `memory` (existing behavior, unchanged, now factored into its own package), `redis`, and `rest`.

**Why:** VLESS currently has no way to delegate user validation externally, so multi-node deployments have to duplicate/distribute user lists across every node by hand. A pluggable validator lets nodes share a single external user store and stay stateless.

Config is fully backward compatible — omitting `validator` keeps the existing in-memory behavior:

```json
{
  "protocol": "vless",
  "settings": {
    "clients": [...],
    "validator": {
      "type": "rest",
      "settings": { "address": "http://user-store:8080" }
    }
  }
}
```

`redis` and `rest` validators are covered by unit tests (RESP2 parsing, `Add`/`Del` concurrency, HTTP client behavior). A reference REST server implementing the same OpenAPI contract used by the `rest` validator's generated client is included under `proxy/vless/validator/rest/test_serv` for local testing.


Fangliding:
。。。

bitwiresys:
> 。。。

can you please explain the reason for closing the PR?

Fangliding:
vibe且为了一个非主要的功能引入了繁杂的接口和大量的依赖
接口都预留在那了就是一个可实现的interface而已二开接入什么乱七八糟的数据库或者外部验证器都非常简单 没有必要塞这么大一坨预制的东西进核心

PR #6621, opened 2026-08-14
splithttp: fall back to header padding for query placement on responses

### LINKED ISSUE #6619: XHTTP: `xPaddingPlacement: "query"` silently produces no padding on server responses (POST ack & GET stream-down headers)
## Title
XHTTP: `xPaddingPlacement: "query"` silently produces no padding on server responses (POST ack & GET stream-down headers)

## Summary
When `xPaddingObfsMode: true` and `xPaddingPlacement: "query"` are configured, request-side padding (client → server) works correctly for both POST (uplink) and GET (downlink open) requests. However, on the **server → client response side**, `ApplyXPaddingToResponse()` in `transport/internet/splithttp/xpadding.go` has no handling for `PlacementQuery`, so the generated padding value is silently discarded and **no padding is ever written to the response** — not to headers, not to cookies, nothing. This happens for both the POST upload-ack response and the initial GET stream-down response headers.

Since HTTP responses have no URL/query string, `query` placement is inherently meaningless on the response side — but the current behavior is a silent no-op rather than a config validation error or a documented fallback, which makes the gap easy to miss and defeats the purpose of enabling `xPaddingObfsMode` in the first place (users may believe both directions are padded when only the request direction is).

## Environment
- Xray-core version: `v26.7.28`
- Transport: XHTTP , mode `packet-up ` 
- Reverse proxy / CDN in front: `nginx`

## Configuration used

```json
"xhttpSettings": {
  "path": "/x-path",
  "xPaddingObfsMode": true,
  "xPaddingPlacement": "query",
  "xPaddingKey": "_xxx",
  "xPaddingMethod": "tokenish",
  "sessionIDPlacement": "cookie",
  "sessionIDKey": "sessionIDKey",
  "sessionIDTable": "Base62",
  "sessionIDLength": "16-32",
  "seqPlacement": "cookie",
  "seqKey": "seqKey"
}
```
(client-side `extra` block mirrors the same fields)

## Observed behavior
- Captured with a raw HTTP/TLS proxy in front of the server: every POST upload-ack response and the initial GET stream-down response come back with an identical, fixed set of headers and **no `X-Padding` header, no extra `Set-Cookie` padding entry, and no other size-varying field**.
- Response size is effectively constant across requests (aside from `Date`), which is exactly the kind of fixed-length fingerprint that `xPaddingBytes`/`xPaddingObfsMode` is meant to eliminate.
- By contrast, switching `xPaddingPlacement` to `"header"` or `"cookie"` produces a response with a randomly sized `X-Padding` header / `Set-Cookie` entry each time, as expected.

## Expected behavior
One of:
1. Config validation should reject `xPaddingPlacement: "query"` at startup with a clear error (e.g. "query placement is only valid for requests; use header/cookie/queryInHeader for response padding"), **or**
2. The response path should gracefully fall back to a default placement (e.g. the `X-Padding` header) when the configured placement is not response-compatible, with a log line noting the fallback.

Silently doing nothing is the worst of the three options because it looks like padding is working (no errors, no warnings) while it is not.

## Root cause (code-level)

In `transport/internet/splithttp/hub.go`, `requestHandler.ServeHTTP()` builds the padding config for every incoming request (GET/POST/OPTIONS) and unconditionally calls `ApplyXPaddingToResponse`:

```go
length := int(h.config.GetNormalizedXPaddingBytes().rand())
config := XPaddingConfig{Length: length}

if h.config.XPaddingObfsMode {
    config.Placement = XPaddingPlacement{
        Placement: h.config.XPaddingPlacement,
        Key:       h.config.XPaddingKey,
        Header:    h.config.XPaddingHeader,
    }
    config.Method = PaddingMethod(h.config.XPaddingMethod)
} else {
    config.Placement = XPaddingPlacement{
        Placement: PlacementHeader,
        Header:    "X-Padding",
    }
}
h.config.ApplyXPaddingToResponse(writer, config)
```

In `transport/internet/splithttp/xpadding.go`:

```go
func (c *Config) ApplyXPaddingToResponse(writer http.ResponseWriter, config XPaddingConfig) {
    placement := config.Placement.Placement
    if placement == PlacementHeader || placement == PlacementQueryInHeader {
        c.ApplyXPaddingToHeader(writer.Header(), config)
        return
    }
    paddingValue := GeneratePadding(config.Method, config.Length)
    switch placement {
    case PlacementCookie:
        ApplyPaddingToResponseCookie(writer, config.Placement.Key, paddingValue)
    }
}
```

When `placement == PlacementQuery` ("query"):
- It doesn't match the `if` (Header / QueryInHeader).
- It doesn't match the `switch` `case PlacementCookie`.
- There is no `default` branch.
- `paddingValue` is computed (`GeneratePadding` is called, consuming randomness) and then discarded.

Result: no bytes are written anywhere for the response in this configuration — a silent no-op.

For contrast, the **request-side** equivalent (`ApplyXPaddingToRequest`, same file) does implement `PlacementQuery` via `ApplyPaddingToQuery(req.URL, ...)`, which is why request padding (both POST uploads and the GET used to open the download stream) works fine — only the response side is missing the case.

## Suggested fix
Add an explicit case (or a `default`) in `ApplyXPaddingToResponse` for `PlacementQuery` — either:
- fall back to `PlacementHeader` with a fixed/default header name (mirroring the `else` branch used when `XPaddingObfsMode` is false), or
- reject this combination at config-parsing time in `infra/conf` with a descriptive error.

### Additional suggestion: consider body-based padding

Independent of how `PlacementQuery` is handled, it's worth noting that **the POST upload-ack response body is always empty (`Content-Length: 0`) regardless of `xPaddingPlacement`** — padding is currently only ever written to a header or a cookie, never to the response body itself. A constantly-zero response body is itself a distinguishing feature that some CDNs/DPI systems can key on (a "real" API endpoint rarely returns an identically-empty body on every call).

Since the packet-up ack response carries no application data anyway (the actual proxied return traffic goes over the separate GET stream), there's no functional reason the body has to stay empty. Writing a random-length padding blob into the response body — with the length drawn from `xPaddingBytes` the same way header/cookie padding is — would additionally vary `Content-Length` itself, not just header sizes, which header/cookie-only padding cannot do. This could be introduced as a new `PlacementBody` case for `ApplyXPaddingToResponse`, gated the same way as the request-side data placements already are (`config.go`'s existing `PlacementBody` for uplink data placement is a useful precedent to follow for the wiring).

I'd suggest treating this as a follow-up enhancement to the fix above rather than blocking on it, since it changes response semantics slightly (previously-zero-length responses now carry a body) and may need its own review/testing pass.

kaixindmr:
## What

When `xPaddingPlacement: "query"` is configured with `xPaddingObfsMode: true`, the server never pads its responses. `ApplyXPaddingToResponse` in `transport/internet/splithttp/xpadding.go` generates the padding value but has no `PlacementQuery` case — HTTP responses carry no query string — so the value is silently discarded. The client → server direction is padded, the server → client direction is not.

## Why

With obfs mode enabled, users reasonably expect both directions to be padded. The silent no-op makes the gap easy to miss and undermines the traffic-shaping purpose of `xPaddingObfsMode`.

## Fix

Fall back to the configured padding header (defaults to `X-Padding`) when placement is `query` on the response side, so the response direction stays padded without changing the request-side behavior or breaking existing configs. The client never validates response padding, so the extra header is safe.

## Tests

- New `TestApplyXPaddingToResponse` covers query → header fallback (default and custom header), header placement, cookie placement, and zero-length no-op.
- New `TestApplyXPaddingToRequestQueryPlacement` verifies request-side query padding still lands in the URL query while the response falls back to the header.
- `go test ./transport/internet/splithttp/...` and `go build ./...` pass.

Fixes #6619

### DIFF
--- a/transport/internet/splithttp/xpadding.go
+++ b/transport/internet/splithttp/xpadding.go
@@ -245,6 +245,14 @@ func (c *Config) ApplyXPaddingToResponse(writer http.ResponseWriter, config XPad
 	switch placement {
 	case PlacementCookie:
 		ApplyPaddingToResponseCookie(writer, config.Placement.Key, paddingValue)
+	case PlacementQuery:
+		// HTTP responses carry no query string, so "query" placement has no
+		// place to put padding on the response side. Fall back to the
+		// configured padding header instead of silently discarding the
+		// generated value, so both directions stay padded in obfs mode.
+		if header := config.Placement.Header; header != "" && paddingValue != "" {
+			writer.Header().Set(header, paddingValue)
+		}
 	}
 }
 
--- a/transport/internet/splithttp/xpadding_test.go
+++ b/transport/internet/splithttp/xpadding_test.go
@@ -1,0 +1,139 @@
+package splithttp_test
+
+import (
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	. "github.com/xtls/xray-core/transport/internet/splithttp"
+)
+
+func TestApplyXPaddingToResponse(t *testing.T) {
+	newConfig := func(placement, key, header string) *Config {
+		c := &Config{
+			XPaddingObfsMode: true,
+			XPaddingMethod:   string(PaddingMethodRepeatX),
+		}
+		c.XPaddingKey = key
+		c.XPaddingHeader = header
+		c.XPaddingPlacement = placement
+		return c
+	}
+
+	tests := []struct {
+		name          string
+		placement     string
+		key           string
+		header        string
+		length        int
+		wantHeaderKey string
+		wantCookie    string
+	}{
+		{
+			name:          "query placement falls back to padding header",
+			placement:     PlacementQuery,
+			key:           "x_padding",
+			header:        "X-Padding",
+			length:        100,
+			wantHeaderKey: "X-Padding",
+		},
+		{
+			name:          "query placement uses custom header",
+			placement:     PlacementQuery,
+			key:           "x_padding",
+			header:        "X-Custom-Padding",
+			length:        100,
+			wantHeaderKey: "X-Custom-Padding",
+		},
+		{
+			name:          "header placement sets configured header",
+			placement:     PlacementHeader,
+			key:           "x_padding",
+			header:        "X-Padding",
+			length:        100,
+			wantHeaderKey: "X-Padding",
+		},
+		{
+			name:       "cookie placement sets cookie",
+			placement:  PlacementCookie,
+			key:        "x_padding",
+			header:     "X-Padding",
+			length:     100,
+			wantCookie: "x_padding",
+		},
+		{
+			name:          "zero length produces no output",
+			placement:     PlacementQuery,
+			key:           "x_padding",
+			header:        "X-Padding",
+			length:        0,
+			wantHeaderKey: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := newConfig(tt.placement, tt.key, tt.header)
+			rec := httptest.NewRecorder()
+			config := XPaddingConfig{
+				Length: tt.length,
+				Placement: XPaddingPlacement{
+					Placement: tt.placement,
+					Key:       tt.key,
+					Header:    tt.header,
+				},
+				Method: PaddingMethodRepeatX,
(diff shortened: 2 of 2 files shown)


Fangliding:
服务端本来就不存在query 属于配置错误

PR #6624, opened 2026-08-14
Tunnel inbound: SO_REUSEPORT fix and IPv6 support for the OpenBSD transparent proxy

adsdevdoo:
Follow-up to #6546.

On OpenBSD every UDP reply to a client is sent from a socket bound to the
original destination with `SO_BINDANY`, and `FakeUDP` creates one such socket
per session. Two clients talking to the same destination at the same time, two
phones resolving over the same DNS server for instance, therefore bind the same
address and port twice. The second bind fails with `EADDRINUSE`, surfacing as
`fake: bind ...: address already in use`, and that session never gets an answer.

`SO_REUSEADDR` alone does not cover this. In `in_pcbaddrisavail_lock()` the
reuse flag starts as `so->so_options & SO_REUSEPORT` and is only widened to
accept `SO_REUSEADDR` for multicast addresses, so a duplicate binding on a
unicast address needs `SO_REUSEPORT` on both sockets:

```c
int reuseport = (so->so_options & SO_REUSEPORT);

if (IN_MULTICAST(sin->sin_addr.s_addr)) {
	if (so->so_options & (SO_REUSEADDR|SO_REUSEPORT))
		reuseport = SO_REUSEADDR|SO_REUSEPORT;
}
...
if (t && (reuseport & t->inp_socket->so_options) == 0)
	error = EADDRINUSE;
```

`proxy/dokodemo/fakeudp_linux.go` already sets both options, this brings the
OpenBSD implementation in line.

Tested on OpenBSD 7.7/7.8/7.9 amd64 on a live router with several LAN clients.

### DIFF
--- a/proxy/dokodemo/fakeudp_openbsd.go
+++ b/proxy/dokodemo/fakeudp_openbsd.go
@@ -12,12 +12,29 @@ import (
 )
 
 func FakeUDP(addr *net.UDPAddr, mark int) (net.PacketConn, error) {
-	ip4 := addr.IP.To4()
-	if ip4 == nil {
-		return nil, &net.OpError{Op: "fake", Err: fmt.Errorf("IPv6 is not supported by the OpenBSD transparent UDP patch")}
+	domain := unix.AF_INET6
+	var sockaddr unix.Sockaddr
+	if ip4 := addr.IP.To4(); ip4 != nil {
+		domain = unix.AF_INET
+		sa := &unix.SockaddrInet4{Port: addr.Port}
+		copy(sa.Addr[:], ip4)
+		sockaddr = sa
+	} else if ip6 := addr.IP.To16(); ip6 != nil {
+		sa := &unix.SockaddrInet6{Port: addr.Port}
+		copy(sa.Addr[:], ip6)
+		if addr.Zone != "" {
+			iface, err := net.InterfaceByName(addr.Zone)
+			if err != nil {
+				return nil, &net.OpError{Op: "fake", Err: fmt.Errorf("resolve zone %s: %w", addr.Zone, err)}
+			}
+			sa.ZoneId = uint32(iface.Index)
+		}
+		sockaddr = sa
+	} else {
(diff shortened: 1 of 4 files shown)


Fangliding:
6546晾了三个星期为什么刚好合并两天才发现。。

adsdevdoo:
Fair point. It only shows up when two sessions reach the same destination
at the same time, and my first round of testing was with a single client.
I hit it later, while running IPv6 on a live router with several devices
behind it.

The fix was already sitting in my IPv6 branch. I split it out into its own
PR so that a one-line fix to code that is already merged would not have to
wait for review of a whole feature.

RPRX:
~~直接把 IPv6 的代码合并到这个 PR 吧，这 OpenBSD 专属代码没人有能力深入 review~~

adsdevdoo:
Done, added the IPv6 commit here. The PR is now two commits:

- the SO_REUSEPORT fix, unchanged;
- IPv6 support for the transparent proxy. The original destination of an
  inet6 flow arrives in different control messages, IPV6_PKTINFO for the
  address and IPV6_RECVDSTPORT for the port, and the reply socket has to be
  created in the matching family, because OpenBSD always forces IPV6_V6ONLY
  and a single socket never serves both. The inbound sockopt setter now
  tries both option pairs and keeps whichever the socket's family accepts.

Tested on OpenBSD 7.7/7.8/7.9 amd64 on live routers, IPv6 via an HE tunnel:
TCP and UDP (DNS, QUIC) work in both families.

One piece is left out on purpose: outbound UDP to an IPv6 destination also
needs a small fix in the system dialer, but that one is not OpenBSD-specific,
so I will send it as a separate PR.

RPRX:
> One piece is left out on purpose: outbound UDP to an IPv6 destination also needs a small fix in the system dialer, but that one is not OpenBSD-specific, so I will send it as a separate PR.

这个策略是对的，这个 PR 只包含 OpenBSD 平台特定代码，所以将会得到快速合并

Issue #5576, opened 2026-01-19
VLESS + XHTTP + REALITY 在 flow=xtls-rprx-vision 下无法连接，报 “XTLS only supports TLS and REALITY directly for now”，但 26.1.18 提示必须迁移到 flow


lyd123qw2008:
### 完整性要求

- [x] 我读完了 issue 模板中的所有注释，确保填写符合要求。
- [x] 我保证阅读了文档，了解所有我编写的配置文件项的含义，而不是大量堆砌看似有用的选项或默认值。
- [x] 我提供了完整的配置文件和日志，而不是出于自己的判断只给出截取的部分。
- [x] 我搜索了 issues, 没有发现已提出的类似问题。
- [x] 问题在 Release 最新的版本上可以成功复现

### 描述

  - 版本：Xray 26.1.18（Linux）
  - 场景：VLESS + XHTTP + REALITY
  - 现象：按 26.1.18 的提示迁移到 flow: "xtls-rprx-vision"（客户端/服务端都加）后，连接失败。
  - 客户端日志关键错误：

    proxy/vless/outbound: XTLS only supports TLS and REALITY directly for now.
  - 结论：XHTTP + REALITY 不能使用 Vision flow，但 26.1.18 的弃用提示会误导用户必须迁移到 flow。
  - 期望：
      1. 文档/提示明确 xHTTP 不能使用 Vision flow；或
      2. 警告只对支持 Vision 的传输方式提示。

### 重现方式

  1. 使用下方服务端配置启动 Xray
  2. 使用下方客户端配置启动 Xray
  3. 通过客户端 SOCKS 访问任意 HTTPS 域名（如 curl --socks5-hostname 127.0.0.1:10808 https://www.cloudflare.com）
  4. 客户端日志出现：XTLS only supports TLS and REALITY directly for now，连接失败
  5. 将两端 flow 去掉后，连接恢复正常（验证是 flow + xHTTP 组合导致）

### 客户端配置

<details><pre><code>
  {
    "log": {
      "loglevel": "debug"
    },
    "dns": {
      "dnsLog": true,
      "servers": [
        "1.1.1.1",
        "8.8.8.8"
      ]
    },
    "inbounds": [
      {
        "tag": "socks",
        "port": 10808,
        "listen": "127.0.0.1",
        "protocol": "mixed",
        "settings": {
          "auth": "noauth",
          "udp": true
        }
      }
    ],
    "outbounds": [
      {
        "tag": "proxy",
[...]

PR #6629, opened 2026-08-15
VLESS: Support XTLS Vision on non-RAW transports (e.g. XHTTP)

dkjsiogu:
﻿## What does this PR do?

Allows the VLESS flow `xtls-rprx-vision` to be used on non-RAW transports (e.g. XHTTP, WebSocket, gRPC), instead of rejecting them with `"XTLS only supports TLS and REALITY directly for now."`.

Currently both the VLESS inbound and outbound hard-require the connection to be a `tls.Conn` / `reality.Conn` / `encryption.CommonConn` when Vision is enabled. XHTTP (and other packet/capsule transports) wraps the TLS conn, so Vision was rejected even though the padding part of the protocol works fine on any reliable stream.

## How it works

- When the transport is not RAW (checked with `proxy.IsRAWTransportWithoutSecurity`), `CanSpliceCopy` is set to `3` (same convention as Shadowsocks/Hysteria/WireGuard/Trojan) so no splice/penetration is attempted.
- `VisionReader` no longer assumes `input`/`rawInput` are non-nil: the direct-copy branch now nil-checks them and releases them safely.
- Vision still runs in padding mode as usual: padding + filtering are applied, and after the inner TLS handshake is detected both sides keep relaying raw bytes over the existing stream.

## Tests

- `go build ./...` passes.
- End-to-end verified against a real server (Xray 26.7.28 built from this branch, `main` commit `7d214f8b`):
  - VLESS + XHTTP + REALITY + `flow=xtls-rprx-vision` (client and server): TCP transfers, 10MB downloads, and UDP (via mux) all work.
  - Control groups (TCP+REALITY+Vision, XHTTP+REALITY without flow) also pass.

Related: #5576

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -258,16 +258,20 @@ func (w *VisionReader) ReadMultiBuffer() (buf.MultiBuffer, error) {
 
 	if *switchToDirectCopy {
 		// XTLS Vision processes TLS-like conn's input and rawInput
-		if inputBuffer, err := buf.ReadFrom(w.input); err == nil && !inputBuffer.IsEmpty() {
-			buffer, _ = buf.MergeMulti(buffer, inputBuffer)
+		if w.input != nil {
+			if inputBuffer, err := buf.ReadFrom(w.input); err == nil && !inputBuffer.IsEmpty() {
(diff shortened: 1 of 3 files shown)


Fangliding:
meaningless

PR #6637, opened 2026-08-17
VLESS inbound: Add pull-through external HTTP validator

needthisforctf:
Adds another implementation of Validator that checks uuids against external server via HTTP queries, instead of JSON file modifications or gRPC handler API.

## Reasoning

The gRPC `HandlerService` (`AddUserOperation`/`RemoveUserOperation`) is inherently **push-based**. This has numerous disadvantages:

- A lot more work has to be done by panels and node harnesses to ensure consistency of inbounds across multiple Xray servers; and even that provides only _eventual consistency_. The problem is worsening with the increasing node count: essentially, each user addition produces N network requests to nodes.
- Inbound has to contain all of the users authenticating against it at any given time. Sure, in-memory search against them is fast even after 100k+, but loading users to the inbound after Xray boot takes considerable time and has to be front-loaded for the inbound to become functional. Also, keeping all of the required user uuids in an inbound is a bigger potential attack surface — better have _some_ user uuids stolen than _all_ of them.
- IP limiting is retroactive, not proactive — node harness has to poll Xray api for list of the client IPs per user, which creates excessive workload that still lets a user with known bad IP slip before being caught. And since uuid auth authenticates _connections_, not _packets_, the malicious user can keep sending data after being removed from the inbound, so it's better to not let the malicious user in than later show him out.

## Implementation

```jsonc
"inbounds": [{
  "protocol": "vless",
  "settings": {
    "clients": [
      { "id": "...", "email": "static@example.com" }  // still checked first, still gRPC-manageable
    ],
    "decryption": "none",
    "validator": {
      "type": "external",              // "memory" (default) or "external"
      "url": "https://auth.example.com/check",
      "timeout": 2,                    // seconds, optional, default 2
      "cacheTtl": 300,                 // seconds, optional, default 300
      "negativeTtl": 30,               // seconds, optional, default 30
      "outbound": "auth-out",          // optional: dial the endpoint through a tagged outbound
      "headers": {                     // optional: sent as-is on every lookup request
        "Authorization": "Bearer s3cr3t",
        "X-Node-Region": "fra"
      }
    }
  }
}]
```

`ExternalValidator` works like a pull-through cache. With `"type": "external"`, the local clients list is still checked first (and still fully manageable through the existing gRPC API); only UUIDs not found locally fall through to the endpoint, as `GET <url>?uuid=<uuid>`, expecting either:

- 200 with `{"email": "...", "level": 0, "flow": "", "ttl": 0}` (`ttl` optionally overrides `cacheTtl` per user, e.g. for short-lived trial accounts, will get to that later), or
- any non-200 status, treated as "not a user" and cached under `negativeTtl` (5s on transport errors and 401s, so a recovered endpoint or a fixed credential is retried quickly rather than waiting out a full negative-cache window).

`headers` is an arbitrary string map, mostly intended for Bearer tokens, but not limited to it — any static per-request metadata the backend wants (region, node id, a shared secret) goes here unchanged.

Lookups are deduplicated via `singleflight` so concurrent connections on an unresolved UUID produce one request, not one per connection.

## Per-IP admission

To address the elephant in the room: it's optional. The way it works, if backend wants to cap distinct source IPs per user, each lookup can also carry `&ip=<source-ip>` and results are cached per `(uuid, ip)` pair rather than per `uuid` alone — so an endpoint enforcing `"N devices per account"` only needs to answer truthfully per IP; the inbound handles the caching split.

Ofc, it's not only useful to provide low bound cap, Xray sends ip, backend can also use it as an anti-DDoS mark, denying after a huge number of unique IPs, or based on region, etc etc.

## Do I have to wait for TTL to end to instantly ban the user?

No. RemoveUserOperation (gRPC) still works against the external mode: Del evicts matching cache entries (by UUID or by email) before falling through to the local list, so revoking a user that only exists in the external cache doesn't require waiting out cacheTtl.

## Additional concerns

- **LLM usage:** I have used LLMs for tests and style consistency; the logic is hand-written.
- **Cache behaviour:** Better check the code yourself; it's really straightforward, defaults are sane and it can be turned off completely. Me yapping here would do the code disservice.
- **Error messages and logging:** errors are a bit verbose on purpose (for example, you'd get prompted to check the token if you got 401 back)
- **Why re-implement in-memory validators methods to call them under the hood?** Basically, to make it easier to future-proof. If at any point the implementation of in-memory validator I rely upon changes, the PR that does this will have it easier.

Otherwise, I'm up to to discussion and changes.

### DIFF
--- a/infra/conf/vless.go
+++ b/infra/conf/vless.go
@@ -30,13 +30,28 @@ type VLessInboundFallback struct {
 	Xver uint64          `json:"xver"`
 }
 
+// VLessInboundValidatorConfig selects and configures how VLESS users are
+// validated. Type "memory" (the default, used when the field is omitted)
+// checks only the local "clients"/"users" list. Type "external" additionally
+// queries an HTTP(S) endpoint for users not found locally.
+type VLessInboundValidatorConfig struct {
+	Type        string            `json:"type"`
+	URL         string            `json:"url"`
+	Timeout     uint32            `json:"timeout"`     // seconds, 0 = default
+	CacheTTL    uint32            `json:"cacheTtl"`    // seconds, 0 = default
+	NegativeTTL uint32            `json:"negativeTtl"` // seconds, 0 = default
+	Outbound    string            `json:"outbound"`    // tagged outbound to dial the endpoint through
+	Headers     map[string]string `json:"headers"`     // sent as-is on every lookup request
+}
+
 type VLessInboundConfig struct {
-	Users      []json.RawMessage       `json:"users"`
-	Clients    []json.RawMessage       `json:"clients"`
-	Decryption string                  `json:"decryption"`
-	Fallbacks  []*VLessInboundFallback `json:"fallbacks"`
-	Flow       string                  `json:"flow"`
-	Testseed   []uint32                `json:"testseed"`
+	Users      []json.RawMessage            `json:"users"`
+	Clients    []json.RawMessage            `json:"clients"`
+	Decryption string                       `json:"decryption"`
+	Fallbacks  []*VLessInboundFallback      `json:"fallbacks"`
+	Flow       string                       `json:"flow"`
+	Testseed   []uint32                     `json:"testseed"`
+	Validator  *VLessInboundValidatorConfig `json:"validator"`
 }
 
 // Build implements Buildable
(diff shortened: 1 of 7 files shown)


Fangliding:
解释过了

PR #6644, opened 2026-08-18
splithttp: bound physical H2 carriers for maxConnections

kaiserproger:
## Summary

Make XHTTP `xmux.maxConnections` behave like a bounded physical HTTP/2 carrier pool instead of only bounding logical `http2.Transport` instances.

When `maxConnections` is configured for HTTP/2, this change:

- binds one `http2.ClientConn` to each `XmuxClient`, so one logical carrier cannot silently create additional TCP connections;
- assigns new proxy flows to the least-loaded active carrier while keeping each flow pinned to that carrier for its lifetime;
- rotates expired carriers serially with make-before-break semantics: keep `N` active carriers and at most one planned draining carrier while long-lived flows finish.

There is no XHTTP wire-format or server-side protocol change. Non-`maxConnections` behavior keeps the existing transport path.

## Why

There are two independent layers of connection pooling today:

1. XMUX creates up to `maxConnections` `XmuxClient` objects.
2. Every `XmuxClient` uses an `http2.Transport`, which is itself allowed to create multiple physical H2/TCP connections.

There is also a rotation issue: when an XMUX client reaches its reuse/request/time limit, it is removed from the active slice even when pinned proxy flows are still running. The replacement is then created while the old transport is still draining.

For configurations that use `maxConnections` as an outer TCP/H2 recovery-domain budget, the number therefore does not directly describe the number of physical carrier connections.

## Rotation behavior

For fixed `maxConnections = N`:

- the active pool contains `N` carriers;
- each carrier owns one physical H2 `ClientConn`;
- only one expired carrier is removed from new-flow selection at a time;
- its replacement is created immediately, so service remains make-before-break;
- the old carrier remains alive only for its already-pinned flows;
- the next planned rotation waits until that drain finishes.

Thus normal planned rotation is bounded to `N active + at most 1 draining` physical carrier.

Unexpected connection failures/GOAWAY still prioritize recovery rather than waiting for planned rotation.

## Reproducer / live validation

I ran stock v26.7.28 and the patched build as two separate client processes with the same outbound configuration:

- `maxConnections = 8`
- `hMaxReusableSecs = 10-12` (intentionally shortened for the test)
- 8 long-lived throttled downloads were started first, giving both implementations exactly 8 carrier TCP connections
- after 15 seconds, while all 8 original proxy flows were still pinned and running, 8 new proxy flows were opened

Result after the second group connected:

```text
stock v26.7.28: 16 physical TCP carriers
patched:        9 physical TCP carriers
```

All new requests succeeded in both cases. The patched result is the intended 8 active + 1 draining make-before-break state.

A separate throughput check used an incompressible random file and verified the actual HTTP 206 byte counts. One run produced:

```text
                         stock       patched
1 x 64 MiB              394 Mbps     375 Mbps
4 x 64 MiB aggregate    890 Mbps    1032 Mbps
8 x 64 MiB aggregate    899 Mbps     945 Mbps
```

The throughput differences are within normal network variance; the relevant result is that bounding the physical pool did not introduce an aggregate-throughput regression in this test.

## Tests

Added tests cover:

- serial make-before-break rotation of a fixed pool;
- one fixed H2 carrier accepting exactly one physical TCP connection under 256 concurrent requests with server `MaxConcurrentStreams = 32`;
- least-loaded carrier selection for fixed `maxConnections` pools.

Validation on current `main`:

```text
go test ./transport/internet/splithttp
go test -race ./transport/internet/splithttp -run 'Test(FixedPoolSerialRotation|FixedH2RoundTripperUsesOnePhysicalConnection|MaxConnectionsPreferLeastLoaded)' -count=1
CGO_ENABLED=0 go build ./main
```

### DIFF
--- a/transport/internet/splithttp/client.go
+++ b/transport/internet/splithttp/client.go
@@ -40,7 +40,13 @@ type DefaultDialerClient struct {
 }
 
 func (c *DefaultDialerClient) IsClosed() bool {
-	return c.closed
+	if c.closed {
+		return true
+	}
+	if carrier, ok := c.client.Transport.(interface{ IsClosed() bool }); ok {
+		return carrier.IsClosed()
+	}
+	return false
 }
 
 func (c *DefaultDialerClient) OpenStream(ctx context.Context, url string, sessionId string, body io.Reader, uploadOnly bool) (wrc io.ReadCloser, remoteAddr, localAddr net.Addr, err error) {
@@ -177,11 +183,14 @@ func (c *DefaultDialerClient) PostPacket(ctx context.Context, url string, sessio
 	return nil
 }
 
-// HTTP/1.1 and HTTP/2 will close itself, we only handle HTTP/3 here
 func (c *DefaultDialerClient) Close() error {
+	c.closed = true
 	transport := c.client.Transport
+	if closer, ok := transport.(io.Closer); ok {
+		return closer.Close()
+	}
 	if h3Transport, ok := transport.(*http3.Transport); ok {
-		h3Transport.Close()
+		return h3Transport.Close()
 	}
 	return nil
 }
--- a/transport/internet/splithttp/dialer.go
+++ b/transport/internet/splithttp/dialer.go
@@ -81,6 +81,81 @@ func getHTTPClient(ctx context.Context, dest net.Destination, streamSettings *in
 	return xmuxClient.XmuxConn.(DialerClient), xmuxClient
 }
 
+type fixedH2RoundTripper struct {
+	mu        sync.Mutex
+	transport *http2.Transport
+	dial      func(context.Context) (net.Conn, error)
+	conn      *http2.ClientConn
+	rawConn   net.Conn
+	closed    bool
+}
+
+func (t *fixedH2RoundTripper) getConn(ctx context.Context) (*http2.ClientConn, net.Conn, bool, error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if t.closed {
+		return nil, nil, false, errors.New("fixed H2 carrier is closed")
+	}
+	if t.conn != nil {
+		state := t.conn.State()
+		if state.Closed || state.Closing {
+			return nil, t.rawConn, false, errors.New("fixed H2 carrier is no longer reusable")
+		}
+		return t.conn, t.rawConn, true, nil
+	}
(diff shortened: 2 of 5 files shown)


kaiserproger:
Additional production-only validation after removing all auxiliary canary processes/services:

- one Xray process only;
- `maxConnections = 8`;
- temporarily shortened `hMaxReusableSecs` to `30-45` seconds;
- three expiry/burst cycles over about four minutes;
- bursts of 64-96 new proxy flows per cycle;
- every request succeeded and exited through the expected remote endpoint;
- physical Frankfurt TCP count stayed at `9` maximum (`8 active + 1 draining`);
- `CLOSE_WAIT = 0` after each cycle.

After the stress test the production reuse window was restored to `660-840` seconds and the pool returned to 8 physical carriers.

I also found and removed stale auxiliary `xray-rotation-*` test services that had contaminated some earlier connection-count observations; the PR description now uses the isolated stock-vs-patched reproducer and the clean production result above.

Fangliding:
我觉得没啥必要做这么严
一个transport只有在超过对端指定的 max stream 的时候会尝试开新连接 这个其实一个transport参数就能抑制住 这个时候已经几百子流+了（这里是你的服务器自己故意指定的一个很低的 MaxConcurrentStreams = 32） 如果用户真有那么多请求还是放过去吧 不然不是用户断网就是逼用户改参数 最后目的还是没达到结果是一样的
至于drain活着是期望行为

kaiserproger:
Thanks — you were right about the simpler mechanism.

I tested the stock v26.7.28 client in the real deployment by changing only the peer-advertised HTTP/2 stream limit on nginx:

- `http2_max_concurrent_streams 16` -> the stock Xray process opened **23 physical TCP connections** with `xmux.maxConnections = 8` under normal live traffic;
- `http2_max_concurrent_streams 1024` -> after recreating the H2 pool, the same stock Xray returned to **8 physical TCP connections**;
- then I ran 128 concurrent throttled range-download proxy flows through that stock client; it still remained at **8 physical TCP connections**, with `CLOSE_WAIT = 0`.

So for my deployment raising the peer stream limit is enough and keeping draining carriers alive is indeed the right behavior. I am reverting the custom Xray build in production and keeping stock XMUX with time-based rotation. Thanks for pointing this out.

PR #6646, opened 2026-08-20
Conf: Validate vnext/servers for unencrypted VLESS/Trojan

barkure:
#6303 forbids unencrypted VLESS/Trojan outbounds to public addresses, but the check only looked at top-level `address` / `encryption`. The old `vnext` / `servers` style (V2rayN) therefore skipped the prohibition, while the simplified style (V2rayNG) was blocked.

This validates both styles, matching `Build()`:

- VLESS: top-level `Address`+`Encryption`, or each `vnext` address plus per-user `encryption`
- Trojan: top-level `Address`, or each `servers` entry (nil-safe)

Private IPs/domains, VLESS encryption other than `none`, and TLS/REALITY remain allowed. Coverage goes through `OutboundDetourConfig.Build()`.

### DIFF
--- a/infra/conf/xray.go
+++ b/infra/conf/xray.go
@@ -248,17 +248,58 @@ func validateOutboundTransportSecurity(rawConfig interface{}, senderSettings *pr
 	}
 
 	if vlessCfg, ok := rawConfig.(*VLessOutboundConfig); ok {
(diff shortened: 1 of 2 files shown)


Fangliding:
噗 AI写的啥啊

PR #6661, opened 2026-08-21
Sniffing: Add `portsExcluded` to skip the sniffer for given destination ports

rokokol:
Adds `portsExcluded` to `sniffing`: a port list checked in the dispatcher **before** the sniffer reads anything, so a connection to a listed destination port takes the same path as sniffing being switched off.

Follows up on discussion #6654 — sending the patch since it is small enough to judge from the diff. Close it if the shape is unwanted.

### The problem

Where the peer speaks first — SSH, IMAP, SMTP — the client sends nothing for the sniffer to read, so `cacheDeadline` in `app/dispatcher/default.go` burns down in full before the dispatch continues. The cost is fixed and paid on every such connection.

`domainsExcluded` and `ipsExcluded` do not help: both are consulted in `shouldOverride`, i.e. *after* the sniffer has already waited, and they only suppress the destination override. #5927 shows that placement is deliberate, which is why this is a separate knob rather than a change to those.

### The change

- `ports_excluded` in `SniffingConfig` (field 7), `portsExcluded` in the JSON config, taking the usual port-list syntax: `"22,143,993-995"`
- `SniffingRequest.ExcludeForPort` as a `net.MemoryPortList`, filled by `BuildSniffingRequest`
- one check in the dispatcher — `shouldSniff(request, destination)` replaces the bare `request.Enabled` test at both call sites

```json
"sniffing": {
  "enabled": true,
  "destOverride": ["http", "tls"],
  "portsExcluded": "22,143,993-995"
}
```

### Measurements

socks inbound → freedom outbound, against a loopback server that sends its banner on connect, 30 connections each, median from SOCKS5 CONNECT to the first byte:

| setup | median |
|---|---|
| direct, no proxy | 0.2 ms |
| sniffing disabled | 66.3 ms |
| sniffing on, port excluded | 67.0 ms |
| sniffing on, port not excluded | 268.1 ms |

The excluded row lands on the sniffing-disabled row. Scope check on a single instance configured with `"portsExcluded": "2222"`: port 2222 answers in 66.8 ms, port 2223 in 267.7 ms, so the exclusion stays on the listed ports.

### Tests

- `shouldSniff` table test in `app/dispatcher` — disabled, no exclusions, single port, both range ends, just past the range
- `BuildSniffingRequest` with and without `portsExcluded` in `app/proxyman`; the nil case would panic in `PortListFromProto` without the guard
- `TestSniffingConfig_Build` in `infra/conf` extended over the JSON → proto path

`go build ./...` and `go test ./...` pass and `go run ./infra/vformat/main.go -mode check` is clean. `config.pb.go` was regenerated with protoc v33.5 and protoc-gen-go v1.36.11, so its header still matches every other `.pb.go` for the `check-proto` job. One local failure, `TestTCPLocalNameServer` in `app/dns`, reproduces identically on a clean checkout here — network-dependent, unrelated.

Docs for the field belong in Xray-docs-next; happy to send that too if this lands.

### DIFF
--- a/app/dispatcher/default.go
+++ b/app/dispatcher/default.go
@@ -229,6 +229,13 @@ func trackOnlineIP(ctx context.Context, sm stats.Manager, email, ip string) {
 	}
 }
 
+// shouldSniff reports whether the payload is worth reading at all. Ports listed
+// in portsExcluded are skipped here, before the sniffer waits for a first packet
+// the client may never send.
+func shouldSniff(request session.SniffingRequest, destination net.Destination) bool {
+	return request.Enabled && !request.ExcludeForPort.Contains(destination.Port)
+}
+
 func (d *DefaultDispatcher) shouldOverride(ctx context.Context, result SniffResult, request session.SniffingRequest, destination net.Destination) bool {
 	domain := result.Domain()
 	if domain == "" {
@@ -284,7 +291,7 @@ func (d *DefaultDispatcher) Dispatch(ctx context.Context, destination net.Destin
 
 	sniffingRequest := content.SniffingRequest
 	inbound, outbound := d.getLink(ctx)
-	if !sniffingRequest.Enabled {
+	if !shouldSniff(sniffingRequest, destination) {
 		go d.routedDispatch(ctx, outbound, destination)
 	} else {
 		go func() {
@@ -340,7 +347,7 @@ func (d *DefaultDispatcher) DispatchLink(ctx context.Context, destination net.De
 	}
 	outbound = WrapLink(ctx, d.policy, d.stats, outbound)
 	sniffingRequest := content.SniffingRequest
-	if !sniffingRequest.Enabled {
+	if !shouldSniff(sniffingRequest, destination) {
 		d.routedDispatch(ctx, outbound, destination)
 	} else {
 		cReader := &cachedReader{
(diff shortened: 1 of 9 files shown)


Fangliding:
IAMP和SMTP很少被个人用户用作代理 专门的邮件用途完全可以关嗅探 ssh的并发也不多是典型的长时间连接 我觉得这点延迟就算了 主要是不想xxxxExcluded过于膨胀 如果目标已知那么合理使用 domainsExcluded 也行
以及ssh我记得并不是严格的server talk first 你可以找一个会事先发送自己的信息的

rokokol:
Fair on ssh, and I checked rather than argued: OpenSSH sends its ident first, without waiting for the server.

```
$ nc -l 127.0.0.1 2222      # listener stays silent
SSH-2.0-OpenSSH_10.4
```

22 bytes on connect, so the sniffers reject immediately, `Sniff` returns `errUnknownContent`, and the `default:` branch in `sniffCall` returns without spending `cacheDeadline`. My benchmark used a server that speaks first — it measures the deadline path correctly, but it does not describe ssh. Struck from the case.

What is left is protocols where the server genuinely opens: SMTP, IMAP, POP3, and also MySQL, PostgreSQL, FTP.

On turning sniffing off for mail — that works when mail has its own inbound. The setup I hit this on is a single client-facing inbound carrying everything, where there is nothing to split on: disabling sniffing there costs the domain routing it is enabled for. That is the case the field was meant for, and I did not state it clearly enough in the description.

Not asking to reopen — the call on the config surface is yours. Leaving this here for whoever arrives with the same setup.

PR #6662, opened 2026-08-21
finalmask: let quic-go raise the UDP socket buffers on socket-holding mask conns

faryvojnik:
## Problem

The UDP mask conns in `transport/internet/finalmask` hold the socket behind an **interface** field, so `SetReadBuffer` and `SetWriteBuffer` of the underlying `*net.UDPConn` are not promoted.

quic-go raises the socket buffers only if the `net.PacketConn` it is handed asserts to those methods, and it discards the failure — `_ = setReceiveBuffer(pc)` in `sys_conn.go`. The assertion fails against a mask conn, the error goes nowhere, and the socket stays at `net.core.rmem_default` instead of the 8 MiB quic-go asks for.

On a hysteria2 outbound with `obfs=salamander` this shows up as drops on the receive path with no error anywhere: UDP has no feedback, so QUIC just sees loss and retransmits. The user gets a slower connection and the logs are clean.

## Fix

`WrapPacketConnClient` / `WrapPacketConnServer` take and return a dedicated interface instead of `net.PacketConn`:

```go
type PacketConn interface {
	net.PacketConn
	SetReadBuffer(bytes int) error
	SetWriteBuffer(bytes int) error
}
```

Mask conns that embed the socket now embed `finalmask.PacketConn`, so the two setters are promoted and reach the real socket with no code of their own: `headerManagerConn`, `mkcp/header`, `mkcp/aes128gcm`, `mkcp/original`, `header/custom`, `noise`, `salamander`, `xdns` client and server, `realm`. The masks that keep the conn in a named field — `sudoku`'s `udpConn` and `xicmp`'s client/server — delegate in one line each.

Because the interface is the signature, a mask that regresses to `net.PacketConn` stops satisfying `Udpmask` and fails the build, instead of silently dropping the raise again. `transport/internet/memory_settings.go` asserts `instance.(finalmask.Udpmask)` unchecked, so the alternative is a startup panic. `packetconn_test.go` pins that with a compile-time conformance list plus runtime cases that check the setters actually reach the socket through each wrapper shape.

One adapter sits at the boundary for conns that have no setters at all:

```go
func WrapConn(pc net.PacketConn) PacketConn
```

It returns the conn unchanged when it already satisfies `PacketConn`, and wraps it in a no-op otherwise. `internet.FakePacketConn` is the case that matters — it embeds a `net.Conn` interface, so no setter is reachable through it, and the adapter returns nil rather than pretending. `UnwrapUdpMask` looks back through the adapter, so `xicmp`'s rejection of `FakePacketConn` and `udphop.UdpHopPacketConn` still sees the real conn.

Call sites wrapped: `hysteria` dialer and hub, `kcp` dialer, `splithttp` dialer and hub, `udp` dialer and hub, `proxy/wireguard` client and server.

## Why there is no `SyscallConn`

`wrapConn` discards the buffer errors, but it **returns** the error from `SyscallConn` and from the `setDF` that follows it. A wrapper that advertises a `SyscallConn` it cannot deliver therefore aborts `quic.DialEarly` on any path where the conn is not a real `*net.UDPConn` — `internet.FakePacketConn`, i.e. hysteria2 dialled through another outbound. `udphop` can declare it because its conns always are real sockets. The reasoning sits as a comment on `PacketConn` so the trio does not get "completed" later.

## Verification

Measured on a live relay running hysteria2 with `obfs=salamander`. Masked socket with the patch: `rb=16777216`, which is the kernel storing twice quic-go's 8 MiB `DesiredReceiveBufferSize`. Unmasked sockets on the same host, same kernel: `rb=4194304`, i.e. `net.core.rmem_default` untouched. `SetWriteBuffer` takes as well.

No throughput claim: on my load the rcvbuf drop counter does not move, so the change is about the raise reaching the kernel at all, not about a measurable speed win.

Tests: `packetconn_test.go` covers each wrapper shape (embedded interface, named field, adapter) and the no-op case for a conn without the setters. `go build ./...` green on linux, windows and darwin.

### DIFF
--- a/proxy/wireguard/client.go
+++ b/proxy/wireguard/client.go
@@ -27,6 +27,7 @@ import (
 	"github.com/xtls/xray-core/features/stats"
 	"github.com/xtls/xray-core/transport"
 	"github.com/xtls/xray-core/transport/internet"
+	"github.com/xtls/xray-core/transport/internet/finalmask"
 	"golang.zx2c4.com/wireguard/device"
 )
 
@@ -279,7 +280,7 @@ func (h *Handler) init(ctx context.Context) error {
 			panic(reflect.TypeOf(c))
 		}
 		if h.streamSettings.UdpmaskManager != nil {
-			newConn, err := h.streamSettings.UdpmaskManager.WrapPacketConnClient(pktConn)
+			newConn, err := h.streamSettings.UdpmaskManager.WrapPacketConnClient(finalmask.WrapConn(pktConn))
 			if err != nil {
 				pktConn.Close()
 				return nil, errors.New("mask err").Base(err)
--- a/proxy/wireguard/server.go
+++ b/proxy/wireguard/server.go
@@ -23,6 +23,7 @@ import (
 	"github.com/xtls/xray-core/features/stats"
 	"github.com/xtls/xray-core/transport"
 	"github.com/xtls/xray-core/transport/internet"
+	"github.com/xtls/xray-core/transport/internet/finalmask"
 	"github.com/xtls/xray-core/transport/internet/stat"
 	"golang.org/x/crypto/curve25519"
 	"golang.zx2c4.com/wireguard/device"
@@ -263,7 +264,7 @@ func (s *Server) Start() error {
 			return nil, err
 		}
 		if s.streamSettings.UdpmaskManager != nil {
-			newConn, err := s.streamSettings.UdpmaskManager.WrapPacketConnServer(pktConn)
+			newConn, err := s.streamSettings.UdpmaskManager.WrapPacketConnServer(finalmask.WrapConn(pktConn))
 			if err != nil {
 				pktConn.Close()
 				return nil, errors.New("mask err").Base(err)
--- a/transport/internet/finalmask/finalmask.go
+++ b/transport/internet/finalmask/finalmask.go
@@ -9,11 +9,50 @@ import (
 	"github.com/xtls/xray-core/common/errors"
 )
 
+// quic-go raises UDP socket buffers only if the net.PacketConn asserts to SetRead/WriteBuffer; masks hold the socket
+// in an interface field, so nothing is promoted and the raise is lost. SyscallConn is deliberately absent: quic-go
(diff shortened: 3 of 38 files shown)


LjhAUMEM:
SetReadBuffer / SetWriteBuffer 导致的内核丢包严重的话测速应该会发现明显差距，SetWriteBuffer 几乎没影响，SetReadBuffer 个人使用几乎体验不出来，所以之前没弄这种专门针对 quic 的优化，但对于大型机场就不好说了

要添加的话比较好的建议是对 WrapPacketConn 的参数以及返回改用一种新的自定义 Conn Interface

Fangliding:
没有测出实际差距就算了 quic-go这个无脑抛报错日经问题了 而且函数签名理论上不用改?

LjhAUMEM:
> 没有测出实际差距就算了

应该就像我说的个人差别不大，不过是个理论上可以优化的地方，不过关了就关了吧，暂时还不想碰这里，不然和现有的几个 pr 会冲突

> quic-go这个无脑抛报错日经问题了

与其说问题不如是特性，SetReadBuffer / SetWriteBuffer 的错误会忽略，所以实不实现都不影响运行，SyscallConn 如果实现了但报错会停止运行

> 而且函数签名理论上不用改?

可以只为具体 struct 实现，没有声明约束要确保所有 struct 都实现只能翻文件确认

LjhAUMEM:
除了方便检查外还有一个点是如果 WrapPacketConnClient / WrapPacketConnServer 改用自定义 interface + embed 的方式则每个 struct 也只需要改动一行，对初始传入的 net.PacketConn 使用 NewPacketConn 进行包装即可

Fangliding:
觉得速度有问题就自己sysctl改吧 一样的 或者设置sockopt应该也可以

LjhAUMEM:
也可以，anyway，fm 要动的话应该差不多就是我说的方案，我记得我之前还想过几个方案但因为时间太长忘记了

> off-topic

现在距离 fm 剥离出独立库的唯一阻碍应该就剩 udphop 的动态 dial 了

PR #6664, opened 2026-08-22
Fix uTP sniffing never matching and correct the header layout

n0ctal:
## Summary

`SniffUTP` could never return a match, and the header was parsed at the wrong offsets. Both are fixed, with tests.

## Why

The closing check compared `timestamp_microseconds` taken from the packet against `time.Now().UnixMicro()`, with a threshold of `float64(24*time.Hour)` — a nanosecond count:

```go
if math.Abs(float64(time.Now().UnixMicro()-int64(timestamp))) > float64(24*time.Hour) {
    return nil, errNotBittorrent
}
```

A `uint32` timestamp tops out at 4 294 967 295 while the epoch microsecond count is around 1.79e15. Even in the most favourable case the difference is 1 787 389 615 077 627 against a threshold of 86 400 000 000 000 — exceeded 21 times over, and 20 687 times over against 24 hours expressed in microseconds. The condition held for every possible input, so the function always returned `errNotBittorrent` and uTP has never been classified.

The comparison has no meaning beyond the units either: `timestamp_microseconds` carries the sender's own clock from an arbitrary epoch, so it cannot be validated against the receiver's wall clock at all.

Because the positive path never ran, the rest of the parse went unnoticed. The extension chain was walked from offset 2, where BEP 29 places `connection_id`, while extension records begin after the twenty byte fixed header.

## Scope

- read the fixed header fields at their documented offsets
- walk extension records after that header, requiring a selective ack length that is non zero and a multiple of four
- drop the timestamp comparison
- reject a zero connection id and an implausible receive window, so that reviving the sniffer does not pull in other UDP protocols that share the shape of the header
- add tests for the accepted and rejected shapes

## Validation

`go build ./...`, `go vet ./...` and `go test ./common/protocol/...` are clean. The new table covers thirteen cases.

That the old code never matched is shown by the tests themselves: restoring only `bittorrent.go` makes every positive case fail with `not bittorrent header`, including a packet whose timestamp is the largest a `uint32` can hold.

False positives over 1 000 000 generated packets of each kind:

| payload | timestamp check simply removed | this change |
| --- | --- | --- |
| DNS shaped queries | 177 | 1 |
| random UDP of 20 to 80 bytes | 64 | 0 |

A WireGuard handshake initiation is a message type of 1 followed by three reserved zero bytes, which land on the version, extension and connection id fields; it is rejected here. When `proxy/wireguard/bind.go` writes a non zero client id into those same bytes, 17 handshakes per 1 000 000 random client ids still match.

## Risk

Behaviour changes from never matching anything to actually classifying uTP, so a configuration carrying `protocol: ["bittorrent"] -> blocked` will start dropping traffic it previously passed. Sniffing stays a heuristic over a twenty byte header; the measurements above bound what is left.

### DIFF
--- a/common/protocol/bittorrent/bittorrent.go
+++ b/common/protocol/bittorrent/bittorrent.go
@@ -3,11 +3,8 @@ package bittorrent
 import (
 	"encoding/binary"
 	"errors"
-	"math"
-	"time"
 
 	"github.com/xtls/xray-core/common"
-	"github.com/xtls/xray-core/common/buf"
 )
 
 type SniffHeader struct{}
@@ -19,6 +16,14 @@ func (h *SniffHeader) Protocol() string {
 func (h *SniffHeader) Domain() string {
 	return ""
 }
+
+const (
+	utpHeaderSize            = 20
+	utpVersion               = 1
+	utpPacketTypeMax         = 4
+	utpExtensionSelectiveAck = 1
+	utpWindowSizeMax         = 16 << 20
+)
 
 var errNotBittorrent = errors.New("not bittorrent header")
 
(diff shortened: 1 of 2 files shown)


Fangliding:
我自己写一个吧

PR #6668, opened 2026-08-23
Extend BitTorrent sniffer

MakostaDev:
Currently, the BitTorrent sniffer can only detect the BitTorrent Handshake and uTP (uTP is currently broken, but PR #6667 fixes it).

I decided to add detection for several protocols from the BitTorrent family:

- UDP Tracker ([BEP15](https://www.bittorrent.org/beps/bep_0015.html))
- DHT ([BEP5](https://www.bittorrent.org/beps/bep_0005.html) & [BEP42](https://www.bittorrent.org/beps/bep_0042.html))
- LSD ([BEP14](https://www.bittorrent.org/beps/bep_0014.html))

Also `string(b[1:20])` in `SniffBittorrent()` causes an allocation. It was replaced with `bytes.HasPrefix`.

I tested this using Wireshark and BitComet/qBittorrent

### DIFF
--- a/app/dispatcher/sniffer.go
+++ b/app/dispatcher/sniffer.go
@@ -41,6 +41,9 @@ func NewSniffer(ctx context.Context) *Sniffer {
 			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffBittorrent(b) }, false, net.Network_TCP},
 			{func(c context.Context, b []byte) (SniffResult, error) { return quic.SniffQUIC(b) }, false, net.Network_UDP},
 			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffUTP(b) }, false, net.Network_UDP},
+			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffUDPTracker(b) }, false, net.Network_UDP},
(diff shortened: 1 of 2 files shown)


Fangliding:
torrent sniffer 出于兼容性而保留 实际意义不大 它可以被非常轻松的绕过 所以不打算扩展
个人认为管理torrent的合理方法是封锁tracker ip和限制高位udp端口

MakostaDev:
> torrent sniffer 出于兼容性而保留 实际意义不大 

@Fangliding with the help of a webhook, rules can warn/punish users who use torrents **(this approach is used in the Torrent Blocker of the Remnawave Panel)**. And I know that it's impossible to completely block torrents.

When I checked packets from BitComet, I didn't notice a single uTP packet or BitTorrent Handshake (with default settings).

This PR just extends detection. And it contains very little code

Chara-Freedom:
@RPRX why

MakostaDev:
@RPRX What do you think about this? Do you share **Fangliding**’s view?

If users run BitTorrent, it could result in servers (whether VPS or dedicated) being suspended by the service provider due to copyright infringement risks. An example of such a case:
изображение

This PR just adds detection for three protocols, and detecting them is not difficult! Each protocol has distinct byte patterns, which make them easy to detect. So I don’t see a problem with this...

Also, as I said earlier, with the help of a webhook, rules can warn/punish users who use torrents, and this approach is used by Remnawave Panel

PR #6673, opened 2026-08-24
router: fix data race and refactor

MakostaDev:
This currently could trigger a data race between gRPC API calls and `pickRouteInternal()`, as well as in several balancer functions that look up a balancer by tag.

`ReloadRules()` also had an issue: if it failed partway through, rules and balancers that had already been written stayed in place. I changed it to validate and build the new rules/balancers first, and only swap them in afterward. So on error, the router is left in its original, unchanged state.

A few other functions were also cleaned up for a simpler implementation


Chara-Freedom:
@Fangliding please close this for no reason at all like you usually do https://github.com/XTLS/Xray-core/pull/6668.

Chara-Freedom:
Sorry for being rude.

Fangliding:
我重写一份吧 #6678

PR #6678, opened 2026-08-24
Router: Refactor to fix API data race

Fangliding:
修复API调整路由时的竞争问题 顺便简化逻辑

### DIFF
--- a/app/router/balancing_override.go
+++ b/app/router/balancing_override.go
@@ -8,7 +8,7 @@ import (
 
 func (r *Router) OverrideBalancer(balancer string, target string) error {
 	var b *Balancer
-	for tag, bl := range r.balancers {
+	for tag, bl := range *r.balancers.Load() {
 		if tag == balancer {
 			b = bl
 			break
--- a/app/router/balancing_override.go
+++ b/app/router/balancing_override.go
@@ -2,24 +2,7 @@ package router
 
 import (
 	sync "sync"
-
-	"github.com/xtls/xray-core/common/errors"
 )
-
-func (r *Router) OverrideBalancer(balancer string, target string) error {
-	var b *Balancer
-	for tag, bl := range *r.balancers.Load() {
-		if tag == balancer {
-			b = bl
-			break
-		}
-	}
-	if b == nil {
-		return errors.New("balancer '", balancer, "' not found")
-	}
-	b.override.Put(target)
-	return nil
-}
 
 type overrideSettings struct {
 	target string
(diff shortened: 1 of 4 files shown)


app/router/balancing_override.go
MakostaDev: Maybe do this instead:
```go
b, ok := (*r.balancers.Load())[balancer]
if !ok {
    return errors.New("balancer '", balancer, "' not found")
}
```

Also, maybe worth adding a `getBalancer(tag)` function for more readable code, which would replace this and the three `(*r.balancers.Load())[tag]` in `app/router/balancing.go`

app/router/router.go
```diff
@@ -68,26 +72,23 @@ func (r *Router) Init(ctx context.Context, config *Config, d dns.Client, ohm out
...
 			notifier, err := NewWebhookNotifier(wh)
 			if err != nil {
-				r.closeWebhooks()
 				return err
 			}
 			rr.Webhook = notifier
 		}
 		btag := rule.GetBalancingTag()
 		if len(btag) > 0 {
-			brule, found := r.balancers[btag]
```
MakostaDev, line 78: `Init()` and `ReloadRules()` have almost identical structure when iterating over balancers and rules. I think it's worth extracting this into a separate function to keep them in sync. Also, unlike `ReloadRules()`, `Init()` doesn't check for duplicates! 
So it's worth doing at least for the future, in case something changes in `Init()` but gets forgotten in `ReloadRules()`

app/router/balancing_override.go
Fangliding: 它压根没有调用者 删掉好了

app/router/router.go
Fangliding: dedup可以补上
不过我个人的标准是事不过三 当同样的实现需要复制超过三个再抽象到函数

app/router/balancing_override.go
Fangliding: getBalancer(tag) 我觉得并不比 (*r.balancers.Load())[tag] 可读性更佳

app/router/router.go
MakostaDev: > 不过我个人的标准是事不过三 当同样的实现需要复制超过三个再抽象到函数

Hmm, okay. I still reckon it’s worth doing for the future. But I won’t argue about it

app/router/router.go
MakostaDev: `Init()` still doesn't check whether the ruleTag exists

app/router/router.go
Fangliding: 我发现init可以直接调用reloadrule 不用新声明一个函数 这样就没问题了

PR #6679, opened 2026-08-24
Transport: add source-aware PROXY protocol v2

olegmavdeev:
## Summary

- add a source-aware PROXY protocol v2 mode with explicit trusted IP/CIDR prefixes
- optionally restrict PROXY protocol handling to selected inbound listener ports
- reject PROXY v1, LOCAL/UNKNOWN commands, datagram headers, malformed TLVs and invalid endpoints
- preserve PROXY-derived peer addresses over X-Forwarded-For
- reject incompatible transports and legacy acceptProxyProtocol combinations at startup

## Behavior

When `proxyProtocolListenPorts` is empty:

- trusted physical sources must send a valid PROXY protocol v2 header
- untrusted sources bypass PROXY parsing and their bytes remain untouched

When `proxyProtocolListenPorts` is configured:

- selected ports require a trusted physical source and valid PROXY protocol v2
- untrusted connections to selected ports are dropped before application data is read
- other inbound ports remain ordinary direct listeners

## Tests

Added coverage for:

- trusted and untrusted IPv4/IPv6 sources
- same-port and dedicated-port policies
- malformed and fragmented headers
- PROXY v1 rejection
- command, transport, endpoint and TLV validation
- listener accept-loop liveness
- WebSocket, HTTP Upgrade, XHTTP and gRPC address precedence
- unsupported transports and hot-added inbound validation

### DIFF
--- a/app/proxyman/inbound/always.go
+++ b/app/proxyman/inbound/always.go
@@ -68,6 +68,9 @@ func NewAlwaysOnInboundHandler(ctx context.Context, tag string, receiverConfig *
 	if err != nil {
 		return nil, errors.New("failed to parse stream config").Base(err).AtWarning()
 	}
+	if err := internet.ValidateProxyProtocolTransportConfig(mss.SocketSettings, mss.ProtocolSettings); err != nil {
+		return nil, errors.New("invalid PROXY protocol transport configuration").Base(err)
+	}
 
 	newCtx := session.ContextWithInbound(ctx, &session.Inbound{Tag: tag, Source: src})
 	newCtx = session.ContextWithContent(newCtx, &session.Content{SniffingRequest: sniffingRequest})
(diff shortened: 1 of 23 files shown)


Fangliding:
哪有写成这样的。。。
而且不是支持v2么

olegmavdeev:
Thanks for reviewing. To clarify, this PR is not intended to add basic
PROXY protocol v2 parsing. It adds an
authorization boundary for consuming PROXY headers.

Xray first checks the immediate TCP peer address against configured IP/CIDR
ranges. Only a trusted peer is allowed and required to send a strict PROXY v2
header, and only then is the client address carried by that header applied.

This is not a general inbound IP allowlist: it controls which upstream proxies
are trusted to assert a client address.

The goal is to prevent arbitrary peers from spoofing PROXY headers by accepting
them only from configured source IP/CIDR ranges, with strict v2 validation.

Would a smaller change limited to trusted source CIDRs in the system listener be acceptable,
or is this functionality considered out of scope?

PR #6680, opened 2026-08-24
Routing: Export HealthCheckSettings

meanwhile131:
Currently, trying to build a `BurstObservatory` config from `conf.BurstObservatoryConfig`:
```go
config.BurstObservatory = &conf.BurstObservatoryConfig{
	SubjectSelector: []string{"out"},
	HealthCheck: &conf.healthCheckSettings{},
}
```
Results in a compiler error:
```
1. name healthCheckSettings not exported by package conf [UnexportedName]
```
`conf.BurstObservatoryConfig` is public, but uses `conf.healthCheckSettings` (a private struct), which prevents creating a `conf.Config` with a burst observatory from an outside package.
Making `healthCheckSettings` public should allow building of a config with a burst observatory.

### DIFF
--- a/infra/conf/observatory.go
+++ b/infra/conf/observatory.go
@@ -23,7 +23,7 @@ func (o *ObservatoryConfig) Build() (proto.Message, error) {
 type BurstObservatoryConfig struct {
 	SubjectSelector []string `json:"subjectSelector"`
 	// health check settings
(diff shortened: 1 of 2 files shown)


Fangliding (approved):

meanwhile131:
@RPRX Thanks for merging! Could you please consider taking another look at #6697?

PR #6682, opened 2026-08-25
HTTP outbound: Fix `Host` header on CONNECT

1molchuan:
## Summary

- send a `Host` configured in the outbound's `headers` map on the CONNECT request
- keep the CONNECT request-URI pointing at the tunnel target when that happens
- move header application into `applyRequestHeaders` and cover both with tests

## Root cause

`Request.Write` excludes `Header["Host"]` and emits `req.Host` instead, so a
`Host` placed in the `headers` map is silently dropped. An upstream proxy that
authenticates on a `Host` different from the tunnel target cannot be used at
all.

Assigning `req.Host` alone is not a fix either. For CONNECT, `Request.Write`
also derives the request-URI from `req.Host`:

```go
ruri := r.URL.RequestURI()
if r.Method == "CONNECT" && r.URL.Path == "" {
    ruri = host          // host is r.Host when set
    if r.URL.Opaque != "" {
        ruri = r.URL.Opaque
    }
}
```

so the request becomes `CONNECT front.example.com:443` and the tunnel is opened
to the wrong host. That failure is quiet at the HTTP layer — the proxy answers
`200 Connection established` — and only surfaces later as a TLS handshake
landing on the proxy's own certificate. Pinning the real target in `URL.Opaque`
keeps the request-URI intact while the configured value goes out as the `Host`
header.

## Tests

`TestApplyRequestHeadersSendsConfiguredHost` and
`TestApplyRequestHeadersMatchesHostCaseInsensitively` fail without the change.
`TestApplyRequestHeadersKeepsTheTunnelAuthority` additionally fails against the
naive version that only assigns `req.Host`:

```
request line = "CONNECT front.example.com:443 HTTP/1.1", want "CONNECT 203.0.113.7:443 HTTP/1.1"
```

### DIFF
--- a/proxy/http/client.go
+++ b/proxy/http/client.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


Fangliding:
恐怕这样并不符合RFC https://www.rfc-editor.org/info/rfc9112/#section-3.2.3

PR #6683, opened 2026-08-25
HTTP outbound: Fix `headers` for requests without an inbound

1molchuan:
## Summary

- fall back to an unspecified source when the context carries no inbound, instead of refusing to build the headers

## Root cause

`fillRequestHeader` returns an error when there is no inbound in the context.
`burstObservatory` dispatches its health checks through `tagged.Dialer`, which
carries no inbound, so an HTTP outbound with a configured `headers` map fails
every probe. Because it fails during dispatch, a balancer never collects a
sample for that outbound at all — with `leastPing` it keeps using
`fallbackTag`, and the outbound is never actually rated. Every probe reports
`io: read/write on closed pipe`, including probes of outbounds that are
perfectly healthy.

## Note on an earlier version of this PR

This originally also guarded `outbounds[len(outbounds)-1]` and claimed it could
panic. That was wrong: `Process` already indexes the same expression before
this is reached, so outbounds is never empty here. The panic was only reachable
by calling the unexported function directly from a test, which is not a real
code path. That part has been dropped, along with the conditional that only
demanded metadata for templated values — falling back to `0.0.0.0:0` is
simpler and covers the case that actually occurs.

## Tests

`TestFillRequestHeaderWithoutInbound` fails on `main` with
`missing inbound or outbound metadata from context`. The others cover that a
present inbound is still used, and that an empty header list is a no-op.

### DIFF
--- a/proxy/http/client.go
+++ b/proxy/http/client.go
@@ -169,19 +169,22 @@ func fillRequestHeader(ctx context.Context, header []*Header) ([]*Header, error)
 		return header, nil
 	}
 
-	inbound := session.InboundFromContext(ctx)
 	outbounds := session.OutboundsFromContext(ctx)
 	ob := outbounds[len(outbounds)-1]
 
-	if inbound == nil || ob == nil {
-		return nil, errors.New("missing inbound or outbound metadata from context")
+	// A request can reach here without an inbound, for example a health check
+	// dispatched through tagged.Dialer. Fall back to an unspecified source
+	// rather than refusing to build the headers.
+	source := net.TCPDestination(net.AnyIP, 0)
+	if inbound := session.InboundFromContext(ctx); inbound != nil {
+		source = inbound.Source
 	}
 
 	data := struct {
 		Source net.Destination
 		Target net.Destination
 	}{
-		Source: inbound.Source,
+		Source: source,
 		Target: ob.Target,
 	}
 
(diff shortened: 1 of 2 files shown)


Fangliding:
outbound已经事先检查过了 inbound我觉得回落到0.0.0.0:0更加好一点 这个功能我都不知道大概没人用 只要解决无源请求会炸的问题就行了

1molchuan:
你说得对，outbound 那里确实已经检查过了 —— `Process` 开头就取了同一个下标，所以走到这里 outbounds 不可能为空。我原先说会 panic 是错的，那只有在测试里直接调用未导出函数才触发，不是真实路径。已经改掉，PR 描述里也说明了。

现在按你的建议改成最小修复：inbound 缺失时 source 回落到 `0.0.0.0:0`，模板照常渲染，不再有那个条件判断。outbound 的取值保持原样没动。

`{{.Source}}` 这个功能可能确实没人用，不过健康检查那条路径会踩到它 —— 只要 http outbound 配了 `headers` 又被 balancer 探测，每次探测都在分发阶段失败，`leastPing` 就永远拿不到样本、一直退回 `fallbackTag`。

Fangliding:
这机器人。。。 我都已经改完了 算了这种小改动我自己放另一个PR里吧

PR #6688, opened 2026-08-27
Transport: Bind the UDP outbound socket in the destination family

adsdevdoo:
The generic half I mentioned in #6624.

An outbound UDP socket with no sendThrough address is bound to the 0.0.0.0
wildcard. On most platforms that's fine: the runtime turns it into a
dual-stack socket on [::] (see ipToSockaddrInet6 in net/ipsock_posix.go), so
IPv6 destinations work anyway. OpenBSD and DragonFly have no IPv4-mapped
addresses at all, there the socket stays AF_INET and a send to an IPv6 peer
fails with "non-IPv4 address".

So bind the wildcard of the destination's own family instead: 0.0.0.0 for a
v4 destination, :: for a v6 one. On platforms with the mapping both wildcards
produce exactly the same dual-stack socket as before, nothing changes there.
The network string passed to ListenPacket is untouched too, so sockopt
controllers keep getting the same arguments.

Tested on OpenBSD 7.7/7.8/7.9 amd64: with this and #6624 the transparent
proxy handles IPv6 UDP (DNS, QUIC) end to end on a live router.

### DIFF
--- a/transport/internet/system_dialer.go
+++ b/transport/internet/system_dialer.go
@@ -50,18 +50,26 @@ func (d *DefaultSystemDialer) Dial(ctx context.Context, src net.Address, dest ne
 	errors.LogDebug(ctx, "dialing to "+dest.String())
 
 	if dest.Network == net.Network_UDP {
-		srcAddr := resolveSrcAddr(net.Network_UDP, src)
-		if srcAddr == nil {
-			srcAddr = &net.UDPAddr{
-				IP:   []byte{0, 0, 0, 0},
-				Port: 0,
-			}
-		}
-		var lc net.ListenConfig
 		destAddr, err := net.ResolveUDPAddr("udp", dest.NetAddr())
 		if err != nil {
 			return nil, err
 		}
+		srcAddr := resolveSrcAddr(net.Network_UDP, src)
+		if srcAddr == nil {
+			// Bind the wildcard of the destination's own address family.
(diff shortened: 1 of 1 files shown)


Fangliding:
不是 怎么上个pr放了这么久刚合一天都没有又追加一个 两次了 不能一起弄完吗

adsdevdoo:
That was the plan agreed in #6624, actually: keep that one OpenBSD-only so it could be merged quickly, and send this piece separately because it touches every platform and deserves its own review. It was already mentioned there and RPRX was fine with it.

Fangliding:
好吧 我看到了 bsd破事真多

RPRX:
改成这样写在其它系统上也没问题，不过下面应用 sockopt 或许会受影响？可以先合了看看

RPRX:
> 好吧 我看到了 bsd破事真多

~~或许只是因为 OpenBSD 追求严谨性安全性，可能在 FreeBSD 上单个 socket 能双栈~~

Fangliding:
正常系统不会有影响 正常系统一个ipv6 socket可以同时管ipv4和v6 对ipv4的操作自动映射到一个划分好的ipv6段 所以go的行为是无论0000还是::全部用 :: 然后使用双栈特性 oepnbsd不支持而已 理由还是security reason 就跟不让用syscall一样

PR #6691, opened 2026-08-27
TUN inbound: Support `autoSystemRoutingTable` and `autoOutboundsInterface` on FreeBSD as well

brookwko:
## What

Implements the two auto-route options of the TUN inbound on FreeBSD, the last platform where they were still stubs. This closes the FreeBSD gap left open in #5594 after Windows (#5887) and macOS/Linux (#6366) were done.

- `autoSystemRoutingTable`: installs the configured CIDRs as routes into the TUN. A `0.0.0.0/0` / `::/0` entry is split into eight more-specific prefixes (`1.0.0.0/8 … 128.0.0.0/1`, v6 mirrored) so the system's real default route survives for outbound-interface discovery — same scheme as the darwin implementation.
- `autoOutboundsInterface` (`"auto"` or a fixed name): steers outbound sockets back onto the physical egress so they don't loop through the TUN.
- `gateway`: assigns the point-to-point address pair to the TUN (previously ignored on FreeBSD).

## The one design point worth reviewing: outbound escape via a second FIB

darwin/linux bind each outbound socket to the egress interface (`IP_BOUND_IF` / `SO_BINDTODEVICE`). FreeBSD has no per-socket interface bind, so instead it uses a dedicated routing table (FIB):

- `Start()` mirrors the physical default route (and the connected prefix its gateway resolves through) into FIB 1, keeping it in sync via a routing-socket monitor.
- `setinterface` marks each outbound socket with `setsockopt(SO_SETFIB, 1)` — the same per-socket table selection `route(8) -fib` uses.

This needs the boot tunable `net.fibs>=2` (a routing table is a boot-time resource on FreeBSD). If it's requested without that, `NewTun` fails early with a message telling the user to add `net.fibs=2` to `/boot/loader.conf` — before any dialer controller is registered, so a failed start can't leave outbound sockets pointed at a table that was never set up. Reserving FIB 1 and requiring `net.fibs>=2` mirrors established FreeBSD tunnel practice (cf. wg-quick's `tunnelfib`).

System routes use interface routes rather than gateway routes: the tun(4) device is a broadcast interface whose point-to-point peer address doubles as the subnet broadcast, so routing *to* that gateway is refused with EACCES — interface routes sidestep it (what wg-quick does on FreeBSD).

## Testing

Verified end-to-end on a FreeBSD 14.4-RELEASE VM (`net.fibs=2`), full-tunnel config (`autoSystemRoutingTable: ["0.0.0.0/0"]`, `autoOutboundsInterface` unset → `"auto"`):

- The TUN comes up with the configured address; the eight split routes land in FIB 0 pointing at the TUN; the real default route is untouched.
- A `fetch` to an arbitrary public IP from inside the guest travels through the TUN, out via the SO_SETFIB-escaped outbound, to the upstream server and back (proves the escape prevents the outbound from looping into the TUN).
- FIB 1 shows the mirrored connected + default routes.
- A route-table change triggers `[tun] update interface` and re-syncs FIB 1.
- On shutdown every route and the interface are removed and a restart succeeds (the interface is explicitly destroyed — the wireguard tun library leaves it behind on FreeBSD otherwise).
- Package unit tests pass on FreeBSD.

`GOOS=freebsd go build ./...`, `go vet`, and the gofumpt check are all clean; the host test suite is unaffected.

## Not in scope (possible follow-ups)

- `sockopt.interface` (transport-level) is still unimplemented on FreeBSD.
- The BSD route-socket helpers are duplicated with tun_darwin.go per the existing per-platform file style; a shared `darwin || freebsd` file could dedup them later.
- The escape FIB number is fixed at 1; a config knob could make it selectable.

Refs #5594, #6366, #5887.

### DIFF
--- a/proxy/tun/tun_freebsd.go
+++ b/proxy/tun/tun_freebsd.go
@@ -3,28 +3,60 @@
 package tun
 
 import (
+	"context"
 	"errors"
 	"net"
-	_ "unsafe"
+	"net/netip"
+	"os"
+	"slices"
+	"sync"
+	"unsafe"
 
 	"golang.zx2c4.com/wireguard/tun"
 	"gvisor.dev/gvisor/pkg/buffer"
 	"gvisor.dev/gvisor/pkg/tcpip"
 	"gvisor.dev/gvisor/pkg/tcpip/stack"
 
+	"golang.org/x/net/route"
 	"golang.org/x/sys/unix"
 
 	"github.com/xtls/xray-core/common/buf"
+	xerrors "github.com/xtls/xray-core/common/errors"
 )
 
-const tunHeaderSize = 4
+const (
+	tunHeaderSize         = 4
+	defaultFreeBSDGateway = "169.254.10.1/30"
+
+	// escapeFib is the routing table outbound sockets are switched to so
+	// their traffic bypasses the TUN routes installed in the default FIB
+	// (FreeBSD's substitute for the per-socket interface binding other
+	// platforms use). Requires the boot tunable net.fibs >= 2.
+	escapeFib = 1
+)
 
 //go:linkname procyield runtime.procyield
 func procyield(cycles uint32)
 
 type FreeBSDTun struct {
-	device tun.Device
-	mtu    uint32
+	device        tun.Device
+	options       *Config
+	tunIndex      int
+	autoInterface bool
+
+	systemRoutes     []netip.Prefix
+	escapeMu         sync.Mutex
+	escapeRoutes     []escapeRoute
+	routeMonitor     *os.File
+	routeMonitorOnce sync.Once
+}
+
+// escapeRoute remembers one route written into the escape FIB, in the exact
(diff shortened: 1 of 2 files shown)


RPRX:
首先 https://github.com/XTLS/Xray-core/pull/6624#issuecomment-5413492355 ，~~虽然不知道这代码是啥写的但即使是 AI 写的也认了，反正特定系统~~

其次我大概了解一些 FreeBSD 和 OpenBSD 了，前者应用范围更广，后者主要应用于网关，~~怪不得前者有人 PR TUN、后者有人 PR TPROXY，我觉得小众系统相对于 Linux 的优势主要是挖洞的人更少吧，不过 AI 出来了也难说，但至少后者用在网关上感觉比 Linux 更放心~~

最后 https://github.com/XTLS/Xray-core/pull/6624#issuecomment-5431059650 ，先合并了，存在 bug 的话真正用它的人自会反馈

PR #6694, opened 2026-08-28
XHTTP client: Fix a data race in WaitReadCloser

Eninspace:
## Problem

`WaitReadCloser.Read` and `Close` read the `ReadCloser` field without synchronizing with the write in `Set`. An interface value is two words; a concurrent reader can observe the type word already set while the data word is still nil. The `== nil` check passes and the call then dereferences a nil receiver.

Production crash, v26.7.11, linux/amd64, go1.26.5, XHTTP outbound:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa4bf15]

goroutine 32692068 [running]:
golang.org/x/net/http2.transportResponseBody.Read({0x0?}, ...)
	golang.org/x/net@v0.57.0/http2/transport.go:2175 +0x35
github.com/xtls/xray-core/transport/internet/splithttp.(*WaitReadCloser).Read(...)
	transport/internet/splithttp/client.go:210 +0x64
github.com/xtls/xray-core/transport/internet/splithttp.(*splitConn).Read(...)
	transport/internet/splithttp/connection.go:22 +0x22
github.com/xtls/xray-core/common/buf.(*Buffer).ReadFullFrom(...)
	common/buf/buffer.go:340
github.com/xtls/xray-core/proxy/vless/encoding.DecodeResponseHeader(...)
	proxy/vless/encoding/encoding.go:159
github.com/xtls/xray-core/proxy/vless/outbound.(*Handler).Process.func5()
	proxy/vless/outbound/outbound.go:384
```

The receiver is `transportResponseBody{0x0}` — a non-nil interface holding a struct whose `cs *clientStream` is nil, which is what a torn read of an interface value looks like.

## Why waiting on the channel is not enough

`w.Wait` is closed by **two** parties: `Set` and `Close`. Receiving from it only orders the receiver against whoever closed it. On the failure paths in `OpenStream` (dial error, non-200 status, uploadOnly) it is `Close` that closes the channel, so a reader waking up there has no happens-before edge with the write in `Set` and can still observe a torn value. Simply moving the channel receive before the nil check does not fix this — the field itself has to be synchronized.

## Fix

Store the body in an `atomic.Pointer`. `Set` keeps its existing ownership semantics, including the `recover` path that closes the body when the stream was already abandoned.

## Test

`Test_regression_WaitReadCloser_race` runs `Set`, `Read` and `Close` concurrently, the way `OpenStream` does. It fails deterministically under `-race`. Since CI runs without `-race`, the body also mirrors `http2.transportResponseBody` — a struct with one dereferenced pointer — so a torn read panics there instead: caught on `main` in 12 of 12 runs. Adds 0.34s.

## Measurements

A simulation of the real call sites (`dialer.go:458` -> `connection.go:20`/`:33` -> `outbound.go:203`, including the second `reader.Close()` that `splitConn.Close` performs without a guard), 5M iterations per variant:

| | nil deref in `Read` | nil deref in `Close` |
|---|---|---|
| `main` | 22-41 | 132-243 |
| this change | 0 | 0 |

The struct shrinks from 24 to 16 bytes. The hot `Read` path costs about 6ns more, almost all of it the unconditional channel receive rather than the atomic load; a mutex-based variant measured within 1ns of this one.


Eninspace:
Measured on the current head of this branch, with a single closer (`defer conn.Close()` in `proxy/vless/outbound/outbound.go`): 2M iterations of `Set` racing `Close` give **637 double closes** of the response body; with the change below it is **0**. The exact count is scheduling-dependent, but it is never zero without the change, and both the pre-refactor code and this fix are clean. `-race` finds nothing either way.

It matters because `transportResponseBody.Close` re-credits the **connection-level** flow window when the body still has unread bytes. On a byte-for-byte copy of `x/net/http2/pipe.go`, a body with 11 unread bytes returns 22: `closeWithError` detaches the buffer but keeps `unread`, and `Len()` reports it, so the second `Close` sees the same value. A fully read body is unaffected.

The change is `Load` → `Swap(nil)` in `Close` and in the `Done()` branch of `Set`, so whichever side takes the body is the one that closes it:

```go
func (w *WaitReadCloser) Set(rc io.ReadCloser) {
	w.reader.Store(&rc)
	if w.wait.Done() {
		if p := w.reader.Swap(nil); p != nil {
			(*p).Close()
		}
	}
	w.wait.Close()
}

func (w *WaitReadCloser) Close() error {
	w.wait.Close()
	if p := w.reader.Swap(nil); p != nil {
		return (*p).Close()
	}
	return nil
}
```

One behaviour change worth naming: after `Close` the field is empty, so a later `Read` takes the existing `rc == nil` path and returns `io.ErrClosedPipe` instead of reading from an already closed body.

Medium1992:
Confirming this race is hit in production, not only in theory.

**Build:** XTLS/Xray-core `cd4ce973` (main), Go 1.27.1, linux/amd64.
Includes #6665, does not include this PR.

**Conditions:** a node carrying roughly 5000 concurrent connections,
VLESS with encryption over XHTTP (stream-down).

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x485ae0]

goroutine 8467353 [running]:
net/http/internal/http2.transportResponseBody.Read({0x0?}, {0x9b9b0fc8e40?, 0x6?, 0x8?})
net/http/internal/http2/transport.go:2350 +0x30
github.com/xtls/xray-core/transport/internet/splithttp.(*WaitReadCloser).Read(...)
transport/internet/splithttp/client.go:211 +0x68
github.com/xtls/xray-core/transport/internet/splithttp.(*splitConn).Read(...)
transport/internet/splithttp/connection.go:22 +0x28
github.com/xtls/xray-core/proxy/vless/encryption.(*XorConn).Read(...)
proxy/vless/encryption/xor.go:73 +0x3c
io.ReadFull(...)
github.com/xtls/xray-core/proxy/vless/encryption.(*CommonConn).Read(...)
proxy/vless/encryption/common.go:86 +0xd0
github.com/xtls/xray-core/common/buf.(*Buffer).ReadFullFrom(...)
common/buf/buffer.go:340 +0x144
github.com/xtls/xray-core/proxy/vless/encoding.DecodeResponseHeader(...)
proxy/vless/encoding/encoding.go:159 +0xcc
github.com/xtls/xray-core/proxy/vless/outbound.(*Handler).Process.func5()
proxy/vless/outbound/outbound.go:384 +0xbc



The receiver printed for `transportResponseBody.Read` is `{0x0}`: the call
dispatched on a valid type word with a nil data word. That is precisely the
torn interface read this PR removes.

At `cd4ce973`, `client.go:211` is `return w.ReadCloser.Read(b)`, reached
through the unsynchronized fast path on line 206. When that check observes
the type word `Set` has already written but not yet its data word, it skips
`<-w.Wait` entirely, so no happens-before edge with the writing goroutine is
established and the half-written interface is called.

Per connection the window is tiny, but at a few thousand connections per node
it becomes a routine crash.

Glad to run a build with this patch applied and report back.

RPRX:
既然测试有效就先合了，依稀记得 core 里别的地方也有类似的写法，可以看看那些地方有没有这种问题

RPRX:
> An interface value is two words; a concurrent reader can observe the type word already set while the data word is still nil.

~~细看了一眼，确实是典型的 data race，咋这么久了没人报过，我想了下若改成穿进去只判断 data word 是否 nil 会有别的问题~~

https://github.com/XTLS/Xray-core/pull/6665#issuecomment-5429951520 只讨论了 bool，我查了下现代硬件（单行汇编）即使读写 int 也是原子的，但是加解锁能强制刷新 CPU 缓存

PR #6698, opened 2026-08-29
SS2022 outbound: Close connection to server on early returns

JorikSmith:
Process dials server but never closes connection. On normal path sing's
CopyConn cleanup handles it, but three returns before that (read payload,
write payload, client handshake) leave it open. First one only needs client
to hang up inside 100ms ReadMultiBufferTimeout, which returns io.EOF.

Every other outbound does defer conn.Close() right after dialing.


RPRX:
@Fangliding 话说有没有兴趣把隔壁的 SS2022 copy 过来并 vibe 个服务端

Fangliding:
没有 claude到期了(
而且这种小问题可能也没法避免 这个pr倒也不是sing的问题

RPRX:
@Fangliding 用 GitHub 送的 copilot 吧

PR #6703, opened 2026-08-30
Sniffer: refactor HTTP

MakostaDev:
### Changes:
- If the request line uses `absolute-form` or `authority-form`, the sniffer now attempts to extract the Host from the URI. Fallback to the Host header otherwise..
- `PATCH` and `TRACE` methods are now recognized.
- The `:path` attribute can now be populated from more sources. I also considered stripping `?` and `#`, but it turns out attribute value matching in routing rules uses regexp, so users can already configure this flexibly on their own.
- A few minor fixes for more correct sniffing overall.

Tests were added for the new changes. Existing tests were left untouched, even though they use `\n` instead of `\r\n`. There was also a bug in the test file where, on failure, it would call a method on a variable that was nil


Fangliding:
为了这两个少见的请求写成这样还是不太行 url相关估计本来就没多少用的

PR #6706, opened 2026-08-31
fix(proxy): synchronize Vision TLS classification

svlavr:
## Summary

- serialize mutable TLS classification shared by the concurrent Vision reader and writer
- publish an immutable final snapshot for the established data path
- add concurrent packet-filtering and TLS-classification regression tests

## Problem

`VisionReader` and `VisionWriter` operate concurrently on the same `TrafficState`. Both paths read and update the packet counter and TLS classification fields without synchronization. The race detector reports conflicting accesses, and the writer can observe a mixture of classifier states while deciding padding and direct-copy behavior.

## Implementation

The mutable classification phase is protected by a mutex. Once classification finishes, a coherent immutable snapshot is published atomically, so established connections do not take the mutex on the data path. The exported `XtlsFilterTls` entry point uses the same synchronization.

## Validation

- `go vet ./proxy`
- `go test ./proxy/... -count=1`
- `go test -race ./proxy -run '^TestVisionTrafficStateConcurrent' -count=100`
- `go test -race ./testing/scenarios -run '^TestVlessXtlsVision$' -count=1`
- `go test -race ./testing/scenarios -run '^TestVlessXtlsVisionReality$' -count=1`

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -13,6 +13,8 @@ import (
 	"math/big"
 	"runtime"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pires/go-proxyproto"
@@ -102,6 +104,11 @@ type GetOutbound interface {
 // TrafficState is used to track uplink and downlink of one connection
 // It is used by XTLS to determine if switch to raw copy mode, It is used by Vision to calculate padding
 type TrafficState struct {
+	// TLS classification is shared by the concurrent Vision reader and writer.
+	// Keep the mutable phase serialized and publish an immutable final snapshot
+	// so established connections avoid a mutex on the data path.
+	filterAccess           sync.Mutex
+	finalTLSState          atomic.Pointer[trafficStateSnapshot]
 	UserUUID               []byte
 	NumberOfPacketToFilter int
 	EnableXtls             bool
@@ -111,6 +118,55 @@ type TrafficState struct {
 	RemainingServerHello   int32
 	Inbound                InboundState
 	Outbound               OutboundState
+}
+
+type trafficStateSnapshot struct {
+	numberOfPacketToFilter int
+	enableXtls             bool
+	isTLS12orAbove         bool
+	isTLS                  bool
+}
+
+func (s *TrafficState) snapshotLocked() trafficStateSnapshot {
+	return trafficStateSnapshot{
+		numberOfPacketToFilter: s.NumberOfPacketToFilter,
+		enableXtls:             s.EnableXtls,
+		isTLS12orAbove:         s.IsTLS12orAbove,
+		isTLS:                  s.IsTLS,
+	}
+}
+
+func (s *TrafficState) publishFinalTLSStateLocked(snapshot trafficStateSnapshot) {
+	if snapshot.numberOfPacketToFilter <= 0 {
+		s.finalTLSState.Store(&snapshot)
+	} else {
+		s.finalTLSState.Store(nil)
+	}
+}
+
+func (s *TrafficState) isFilteringTLS() bool {
+	if s.finalTLSState.Load() != nil {
+		return false
+	}
+	s.filterAccess.Lock()
+	defer s.filterAccess.Unlock()
+	snapshot := s.snapshotLocked()
+	s.publishFinalTLSStateLocked(snapshot)
+	return snapshot.numberOfPacketToFilter > 0
+}
+
+func (s *TrafficState) filterTLSIfNeeded(buffer buf.MultiBuffer, ctx context.Context) trafficStateSnapshot {
+	if snapshot := s.finalTLSState.Load(); snapshot != nil {
+		return *snapshot
+	}
+	s.filterAccess.Lock()
+	defer s.filterAccess.Unlock()
+	if s.NumberOfPacketToFilter > 0 {
+		xtlsFilterTLS(buffer, s, ctx)
+	}
+	snapshot := s.snapshotLocked()
+	s.publishFinalTLSStateLocked(snapshot)
+	return snapshot
 }
 
 type InboundState struct {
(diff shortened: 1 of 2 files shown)


Fangliding:
此类修复需要具体触发路径 tls建立握手的初期包应该不会两个方向同时抵达 以及这几个简单的类型并发好像也没事

svlavr:
@Fangliding Thank you. Your concern is valid: the original PR only demonstrated concurrent access with a synthetic unit test and did not provide a concrete runtime trigger path.

I have now reproduced the same TrafficState race through real localhost VLESS Vision traffic using only official Xray components on upstream commit 5e245b082e6be8c8899c34410f488e8ab001aaba.

The important point is that the trigger does not require TLS ClientHello and ServerHello packets to arrive simultaneously. The outer TLS handshake is already complete. The race occurs while Vision is classifying the first buffers of the proxied application traffic.

A concrete trigger is a normal server-first or duplex TCP protocol:

1. A VLESS Vision client establishes the outer TLS connection.
2. The client requests a TCP connection to the destination.
3. The destination accepts the connection and immediately sends a greeting/banner.
4. At the same time, the client continues writing small application payloads.
5. The VLESS inbound creates one TrafficState and passes the same pointer to the uplink VisionReader and downlink VisionWriter.
6. Freedom processes both transfer directions concurrently through task.Run.

For the reproduction, the destination is testing/servers/tcp.Server with SendFirst enabled. Both Xray server and client are started in-process with core.New(...).Start(). All listeners and connections use 127.0.0.1 and dynamically selected local ports. The client uses core.Dial and performs concurrent reads and small writes. The test does not call VisionReader, VisionWriter, or XtlsFilterTls directly.

Running the Xray instances in-process is important. The regular scenario harness uses InitializeServerConfigs, which builds a separate child Xray with plain go build. Therefore, go test -race instruments the parent scenario test but does not instrument the child core process. A passing regular scenario test under -race is not negative race evidence for the core code. In the new reproduction, Handler.Process, task.Run, VisionReader, and VisionWriter all execute inside the race-instrumented test binary.

The normal scenario succeeds and verifies the complete server banner and all echoed client payloads:

    go test ./testing/scenarios -run '^TestUpstreamVlessVisionLocalhostConcurrentTraffic$' -count=1 -timeout 2m

The same real-traffic scenario under the race detector reports:

    WARNING: DATA RACE

    Read:
      github.com/xtls/xray-core/proxy.(*VisionReader).ReadMultiBuffer()
          proxy/proxy.go:255

    Previous write:
      github.com/xtls/xray-core/proxy.XtlsFilterTls()
          proxy/proxy.go:624
      github.com/xtls/xray-core/proxy.(*VisionWriter).WriteMultiBuffer()
          proxy/proxy.go:353

Both goroutines originate from the actual Freedom transfer path:

    proxy/freedom.(*Handler).Process.func4()
    proxy/freedom.(*Handler).Process.func5()
    common/task.Run()

The reverse conflicting pair is also reported:

    VisionWriter.WriteMultiBuffer  proxy/proxy.go:352  read
    VisionReader.ReadMultiBuffer   proxy/proxy.go:256
    XtlsFilterTls                  proxy/proxy.go:624  write

The test creates 24 independent VLESS Vision connections per run. Repeating it with -count=3, for 72 connections in total, reports the same conflict again.

The race command uses:

    go test -race -gcflags=all=-d=checkptr=0 ./testing/scenarios -run '^TestUpstreamVlessVisionLocalhostConcurrentTraffic$' -count=3 -timeout 3m

The checkptr option is needed because the existing Vision VLESS path performs unsafe pointer arithmetic while accessing the wrapped TLS connection internals; otherwise checkptr aborts in proxy/vless/outbound/outbound.go before the traffic reaches the classification race. This option disables checkptr only. The race detector remains enabled for the entire in-process core.

I also set OneTimeLoading on the generated test certificate to avoid mixing this result with an unrelated TLS certificate/OCSP ticker race.

Regarding the fields being simple types: NumberOfPacketToFilter-- is a read-modify-write operation, while the other goroutine reads and decrements the same field. This is a data race under the Go memory model even when the integer is naturally aligned. XtlsFilterTls also updates RemainingServerHello, IsTLS12orAbove, IsTLS, Cipher, and EnableXtls, while the writer uses the classification state for padding and direct-copy decisions.

The race window is short because classification stops after the first buffers, but a new TrafficState is created for every Vision connection, so the same window is recreated for each connection.

I have not observed payload corruption, a crash, or a connection failure in the normal build, so I am not claiming a demonstrated user-visible failure. The bounded conclusion is that unsynchronized TrafficState access is reachable through real supported localhost traffic, rather than only through a synthetic direct-call test.

The runtime stack above is specifically the server-side path:

    VLESS inbound -> VisionReader/VisionWriter -> Freedom -> task.Run

The analogous client outbound path remains statically reachable through its shared TrafficState and concurrent postRequest/getResponse tasks, but I am not presenting the server-side runtime stack as proof of a separate client-side runtime occurrence.

The minimized in-process scenario test is ready locally. I can add it to the PR so the trigger and race stack are directly reviewable and reproducible.

svlavr:
@Fangliding @yuhan6665 I pushed the concrete regression scenario to the PR head branch in [commit cd30c38f](https://github.com/svlavr/Xray-core/commit/cd30c38feab7204498e64552233c423ccf23a0d7).

Because the PR is currently closed, GitHub is still displaying its previous head commit (`1a707b65`). The new test commit is available directly at the link above and should become part of the PR diff when the PR is reopened.

The test runs the VLESS Vision client and server in-process, uses only loopback connections, and exchanges a server-first banner concurrently with client payloads. It does not call the Vision reader, writer, or classifier directly.

Validation:

- Before the synchronization change, the diagnostic run reports the conflicting TrafficState access through the real VLESS inbound -> Freedom path.
- With the synchronization change, the normal test passes and the diagnostic run passes for three repetitions.
- The test also verifies the complete banner and every echoed payload.

Commands:

    go test ./testing/scenarios -run '^TestVlessXtlsVisionConcurrentBidirectionalTraffic$' -count=1 -timeout 2m
    go test -race -gcflags=all=-d=checkptr=0 ./testing/scenarios -run '^TestVlessXtlsVisionConcurrentBidirectionalTraffic$' -count=3 -timeout 3m

Could you please take another look and reopen the PR if this addresses the missing concrete trigger path?

PR #6711, opened 2026-09-02
feat(blackhole): add configurable HTTP status code

raaad1on:
## Summary

This PR adds support for configuring the HTTP status code returned by the blackhole outbound protocol when using HTTP response type.

## Problem

Currently, the blackhole protocol with HTTP response always returns `403 Forbidden`. This limitation makes it unsuitable for several use cases:

- **Captive portal/connectivity checks**: Mobile devices and browsers use specific URLs to detect internet connectivity (e.g., `connectivitycheck.gstatic.com`, `captive.apple.com`). These checks expect `204 No Content` responses to confirm connectivity.
- **Custom HTTP responses**: Some scenarios require different status codes like `200 OK` or `404 Not Found`.

## Solution

Add a `statusCode` field to the HTTP response configuration that allows specifying any valid HTTP status code. The default behavior (`403 Forbidden`) is preserved for backward compatibility when no status code is specified.

### Configuration Example

```json
{
  "outbounds": [{
    "tag": "block-204",
    "protocol": "blackhole",
    "settings": {
      "response": {
        "type": "http",
        "statusCode": 204
      }
    }
  }]
}
```

### Use Case: Local Connectivity Check

```json
{
  "routing": {
    "rules": [{
      "domain": [
        "domain:gstatic.com",
        "domain:msftconnecttest.com",
        "domain:apple.com"
      ],
      "outboundTag": "connectivity-check"
    }]
  },
  "outbounds": [{
    "tag": "connectivity-check",
    "protocol": "blackhole",
    "settings": {
      "response": {
        "type": "http",
        "statusCode": 204
      }
    }
  }]
}
```

## Changes

- **proxy/blackhole/config.proto**: Add `status_code` field to `HTTPResponse` message
- **proxy/blackhole/config.pb.go**: Regenerated protobuf code
- **proxy/blackhole/config.go**: Update `WriteTo()` method to use configured status code with proper HTTP status text lookup via `http.StatusText()`
- **infra/conf/blackhole.go**: Add `statusCode` JSON field support for configuration parsing
- **Tests**: Add comprehensive tests for various status codes (200, 204, 403, 404, 500)

## Backward Compatibility

This change is fully backward compatible:
- When `statusCode` is not specified (or set to 0), the default `403 Forbidden` response is used
- Existing configurations continue to work without modification
- The response format remains the same, only the status code and status text are configurable

## Testing

All existing tests pass, and new tests have been added to verify:
- Default behavior (403) when no status code is specified
[...]

PR #6713, opened 2026-09-02
Blackhole outbound: Add customizable response data support

Fangliding:
#6711
允许回复自定义数据
把之前稀奇古怪和proto糅合在一起的奇怪拓展结构清理了一下 现在干净多了（加功能加test结果diff加一起还是负的）
因为之前的结构和proto糅杂所以break了proto（应该没啥人用这个吧）config结构不变

### DIFF
--- a/infra/conf/blackhole.go
+++ b/infra/conf/blackhole.go
@@ -1,52 +1,42 @@
 package conf
 
 import (
-	"encoding/json"
+	"encoding/base64"
+	"strings"
 
 	"github.com/xtls/xray-core/common/errors"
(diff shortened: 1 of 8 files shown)


needthisforctf:
This would be way more usable if there was an api handle to set/change the data for a certain blackhole outbound. In it's current form the usage is very limited since it's static and each change would need for the core to be reloaded.

Fangliding:
本来就有移除出站增加出站的API

needthisforctf:
Hmmm, yeah, you're right, that's one way to do it.

PR #6715, opened 2026-09-03
Sniffer: Avoid DNS false positives in SniffUTP()

Immortalety:
## Problem

A valid 20-byte DNS query like this can be detected as a uTP `ST_SYN` packet:

```text
Transaction ID: 0x4100
Question:       aa. IN A
Packet length:  20 bytes
```

As a result, `SniffUTP` returns a BitTorrent match for ordinary DNS traffic.

## Fix

Known uTP implementations set `ack_nr` to `0` in `ST_SYN`, since no packets from the peer have been received yet. Validate that field as part of the sniffing check:

```go
if binary.BigEndian.Uint16(b[18:20]) != 0 {
    return nil, errNotBittorrent
}
```

The DNS query above has `QCLASS = 0x0001` in the same two bytes, so it is rejected while normal uTP SYN packets continue to match.


Fangliding:
我曾经考虑过 但是怎么会存在 aa（或者其他两字查询） 这个查询呢 已知实现确实没有但是我不太想加从实现中提取的特征 扩展需要多看那是没办法
而且还需要 id 刚好撞上

Immortalety:
@RPRX Since your opinion may differ from that of one of the maintainers, I would like to ask you to review this PR again. We do have a case where, although in rare but REAL cases, the bittorrent traffic detector can give a false positive, which is easy to avoid. I also wrote an explanation in this comment: https://github.com/XTLS/Xray-core/pull/6735#issuecomment-5609133872.

PR #6720, opened 2026-09-04
XHTTP transport: Add `pathTable` to disguise requests as static files

lib4u:
## Problem

XHTTP's `path` is a single fixed string, shared by every request of every user of an inbound.
Two separate consequences:

**1. The URL shape is distinctive.** With the default placements a request reads

```
GET /api/stream/1b7c6f9e-1d0a-4f0e-9f6a-6c0d2f1a55c1/17
```

A constant prefix followed by a UUID and an incrementing counter, both as path segments, is
not a shape ordinary web traffic produces.

**2. Some CDNs only forward requests that look like file requests.** Several CDNs (we hit
this on a CDNVideo-based one) answer `403` at the edge unless the last path segment contains
a file extension. Measured against such an edge, with the same origin behind it:

| Request | Result |
| --- | --- |
| `/assets/app.js?v=1&t=1` | reaches the origin |
| `/assets/app.js/?v=1&t=1` | `403` at the edge |
| `/assets/app.js/x` | `403` at the edge |
| `/assets/app.js/x.js` | reaches the origin |

A fixed `path` such as `/assets/app.js` satisfies that rule (this is what
[#6307](https://github.com/XTLS/Xray-core/pull/6307) made possible by dropping the forced
trailing slash), but then an entire deployment keeps requesting one and the same file name
forever, and changing it means reissuing every client config.

## What this adds

Three options, deliberately mirroring the existing `sessionIDTable` / `sessionIDLength`:

| Option | Type | Default | Meaning |
| --- | --- | --- | --- |
| `pathTable` | string | `""` (feature off) | Alphabet for the generated segment: a `PredefinedTable` name (`hex`, `base36`, `Base62`, …) or a literal ASCII alphabet |
| `pathLength` | Int32Range | — | Length of the generated segment, e.g. `"8-16"` |
| `pathExtension` | string | `""` | Appended verbatim after the generated segment, e.g. `".js"` |

Client config:

```json
"xhttpSettings": {
  "path": "/assets/",
  "sessionIDPlacement": "query", "sessionIDKey": "v",
  "seqPlacement": "query", "seqKey": "t",
  "pathTable": "hex",
  "pathLength": "8-16",
  "pathExtension": ".js"
}
```

Resulting requests, all from one session:

```
GET /assets/61877619ff3cadff.js?_dc=…&t=1&v=064ef1806d654a7e
GET /assets/61877619ff3cadff.js?_dc=…&t=2&v=064ef1806d654a7e
GET /assets/61877619ff3cadff.js?_dc=…&t=3&v=064ef1806d654a7e
```

Validation mirrors `sessionIDTable`: ASCII-only alphabet, `pathLength.from > 0`, and the same
"2.1B possibilities" room check. `pathExtension` is rejected without `pathTable`, or if it
contains `/`.

## Design notes

**The segment is derived from the session ID, not drawn per request.** Every request of one
session therefore asks for the same "file" — the way a real page keeps fetching one asset —
while a new session picks a new name. Drawing per request would mean a single client pulling
hundreds of distinct scripts per second, which is not what real traffic looks like, so it
would trade one anomaly for another. Derivation also keeps the change local: `sessionId` is
already a parameter of `ApplyMetaToRequest`, so no plumbing through `client.go` /
`browser_client.go` was needed. Modes without a session ID (`stream-one`) fall back to a
randomly drawn segment.

**The segment is always appended last**, after the session ID and the sequence number. It
therefore composes with every placement, including `path`, where it gives that layout a
file-like ending as well (`/assets/<session>/<seq>/<name>.js`).

**No server-side option, and no server-side change.** The server matches the configured path
as a prefix (`strings.HasPrefix` in `hub.go`) and, when the session ID and the sequence number
are not in the path, never looks at the remainder; when they are in the path, it reads them
positionally and ignores anything after them. A client using this feature therefore talks to
an **unmodified** server whose `path` is configured as the prefix (for example `"/assets/"`).

**Defaults are unchanged.** With `pathTable` empty nothing is appended and the request is
byte-for-byte what it is today.

## Testing

Unit tests added to `transport/internet/splithttp/config_test.go`:

- disabled by default;
- stable within one session, different across sessions;
- alphabet, length range and extension are all honoured;
- random when there is no session ID;
- an oversized (> 256 character) alphabet does not hang the rejection-sampling loop;
- `ApplyMetaToRequest` puts the segment last and keeps the configured path as a prefix, both
  with query placements and with path placements.

Manual testing:

- A patched client was pointed at a plain HTTP server that logs request paths: both the
  stream-down and the packet-up requests of one session used the same generated name.
- End-to-end against a **stock, unpatched** Xray v26.7.28 server (`path: "/assets/"`) with a
  real CDN in front of it: the tunnel came up and carried traffic normally, and the origin
  logged the requests quoted above. This is the compatibility claim above, verified.

`config.pb.go` was regenerated with the same toolchain the existing file records
(protoc-gen-go v1.36.11, protoc v6.33.5), so its diff is limited to the three new fields.

## Not included on purpose

- No per-request randomization (see the rationale above); if it is wanted, it is a small
  follow-up.
- No server-side knob — nothing on the server needs to know about this.
- `pathExtension` is a single fixed string rather than a list to pick from; a list would be
  easy to add later if it turns out to be useful.

### DIFF
--- a/infra/conf/transport_method.go
+++ b/infra/conf/transport_method.go
@@ -270,6 +270,9 @@ type SplitHTTPConfig struct {
 	SessionIDKey         string            `json:"sessionIDKey"`
 	SessionIDTable       string            `json:"sessionIDTable"`
 	SessionIDLength      Int32Range        `json:"sessionIDLength"`
+	PathTable            string            `json:"pathTable"`
+	PathLength           Int32Range        `json:"pathLength"`
+	PathExtension        string            `json:"pathExtension"`
 	SeqPlacement         string            `json:"seqPlacement"`
 	SeqKey               string            `json:"seqKey"`
 	UplinkDataPlacement  string            `json:"uplinkDataPlacement"`
@@ -424,6 +427,31 @@ func (c *SplitHTTPConfig) Build() (proto.Message, error) {
 		}
 	}
 
+	if c.PathTable != "" {
+		if predefined, ok := splithttp.PredefinedTable[c.PathTable]; ok {
+			c.PathTable = predefined
+		}
+		room := roomSize(len(c.PathTable), c.PathLength.From, c.PathLength.To)
+		// 2.1B possiblities should be enough
+		if room.Cmp(big.NewInt(2<<30)) < 0 {
+			return nil, errors.New("pathTable or pathLength is too small")
+		}
+		if c.PathLength.From <= 0 {
+			return nil, errors.New("pathLength.from must be greater than 0")
+		}
+		for i := 0; i < len(c.PathTable); i++ {
+			if c.PathTable[i] >= 0x80 {
+				return nil, errors.New("pathTable must contain only ASCII characters")
+			}
+		}
+	} else if c.PathExtension != "" {
+		return nil, errors.New("pathExtension requires pathTable to be set")
+	}
+
+	if strings.ContainsRune(c.PathExtension, '/') {
+		return nil, errors.New("pathExtension must not contain '/'")
+	}
+
 	if c.SeqPlacement != "path" && c.SeqKey == "" {
 		switch c.SeqPlacement {
 		case "cookie", "query":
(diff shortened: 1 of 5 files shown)


Fangliding:
反复请求同一个文件不也很诡异吗。。。
无论如何它和真实的assets都很不太一样
我已经对这种面向CDN的猫鼠游戏有点疲倦了 配置项已经太长太长了

PR #6721, opened 2026-09-05
XHTTP: Add configurable response prelude

ArthurKoba:
Some CDN configurations may wait for response body data before they start forwarding a streaming response. With XHTTP this can sometimes make the downlink stream fail to start or behave inconsistently.

This adds an optional `responsePrelude` setting. The client sends a small value with the downlink request, the server writes the same value at the beginning of the response and flushes it, and the client removes it before passing the actual stream data further.

The intention is just to make the response start earlier for CDN setups where headers alone are not enough.

It is optional and does not change the normal XHTTP behavior when it is not configured. A client without `responsePrelude` continues to use the existing path.

I tested the main request placements, packet-up/downlink behavior, compatibility without the option, and added an end-to-end test with a buffering HTTP intermediary. It was also tested with a real CDN setup where this helped the stream start.

The current default names are `sample_id` for cookie/query and `X-Sample-ID` for header. I tried to keep them generic, but I can change the naming if something else fits XHTTP better.

### DIFF
--- a/infra/conf/transport_method.go
+++ b/infra/conf/transport_method.go
@@ -255,36 +255,44 @@ func (c *TCPConfig) Build() (proto.Message, error) {
 }
 
 type SplitHTTPConfig struct {
-	Host                 string            `json:"host"`
-	Path                 string            `json:"path"`
-	Mode                 string            `json:"mode"`
-	Headers              map[string]string `json:"headers"`
-	XPaddingBytes        Int32Range        `json:"xPaddingBytes"`
-	XPaddingObfsMode     bool              `json:"xPaddingObfsMode"`
-	XPaddingKey          string            `json:"xPaddingKey"`
-	XPaddingHeader       string            `json:"xPaddingHeader"`
-	XPaddingPlacement    string            `json:"xPaddingPlacement"`
-	XPaddingMethod       string            `json:"xPaddingMethod"`
-	UplinkHTTPMethod     string            `json:"uplinkHTTPMethod"`
-	SessionIDPlacement   string            `json:"sessionIDPlacement"`
-	SessionIDKey         string            `json:"sessionIDKey"`
-	SessionIDTable       string            `json:"sessionIDTable"`
-	SessionIDLength      Int32Range        `json:"sessionIDLength"`
-	SeqPlacement         string            `json:"seqPlacement"`
-	SeqKey               string            `json:"seqKey"`
-	UplinkDataPlacement  string            `json:"uplinkDataPlacement"`
-	UplinkDataKey        string            `json:"uplinkDataKey"`
-	UplinkChunkSize      Int32Range        `json:"uplinkChunkSize"`
-	NoGRPCHeader         bool              `json:"noGRPCHeader"`
-	NoSSEHeader          bool              `json:"noSSEHeader"`
-	ScMaxEachPostBytes   Int32Range        `json:"scMaxEachPostBytes"`
-	ScMinPostsIntervalMs Int32Range        `json:"scMinPostsIntervalMs"`
-	ScMaxBufferedPosts   int64             `json:"scMaxBufferedPosts"`
-	ScStreamUpServerSecs Int32Range        `json:"scStreamUpServerSecs"`
-	ServerMaxHeaderBytes int32             `json:"serverMaxHeaderBytes"`
-	Xmux                 XmuxConfig        `json:"xmux"`
-	DownloadSettings     *StreamConfig     `json:"downloadSettings"`
-	Extra                json.RawMessage   `json:"extra"`
+	Host                 string                 `json:"host"`
+	Path                 string                 `json:"path"`
+	Mode                 string                 `json:"mode"`
+	Headers              map[string]string      `json:"headers"`
+	XPaddingBytes        Int32Range             `json:"xPaddingBytes"`
+	XPaddingObfsMode     bool                   `json:"xPaddingObfsMode"`
+	XPaddingKey          string                 `json:"xPaddingKey"`
+	XPaddingHeader       string                 `json:"xPaddingHeader"`
+	XPaddingPlacement    string                 `json:"xPaddingPlacement"`
+	XPaddingMethod       string                 `json:"xPaddingMethod"`
+	UplinkHTTPMethod     string                 `json:"uplinkHTTPMethod"`
+	SessionIDPlacement   string                 `json:"sessionIDPlacement"`
+	SessionIDKey         string                 `json:"sessionIDKey"`
+	SessionIDTable       string                 `json:"sessionIDTable"`
+	SessionIDLength      Int32Range             `json:"sessionIDLength"`
+	ResponsePrelude      *ResponsePreludeConfig `json:"responsePrelude"`
+	SeqPlacement         string                 `json:"seqPlacement"`
+	SeqKey               string                 `json:"seqKey"`
+	UplinkDataPlacement  string                 `json:"uplinkDataPlacement"`
+	UplinkDataKey        string                 `json:"uplinkDataKey"`
+	UplinkChunkSize      Int32Range             `json:"uplinkChunkSize"`
+	NoGRPCHeader         bool                   `json:"noGRPCHeader"`
+	NoSSEHeader          bool                   `json:"noSSEHeader"`
+	ScMaxEachPostBytes   Int32Range             `json:"scMaxEachPostBytes"`
+	ScMinPostsIntervalMs Int32Range             `json:"scMinPostsIntervalMs"`
+	ScMaxBufferedPosts   int64                  `json:"scMaxBufferedPosts"`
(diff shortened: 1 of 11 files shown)


Fangliding:
没听说过这样奇怪的CDN 改动太大了
如果真有这样的行为是那样为什么不使用 packet-up 呢 下行是GET 这样总没有请求体了

ArthurKoba:
> 没听说过这样奇怪的CDN 改动太大了 如果真有这样的行为是那样为什么不使用 packet-up 呢 下行是GET 这样总没有请求体了

@Fangliding, thanks for taking the time to look at this so quickly and for the review. I think there may be a misunderstanding about which side of the HTTP exchange is causing the issue, so I’ll try to explain the exact behavior more clearly below.

The issue is not related to the request body, and `packet-up` does not avoid it.

This already happens in `packet-up` mode:

1. The uplink is split into separate requests.
2. The downlink is still a long-lived GET request with no request body.
3. Xray accepts that GET and starts the response normally.
4. The origin sends the HTTP response headers and flushes them.
5. Some intermediaries/CDNs still do not start forwarding that streaming response to the client until they see actual response body bytes from the origin.

So the problematic part is:

client -> GET downlink -> CDN -> Xray
Xray -> response headers -> CDN
CDN waits
Xray has not produced actual downlink payload yet
client therefore does not receive a usable streaming response

This is especially visible when the downlink stream is opened before there is real application data to send. From Xray's point of view the response is already started, but the intermediary may still buffer it or delay exposing it downstream until the response body becomes non-empty.

`responsePrelude` only addresses that initial response-start condition.

The client generates a short random value and places it in the downlink GET metadata, for example in a cookie, header, query parameter or path.

Example conceptually:

GET /path
Cookie: sample_id=abc123...

The server reads that value only for the stream-down request. Before any proxied payload is written, it writes exactly the same value as the first bytes of the HTTP response body and flushes the response.

So the response becomes conceptually:

HTTP/1.1 200 OK
Transfer-Encoding: chunked

abc123...<normal XHTTP payload continues here>

Once the client receives the response, it expects the same value it originally sent. It reads only that prefix from the response body and removes it before exposing the stream to the rest of Xray.

Therefore upper layers still see exactly the original XHTTP payload:

wire:
[prefix][real XHTTP data]

after client-side stripping:
[real XHTTP data]

There is no persistent framing format, no per-packet framing, and no keepalive protocol added to the downlink. The prelude exists only once at the very beginning of the response. After it is stripped, the rest of the response is the normal raw XHTTP stream.

If `responsePrelude` is not configured, nothing changes and the existing XHTTP behavior is used.

It also does not affect packet-up requests carrying uplink chunks. The prelude is only applied to the dedicated downlink GET, so packet-up upload traffic is unchanged.

The placement/key/table/length options are only there so this metadata does not have to be hard-coded to one HTTP shape. Different intermediaries may treat cookies, headers, query parameters and paths differently, and deployments may already have their own naming or size constraints.

The actual runtime mechanism is small:
- client creates one short value for the downlink request;
- server echoes it once as the first response-body bytes and flushes;
- client removes the same bytes;
- normal XHTTP stream continues unchanged.

I reproduced this through a buffering HTTP intermediary in an end-to-end test, and also with a real CDN setup where response headers alone were not enough to make the downlink start, while sending the first body bytes made the response begin forwarding.

ArthurKoba:
This change is necessary because it addresses a real failure mode rather than only changing request appearance.

The current implementation has been tested to solve the issue without changing the normal XHTTP stream after startup. If more evidence would be useful, I can provide details from the specific CDN setup, packet captures showing the behavior before and after the prelude, and a concrete reproduction demonstrating how the current patch resolves it.

I’m also happy to adjust the implementation or configuration surface if you think there is a better way to fit this into XHTTP.

Fangliding:
> 5\. Some intermediaries/CDNs still do not start forwarding that streaming response to the client until they see actual response body bytes from the origin

源都没发数据出来何来流式转发一说？这不是正常的么

ArthurKoba:
> > 5. Some intermediaries/CDNs still do not start forwarding that streaming response to the client until they see actual response body bytes from the origin
> 
> 源都没发数据出来何来流式转发一说？这不是正常的么

@Fangliding I probably explained the failure point too broadly before.

The problematic CDN accepts the downlink GET and forwards it to Xray normally. Xray then returns `200` and the response headers, but the CDN does not immediately pass that long-lived response through to the client. It keeps the response pending until it sees the first bytes of the response body from the origin.

The problem is that in XHTTP there may be no real downlink payload available at that moment. So the response can stay stuck in this intermediate state: Xray considers the downlink response started, but the client has not actually received a usable response stream yet. As a result, the downlink side of the XHTTP session does not become established properly, and the tunnel cannot continue normally even though the GET request itself was accepted.

Without the prelude the lifecycle is roughly:

downlink GET
→ CDN forwards request to Xray
→ Xray sends `200` + headers
→ CDN waits for first response-body bytes
→ no real downlink payload exists yet
→ response remains pending at the CDN
→ client never gets the usable downlink stream
→ XHTTP session cannot proceed normally

With `responsePrelude`:

downlink GET
→ CDN forwards request to Xray
→ Xray sends `200` + headers
→ Xray immediately writes a short prelude and flushes it
→ CDN sees actual response-body bytes and starts forwarding the response
→ client receives the response, removes the prelude
→ normal raw XHTTP downlink continues unchanged

So the prelude is only used once to get the response through this initial CDN buffering/start condition. It is not a keepalive mechanism and it does not introduce framing for the rest of the stream.

Fangliding:
发出上行数据不用等服务端的200ok啊 客户端可以先发一些请求报文等真正的下行字节回来不就是了

ArthurKoba:
> 发出上行数据不用等服务端的200ok啊 客户端可以先发一些请求报文等真正的下行字节回来不就是了

@Fangliding Yes, I understand what you mean. However, in this case the problem seems to be in the CDN response handling itself.

I am testing this from Russia, where some of the larger CDN services that used to work well are becoming less accessible for ordinary users, so in practice we sometimes have to rely on smaller or less common CDN providers. Those providers can have different or imperfect buffering behavior, and this appears to be one of those cases.

XHTTP already does the expected things on the server side. In the stream-down branch it sets `X-Accel-Buffering: no`, `Cache-Control: no-store`, optionally uses `text/event-stream`, then sends `200` and calls `Flush()`.

After that, no response body is written until real downlink data becomes available. When that happens, `httpServerConn.Write()` writes the payload and flushes again.

With this CDN, one of two things seems to be happening, or possibly both:

- the CDN ignores or does not fully honor the response headers and initial `Flush()`, so the response remains buffered;
- there is a response-start timing requirement, and if body bytes are not produced soon enough after `200`, the response is moved into a different buffering/timeout state.

In either case, later XHTTP downlink data is no longer enough to make the response start correctly, so the return path remains unusable even though packet-up requests can still reach the server.

The prelude changes only this timing point. It writes a small body immediately after `WriteHeader(http.StatusOK)` and flushes it. That gives the CDN actual response-body data at the moment the response is opened. After that, the same CDN forwards the response normally, and all later downlink data goes through the existing `httpServerConn.Write()` path without any further special handling.

Before adding this patch, I spent some time trying the settings that seemed relevant to this specific failure mode on the nginx side, in the CDN configuration, and in Xray itself. I tested the options that could reasonably affect response buffering, streaming, caching and timeout behavior, but none of them changed the result.

If needed, I can provide the exact CDN provider and setup I am using, its relevant configuration, the corresponding Xray config, the nginx config, and packet captures from both the failing case and the working case with the prelude enabled.

Fangliding:
> * there is a response-start timing requirement, and if body bytes are not produced soon enough after `200`, the response is moved into a different buffering/timeout state.

太罕见了 不考虑

ArthurKoba:
> * there is a response-start timing requirement, and if body bytes are not produced soon enough after `200`, the response is moved into a different buffering/timeout state.

@Fangliding I understand that this is a rare case.

At the same time, the patch itself is relatively small in terms of runtime behavior. If `responsePrelude` is not configured, it does not change the existing XHTTP path at all. It only adds an optional mechanism for specific cases where the response otherwise does not start correctly through the intermediary.

For users in Russia, the number of available and working CDN services is shrinking rapidly. Most large providers already require a legal entity or business account. Registering a legal entity itself is not an insurmountable problem, but using such an account for proxying creates a completely different level of risk: large fines, criminal liability, possible imprisonment, and other serious consequences. At the same time, the government is actively developing methods to analyze and block VPN traffic, which makes large CDN providers effectively unusable for ordinary users who need access to the free internet.

So I would not consider this case so rare that it can simply be ignored. I have already observed similar behavior with at least two mid-sized CDN providers, Timeweb and Beget. As the number of accessible large providers keeps decreasing, compatibility with the remaining options becomes significantly more important.

What is the main issue you see with this patch? If the problem is the size of the change, the configuration surface, or insufficient evidence that this behavior is real, I can adjust the implementation, simplify it, or provide more detailed reproduction data and packet captures.

PR #6735, opened 2026-09-08
sniffer: fix a uTP false positive and detect DHT and UDP tracker traffic

TripleA150:
## Summary

Two things:

1. **Fixes a false positive** in `SniffUTP`, where an ordinary DNS query is classified as BitTorrent.
2. **Extends detection** to the two UDP protocols a BitTorrent client actually speaks first and which were not recognised at all: the DHT (BEP 5) and UDP trackers (BEP 15).

`Protocol()` still returns exactly `"bittorrent"`, `app/dispatcher/default.go` is untouched, and the existing sniffer order is unchanged.

## The false positive

`SniffUTP` accepts a standard DNS query as a uTP `ST_SYN`. Every field lines up:

| Bytes | uTP field | What a DNS query has there | Result |
|---|---|---|---|
| `b[0]` | `0x41` — ST_SYN, version 1 | high byte of the transaction id | passes when the id starts with `0x41` |
| `b[1]` | extension list head | low byte of the transaction id | `0x00` means "no extensions", so the chain walk is skipped |
| `b[8:12]` | `timestamp_difference` | `nscount` + `arcount` | **always zero for a query without EDNS0** |
| `len == 20` | a bare header with no payload | 12 byte header + 8 byte question | a two character name fits exactly |

The existing `"dns query"` case in `bittorrent_test.go` aims at this collision but uses a 22 byte packet, which the "extensions consume all payload" check rejects. Shortening the queried name to two characters produces exactly 20 bytes and slips through.

### Fix

Require `ack_nr` to be zero. In a DNS query that offset holds `qclass`, which is `0x0001` for the internet class, so the whole collision class disappears. The check adds 16 bits of strictness and only narrows the set of accepted packets, so it cannot introduce new false positives.

BEP 29 describes `ack_nr` as meaningless in `ST_SYN`, so this was checked against implementations rather than the specification:

- **libutp** — `utp_connect` does `memset(p1, 0, header_size)` and then assigns `version`, `type`, `ext`, `connid`, `windowsize` and `seq_nr`. `ack_nr` is never assigned and stays zero.
- **libtorrent** — `send_syn()` sets `m_ack_nr = 0`, and `send_pkt` writes `h->ack_nr = m_ack_nr`.

### UDP tracker, BEP 15

The connect request is the only packet a client sends first, and it is exactly 16 bytes: the magic `0x41727101980`, `action == 0`, and a transaction id. 64 bits of constant leave no room for accidental matches.

It has to be a separate function rather than a branch inside `SniffUTP`, because 16 bytes is below that function's `len(b) < 20` floor and would be discarded as `ErrNoClue` before any check runs.

### DHT, BEP 5

KRPC messages are bencoded dictionaries, one per datagram. Rather than scanning for substrings, this validates the structure with a minimal bencode reader. A datagram is accepted only when:

- it parses as a bencoded dictionary that **ends exactly on the last byte** of the datagram — the same trick the extension chain walk already uses;
- it carries `y` with a value of `q`, `r` or `e`, and a transaction id `t`;
- a query also carries a method name and arguments holding a **20 byte** node id; a response carries the same id; an error carries a list.

Parsing is bounded at depth 8, so nested lists such as `llllll…` cannot exhaust the stack. There is a test for that.

Only the parts of bencode needed to validate a message are implemented — a sniffer needs the framing, not the contents.

## False positive testing

Three million random datagrams of 16 to 215 bytes were run through all three UDP sniffers:

```
rounds=3000000  utp=0  tracker=0  dht=0
```

That check was scratch work and is not part of the diff. What is committed are explicit negative cases for the protocols that share UDP with BitTorrent — DNS (including the 20 byte collision above), STUN binding requests, WireGuard handshake initiations, uTP SYN packets and tracker connect requests — each cross-checked against every sniffer.

`SniffBittorrent` had no tests at all; it now has a table covering the handshake, truncation, a wrong `pstrlen`, a wrong protocol string, a TLS ClientHello and an HTTP tracker announce.

## Behaviour change

**DHT and UDP tracker traffic now matches `protocol: bittorrent` routing rules.** Traffic that previously fell through will now be matched — users with a `bittorrent` block rule will start blocking it. That is the point of the change, but it is worth knowing before upgrading.

HTTP tracker announces are still classified as `http`, since `SniffHTTP` runs first and wins on them. Changing that would alter existing behaviour, so it is out of scope here.

### DIFF
--- a/app/dispatcher/sniffer.go
+++ b/app/dispatcher/sniffer.go
@@ -41,6 +41,8 @@ func NewSniffer(ctx context.Context) *Sniffer {
 			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffBittorrent(b) }, false, net.Network_TCP},
 			{func(c context.Context, b []byte) (SniffResult, error) { return quic.SniffQUIC(b) }, false, net.Network_UDP},
 			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffUTP(b) }, false, net.Network_UDP},
+			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffUDPTracker(b) }, false, net.Network_UDP},
+			{func(c context.Context, b []byte) (SniffResult, error) { return bittorrent.SniffDHT(b) }, false, net.Network_UDP},
 		},
 	}
 	if sniffer, err := newFakeDNSSniffer(ctx); err == nil {
--- a/common/protocol/bittorrent/bittorrent.go
+++ b/common/protocol/bittorrent/bittorrent.go
@@ -46,6 +46,12 @@ func SniffUTP(b []byte) (*SniffHeader, error) {
 		return nil, errNotBittorrent
 	}
 
+	// ack_nr carries no meaning until the peer replies, and implementations send
+	// a zeroed one. Requiring it rules out DNS queries, whose qclass sits here
+	if binary.BigEndian.Uint16(b[18:20]) != 0 {
+		return nil, errNotBittorrent
+	}
+
 	// Walk the extension chain. Selective ack (1) and extension bits (2)
 	extension, offset := b[1], 20
 	for extension != 0 {
@@ -74,6 +80,33 @@ func SniffUTP(b []byte) (*SniffHeader, error) {
 
 	// extensions should consume all ST_SYN payload
 	if len(b) != offset {
+		return nil, errNotBittorrent
+	}
+
+	return &SniffHeader{}, nil
+}
+
+// udpTrackerMagic opens every BEP 15 connect request.
+const udpTrackerMagic = 0x41727101980
+
+// SniffUDPTracker matches the connect request that opens a session with a UDP
+// tracker (BEP 15). It is the only packet a client sends first, and it is
+// exactly 16 bytes: the magic, action 0, and a transaction id.
+func SniffUDPTracker(b []byte) (*SniffHeader, error) {
+	if len(b) < 16 {
+		return nil, common.ErrNoClue
+	}
+
+	if len(b) != 16 {
+		return nil, errNotBittorrent
+	}
+
+	if binary.BigEndian.Uint64(b[0:8]) != udpTrackerMagic {
+		return nil, errNotBittorrent
+	}
+
+	// action 0 is connect
+	if binary.BigEndian.Uint32(b[8:12]) != 0 {
 		return nil, errNotBittorrent
 	}
 
--- a/common/protocol/bittorrent/bittorrent_test.go
+++ b/common/protocol/bittorrent/bittorrent_test.go
@@ -42,6 +42,12 @@ func TestSniffUTP(t *testing.T) {
 			0x41, 0x00, 0x01, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 			0x01, 'a', 0x02, 'c', 'o', 0x00, 0x00, 0x01, 0x00, 0x01,
 		}, errNotBittorrent},
+		// the same collision at exactly 20 bytes, where the extension chain check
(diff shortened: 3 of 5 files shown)


RPRX:
@Fangliding 这个 PR 得尽快看下，有点 release-blocker 了，~~别常规 DNS 被识别成 BT 然后俄罗斯机场用户账号被封了~~

Fangliding:
https://github.com/XTLS/Xray-core/pull/6715#issuecomment-5520402051
没必要 和这个重复了 我看过别的软件的嗅探好像一样有这个问题 也没出事

TripleA150:
This commit addresses not only an erroneous SniffUTP but also improvements to BitTorrent recognition

Do you need another PR to review it?

@Fangliding @RPRX

Fangliding:
https://github.com/search?q=repo%3AXTLS%2FXray-core+UTP&type=pullrequests

TripleA150:
As the author of one of the commits rightly noted, there are solutions such as the Xray Torrent Blocker that block users when BitTorrent protocol activity is detected.

However, blocking requires detection first. While no one aims to fully identify and block all BitTorrent traffic which would be futile. I do believe it is necessary to improve detection capabilities to support external solutions.

As you suggested, blocking tracker addresses and high UDP ports does not yield the desired results, this has already been proven in practice.

https://github.com/XTLS/Xray-core/pull/6668#issuecomment-5385977900

We would appreciate it if you reopen the two PRs regarding improved detection.

@Fangliding

RPRX:
@Fangliding 看看这个有没有必要搞一下，~~不然大规模出现我说的那种情况就搞笑了~~

Fangliding:
1 不会有人去莫名其妙查 AB YR 这样二字域名
2 哪怕发了 也需要query id以 1/65535 的概率撞上
不会有的

Fangliding:
singbox的嗅探也会误识别这种 也没看到complain

TripleA150:
You're right that the probability is low. A few things I'd add though:

1. The repo already has a test for this exact scenario. The comment at
   bittorrent_test.go:40 reads "txid 0x4100, no EDNS0: the worst case
   colliding with the uTP header" whoever wrote that test already
   considered it worth guarding against. The problem is that the sample
   there is 22 bytes, so it gets rejected by the "extensions must consume
   the whole packet" check rather than by anything DNS-specific. Shorten
   the queried name to two characters and you land on exactly 20 bytes,
   which passes. This PR just makes that existing test actually hold.

2. The 1/65536 figure assumes a randomised txid. A client using an
   incrementing counter, common in simple embedded stub resolvers -
   passes through 0x4100 once per cycle deterministically. That half of
   the condition isn't a lottery, leaving only "the query happens to be a
   two-character single label".

3. The condition isn't a domain like ab.com, it's a two-character
   single-label QNAME with any qtype. Short intranet hostnames, queries
   where the search domain wasn't expanded, and apex lookups of
   two-letter TLDs all produce packets of that shape.

4. On sing-box having the same behaviour with no complaints: the failure
   mode is a single DNS query silently routed as BitTorrent, once in tens
   of millions. A user can't attribute an occasional resolution failure to
   the sniffer, so "no complaints" is what you'd observe whether or not
   the problem is there. It's evidence about detectability, not about
   correctness.

The cost is 2 lines and one uint16 comparison, with no false negatives:
libutp's utp_connect never assigns ack_nr after the memset, and
libtorrent's send_syn() sets m_ack_nr = 0 explicitly.

If you still don't think it's worth it, this part can be dropped on its
own without affecting the substance of the PR - the DHT (BEP 5) and UDP
tracker (BEP 15) detection is independent of it.

Immortalety:
As the author of the previous pull request that fixed the false‑positive detection, I’d like to note that you don’t want to rely on existing implementations of uTP, yet you’re relying on assumptions about the implementation of DNS clients and their operation in real‑world conditions. It seems to me that, of two evils, we should choose the lesser one — the one that potentially causes fewer problems for end users.

PR #6743, opened 2026-09-10
common/buf/readv_windows.go: Fix raw WSARecv blocking thread and preventing connection close

kiangz:
## Problem

On Windows, an xray client (socks inbound + any outbound over a raw TCP transport, e.g. VLESS/WS) stops closing the downstream (socks) connection when the upstream closes after the client has uploaded more than ~16 KB. The downstream client receives the response but never sees EOF and hangs until its own timeout.

Reproduction (`curl -> socks5(xray) -> VLESS/WS -> server that closes after reading the request`):

- POST body <= ~16 KB: connection closes normally.
- POST body > ~16 KB: the socks connection stays ESTABLISHED forever after the WS server sends Close + FIN.

The server implementation is irrelevant: the same hang reproduces with a real xray server as the upstream, while sing-box as the client works fine against the same server.

## Root cause

Go runtime network sockets on Windows are kept in **blocking mode at the Winsock level**; the runtime simulates non-blocking I/O with overlapped operations. `ReadVReader` (engaged after the first full 8 KB read) calls `RawConn.Read` with a callback that issues a **raw synchronous `WSARecv` (nil OVERLAPPED)**. When no data is available, this call blocks its OS thread and nothing can cancel it:

- while blocked it holds the `internal/poll.FD` reference, so `TCPConn.Close()` cannot reach refcount 0: the handle is never closed and **no FIN is ever sent** to the downstream peer (`FD.Close` blocks in `runtime_Semacquire(&fd.csema)` waiting for the last reference);
- the blocked `WSARecv` only returns when the *peer* closes the connection (EOF) or data arrives.

Whether the hang occurs depends on whether the pending upstream read sits in the raw readv path: with <= 16384 bytes uploaded, the readv alloc strategy shrinks back to 1 buffer and subsequent reads go through the interruptible plain-read path, which is exactly why small uploads work and the observed threshold is ~16 KB (16384 = 2 x 8192 `buf.Size`).

Instrumented logs confirming the mechanism (before the fix): the worker logs `closing conn`, `conn.Close()` is entered, yet the downstream peer sees no FIN for 10+ seconds, and the pending read only returns EOF when the downstream peer itself closes.

## Fix

Make `windowsReader.Read` honor the `RawConn.Read` contract ("must not block"): the first call of each multi-buffer read returns `-1` ("not ready"), so the runtime waits for readability with its own interruptible overlapped zero-byte read (cancelled via `CancelIoEx` on close); the real multi-buffer `WSARecv` is only issued once data (or EOF) is available and therefore always returns immediately. Readv batching is preserved. Linux/Unix are unaffected (fds are truly non-blocking there, `readv` returns EAGAIN and integrates with the netpoller).

After the fix, `Close()` unblocks the pending read (`use of closed network connection`), the handle is closed, and the FIN reaches the downstream client immediately.

## Testing

New scenario regression test `testing/scenarios/ws_upload_close_test.go` (socks -> vless/ws client, plain WS server that closes right after receiving the full upload), covering:

- small upload below the readv threshold (8000 B),
- big upload on the raw readv path (20000 B) - hung before the fix,
- server responding before closing (20000 B + response) - the reported production symptom: the downstream client must receive the full response followed by EOF.

All three pass on Windows with the fix; without it the big-upload tests fail with `i/o timeout` on the downstream read. `go build ./...`, `go test ./common/buf/` and `go test ./common/task/` pass.


Fangliding (approved):

RPRX:
问题有点大但看着有点玄学，仍需第三方独立验证问题存在，~~希望不是 AI 脑补的产物~~，既然 @Fangliding approve 了就先合了

~~话说我的 Tor 浏览器都是接 Xray Socks5，后者协议重连后前者恢复较慢（大多数时候有点 broken，恢复时间超过了单纯连接 Tor 节点所需的时间，有时甚至得重开 Xray 乃至 Tor 浏览器）跟这个有关？前者能通过发包来感知到存在错误、开新连接吗？~~

Fangliding:
这个场景只是极度少见 需要连接xray的程序发一个大上行 后续一个比特不发 然后远端close导致xray尝试close这个本地连接 但是判定底层fd还在read状态导致锁死（这是个不可打断的阻塞IO） 但凡连接xray的程序再发一个比特的payload来或者主动关闭连接都可以解除这个状态
但是有关正确性改了就改了吧

RPRX:
> 但凡连接xray的程序再发一个比特的payload来

但是如果那个程序就是处在发了包等回复的状态就正好 GG 了，不过握手第一个包一般不超过 16KB

所以多次读写后进入这状态也受影响吧？h2 协议应该没事（Tor 虽然套了层但应该也没事），~~如果是单纯下载的协议就废了~~

~~一直在想一些中途 ping-pong 的场景，比如如果 tor 里面不是 h2 然后外层协议 idle 保活？不过那也有 timeout，有点烧脑了~~

~~比如说它会不会是 ping 没收到 pong 就默认后继线路没了但既然与守卫节点间的连接还在就还去 call 它，不过据称能解除状态~~

Fangliding:
首先它需要一个相当大的上行行为 少量上行会被xray快速从缓冲区搬走根本不会进readv模式 发完这个大包之后上行链路必须立刻静默
其次remote需要卡这个timing直接关闭连接 不然摸不到这个位置
各种来来回回的小包都会解除锁定 比如h2会产生大量的window update啥的 或者其他心跳行为
所以我说很难遇到

PR #6744, opened 2026-09-10
Burst observatory: add option to limit number of outbounds during init check

arrgti:
BurstObservatory pings all matched outbounds simultaneously during the initial check. With small number of outbounds that's ok, but if we have a lot of them, a large burst of connections at startup may look unusual. Also this create a high load and affect the initial check results.

This adds the optional startupChecks field to pingConfig, which allows limiting the number of outbounds checked during the initial health check. The selected outbounds are picked randomly:

```json
"pingConfig": {
  "interval": "60m",
  "sampling": 4,
  "startupChecks": 8
}
```

startupChecks = 0 (default) - unchanged behavior (all matched outbounds are checked at startup as before)
startupChecks < 0 - unchanged behavior
startupChecks >= len(matched outbounds) - unchanged behavior
startupChecks: N (N > 0, N < number of matched outbounds) - only N randomly chosen outbounds are checked at startup

### DIFF
--- a/app/observatory/burst/config.pb.go
+++ b/app/observatory/burst/config.pb.go
@@ -88,7 +88,9 @@ type HealthPingConfig struct {
 	// ping timeout, int64 values of time.Duration
 	Timeout int64 `protobuf:"varint,5,opt,name=timeout,proto3" json:"timeout,omitempty"`
 	// http method to make request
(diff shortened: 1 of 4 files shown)


Fangliding:
不打算再对obs进行非bug修复的改动了(

PR #6746, opened 2026-09-10
Fix outbound tags cache data race

svlavr:
## Summary

- clear the existing outbound tags cache when handlers change instead of
  replacing its pointer
- synchronize the stop flags in "TestTagsCache"

## Problem

"Manager.Select" reads "m.tagsCache" on its cache-hit path without acquiring
"Manager.access". "AddHandler" and "RemoveHandler" replaced that pointer while
holding "access", but that lock did not synchronize with the unlocked read.

This is reproducible with:

    go test -race ./app/proxyman/outbound -run '^TestTagsCache$' -count=1

The test also contained two independent races in its stop flags, which obscured
the production regression result.

Dynamic handler mutation is used by the handler service and runtime features.
A selection overlapping removal may transiently retain an old candidate. If
that tag is no longer registered when dispatch resolves it, the new link is
closed. Existing flows that already hold a handler are unaffected.

## Fix

Keep the initialized "sync.Map" and call "Clear" under the existing manager
write lock.

This removes the unsafe pointer publication while preserving cache
invalidation and the cache-hit fast path. The module targets Go 1.27;
"sync.Map.Clear" has been available since Go 1.23.

The test stop flags now use "atomic.Bool", without reducing the concurrent
"AddHandler", "RemoveHandler" and "Select" workload.

## Scope

This fix removes unsafe "tagsCache" pointer publication and preserves the
cache-hit fast path. It does not change balancer or selector policy and does
not make handler selection linearizable with concurrent handler mutation.

Unrelated observatory scheduling behavior is not claimed fixed.

## Validation

- "go test ./app/proxyman/outbound -run '^TestTagsCache$' -count=3"
- "go test -race ./app/proxyman/outbound -run '^TestTagsCache$' -count=3"
- "go test -race ./app/proxyman/outbound -count=1"
- relevant router and observatory tests
- "go vet" for affected packages
- "git diff --check"

### DIFF
--- a/app/proxyman/outbound/handler_test.go
+++ b/app/proxyman/outbound/handler_test.go
@@ -107,11 +107,11 @@ func TestTagsCache(t *testing.T) {
 	v.AddFeature(ohm)
 	ctx := context.WithValue(context.Background(), xrayKey, v)
 
-	stop_add_rm := false
+	stop_add_rm := atomic.Bool{}
 	wg_add_rm := sync.WaitGroup{}
 	addHandlers := func() {
 		defer wg_add_rm.Done()
-		for !stop_add_rm {
+		for !stop_add_rm.Load() {
 			time.Sleep(delay)
 			idx := counter.Add(1)
 			tag := fmt.Sprintf("%s%d", tags_prefix, idx)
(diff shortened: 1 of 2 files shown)


Fangliding:
顶多去取到几十微秒前的稍旧的cache而已 本来goroutine的执行顺序就有浮动

PR #6747, opened 2026-09-10
TUN inbound: Preserve UDP packet destinations with traffic stats

svlavr:
## Summary

#5888 made a TUN UDP association preserve per-packet destinations through `ReadMultiBuffer`, `WriteMultiBuffer`, and `Buffer.UDP`.

When inbound traffic statistics are enabled, the TUN connection is currently wrapped in `stat.CounterConnection`. That wrapper exposes byte-oriented `net.Conn` methods and hides the native buffer reader/writer interfaces. On a multidestination UDP association, subsequent packets can therefore follow the first destination.

This change keeps the native buffer path for TUN UDP and performs its inbound accounting at buffer level. TCP keeps the existing `CounterConnection` path.

## Functional reproducer

The regression test uses the same production TUN UDP association for both packets:

- packet #1 has destination A;
- packet #2 has destination B;
- with inbound statistics disabled on the baseline, A receives #1 and B receives #2;
- with inbound statistics enabled on the baseline, packet #2 incorrectly follows A;
- with this change, A/B delivery is preserved with statistics enabled;
- replies preserve their per-packet source destination;
- uplink and downlink counters remain correct.

The test path is:

`udpConnectionHandler.HandlePacket`
→ production `udpConn`
→ `Handler.HandleConnection`
→ `DefaultDispatcher`
→ stock Freedom
→ real localhost UDP sockets.

The regression requires enabled inbound traffic statistics and a multidestination UDP association; this does not claim that all TUN UDP traffic is broken.

## Scope

- TCP behavior is unchanged.
- No router, dispatcher, or Freedom code is changed.
- No mutex or framework is added.
- The production change is local to TUN.
- The per-packet destination behavior from #5888 is preserved rather than redesigned.

## Validation

- `go test ./proxy/tun -count=1`
- `go test -race ./proxy/tun -count=1`
- focused regression tests with `-count=20`
- `go test ./app/dispatcher -count=1`
- `go vet ./proxy/tun`
- `git diff --check`


RPRX:
TUN inbound 开 traffic stats 有点“统计整机流量消耗”的意思了，~~现在这么干的应该不多，先合了看看，$288 会搞“分目标”统计~~

RPRX:
~~看了下代码，core 里一整个流量统计都得重构为对“单连接”的统计（buf.Copy），附带对“单连接”的控制 API https://github.com/XTLS/Xray-core/pull/5844 @Meo597~~

Fangliding:
一半的时间浪费在和conn接口大战buf.Writer接口

RPRX:
~~v2ray 的历史遗产，不过不搞 multi-buffer 的话也搞不了 readv 这些，不过若把 buf.Size 增大到 16384 可能 readv 就不明显了？~~

RPRX:
~~类似 v2 的 pb 遗产~~，你测一下吧，如果能全换成 conn 接口也方便搞空间预留、减少 copy，~~至于 pipe，有些地方还是得用到~~

~~反正 v2 的过度设计除了难清的基本上都清完了，剩下的绝对都是难清的~~

Fangliding:
read/writev是本地跑回环斗蛐蛐核心科技 虽然之前好像被搞坏了

16384的buffer size估计问题不大 我在自己的fork上跑了相当一段时间（因为tls最大也是这么大一块） 似乎没发现什么问题 不过也就是把触发这个问题需要的payload从16kb变成32kb而已

RPRX:
这个 size 更贴合浏览器（不开 vision 的情况下），变化主要影响 XHTTP，比如说现在它一个 TLS record 可能 8000+，就很假

~~但是把这个 size 直接塞进 XHTTP 的话指不定 Golang 的 h2 库会给搞出 16K 大包加一个几百 K 小包的情况，也很假，需要测试~~

svlavr:
Thanks for merging!

My reasoning was pretty simple: enabling traffic stats shouldn't change where TUN UDP packets are routed. It should ~~preferably~~ always preserve each packet's destination. :)

The per-connection statistics/control API sounds interesting too, though that seems like a broader follow-up.

svlavr:
To avoid confusion: this work is currently in a private downstream fork. The public Xray-core fork visible on my GitHub profile is a separate repository and does not contain this architecture.

After #6747 I followed this topic much further there, mainly around traffic observation and byte accounting.

The more data paths I examined, the more it looked like the difficult part was not incrementing a counter. The difficult part was choosing the boundary where the counter lives and defining exactly what its value means.

#6747 is a good small example. `stat.CounterConnection` counted bytes correctly at its own `net.Conn` boundary, but for TUN UDP it hid the native `buf.Reader` / `buf.Writer` interfaces. That also removed the per-packet `Buffer.UDP` destination metadata, so merely enabling statistics changed datapath behavior.

In other words, the arithmetic was correct, but the accounting boundary was wrong for that traffic shape.

After that I started separating byte scopes explicitly. I worked through ordinary TCP flows, UDP associations, decoded MUX children, MUX carrier traffic, and retained XUDP link epochs. It became clear that all of the following may be valid counters while describing different things:

- inbound `net.Conn` I/O;
- bytes accepted by a logical `transport.Link`;
- UDP-association payload;
- MUX/XUDP logical-flow traffic;
- encoded MUX carrier I/O;
- outbound connection I/O.

These values are not interchangeable.

MUX makes the distinction especially visible. Counting at the carrier connection gives the total encoded carrier traffic, including framing and control data, but it cannot honestly attribute those bytes to individual logical children. Counting at the logical links gives useful per-flow values, but those values are no longer the traffic of the carrier itself. Adding the two together would simply double-count different observations of overlapping data.

UDP has a similar problem. One association may carry datagrams for multiple destinations, so a plain byte-stream connection model is insufficient whenever accounting must preserve packet and destination semantics.

Another difficult area is write accounting on error paths.

`net.Conn.Write` returns `(n, err)`, so it exposes the prefix accepted at that particular connection boundary. This does not imply remote delivery, but it does allow the result of that `Write` call to be accounted for precisely.

`buf.Writer.WriteMultiBuffer`, however, returns only `error`. If the writer accepts part of a `MultiBuffer` and then fails, a generic wrapper cannot determine the accepted prefix.

#6747 handles this particularly well for TUN UDP: it submits packets individually and increments the counter only after the `WriteMultiBuffer` call for that packet succeeds. Therefore a later failed packet does not erase the already accepted prefix.

That solution works because the semantic unit is a datagram. It does not directly solve the general stream / `MultiBuffer` case.

`buf.Copy` is also attractive as a common accounting point, but its current `onData` callbacks run before the lower `WriteMultiBuffer` completes. Therefore the resulting value is semantically closer to **offered bytes** than to guaranteed **accepted bytes**. On the successful path those numbers match; on error or partial-write paths they may not.

The read side is somewhat easier: if `ReadMultiBuffer` returns data together with an error, the returned bytes are still observable at that reader boundary. The write side currently lacks the equivalent accepted-length result.

I also think wrapper transparency deserves a wider audit. `CounterConnection` hiding the TUN native buffer interfaces produced a confirmed functional bug, but other wrappers can also hide optional capabilities such as `buf.Writer`, `buf.TimeoutReader`, `syscall.Conn`, `io.ReaderFrom`, `Close`, `Interrupt`, or protocol-specific fast paths. Outside the confirmed TUN case I would treat this as an investigation target rather than claim that every such path is currently broken.

For the same reason, increasing `buf.Size` from 8192 to 16384 may change allocation frequency, `readv` / `writev` behavior, and TLS/XHTTP packetization, but it does not by itself resolve the accounting-boundary problem. I think that change needs measurements on representative paths rather than being treated as an architectural answer.

The connection-level API has the same naming problem. Before exposing a connection ID and associated counters, it would help to decide whether the object represents:

- a physical transport connection;
- an authenticated inbound connection;
- a logical proxied flow;
- a UDP association;
- a MUX carrier;
- or a logical child inside MUX/XUDP.

The lifecycle meaning also matters. A close request, context cancellation, socket close, protocol-session close, and complete owner quiescence are not necessarily the same event. An API event named `DISCONNECTED` should not choose one of those meanings implicitly.

Because of this, I would be cautious about introducing one global notion of per-connection traffic without an explicit entity type and byte scope.

The most robust model in my fork so far has been:

- logical-flow traffic and carrier/connection traffic are separate observations;
- every counter explicitly names the boundary where it is measured;
- incompatible scopes are never added together automatically;
- an unknown partial acceptance remains explicit instead of becoming zero;
- identity, byte accounting, and lifecycle are related, but they do not replace one another.

I am not proposing this as a finished upstream design. This is only the result of working through the current data paths and their error/lifecycle cases.

If this direction moves forward, I can prepare a smaller upstream-oriented comparison of the existing accounting boundaries and focused problem cases: TCP, multidestination UDP, MUX/XUDP, `buf.Copy`, partial writes, richer-interface preservation, and the current `readv` / `writev` paths.

svlavr:
In practice, this turned out to be more than an experiment with different counters. Once I started implementing the statistics in the private fork, several concrete problems appeared that cannot be solved only by choosing between `net.Conn` and `buf.Writer`.

For example, the current `SizeStatWriter` adds the full `MultiBuffer.Len()` before calling the lower writer. If the write fails, the counter has already been increased. So the first question is whether that value means offered bytes or bytes actually accepted by the next boundary.

Direct splice has a different problem: it bypasses the normal logical writer. If accounting is attached only to `transport.Link.Writer`, that traffic is missed. In my fork I had to add a separate progress hook for successfully transferred splice chunks without disabling the fast path itself.

I also had to handle counter continuity explicitly. If an accounting callback is lost, a counter overflows, retained history is evicted, or a consumer misses events, it is unsafe to continue returning a number as if the series were still complete. I ended up using explicit series generations, a coverage state, and snapshot-based resynchronization. Otherwise two perfectly normal-looking samples cannot safely be used even to calculate a rate after a reset or gap.

The event stream has a similar issue. Events should not be the only source of byte totals because they may be lost or evicted. In my implementation the cumulative counters are separate from events. Events are notifications, and after a gap the consumer receives a watermark and resynchronizes from a snapshot.

So if this kind of statistics is developed further in upstream, I think these questions need to be solved together:

- where each byte scope is measured;
- how partial writes are represented;
- how splice and other bypassing fast paths are accounted for;
- how accounting loss is exposed;
- how per-flow or per-destination state is bounded;
- and how a consumer recovers after missing events.

Otherwise it is easy to build an API that reports plausible values most of the time but silently becomes wrong on an error path, during splice, or after an event gap.

I am also working separately on targeted flow closing. That has its own hard problem: a logical flow may live inside a shared MUX/XUDP carrier. Closing the carrier is easy, but it may also kill unrelated siblings, while the logical child does not necessarily have its own socket that can simply be closed.

I look at how similar problems are handled in other proxy cores and networking projects. I see them more as "analogues" and sources of ideas than as competitors. But their connection models cannot be copied directly into Xray because the boundaries between a logical flow, a carrier, and a physical connection are different here.

If there is interest, I can later bring smaller, isolated cases and tests, both for statistics and for targeted flow closing.

Meo597:
> ~看了下代码，core 里一整个流量统计都得重构为对“单连接”的统计（buf.Copy），附带对“单连接”的控制 API #5844 @Meo597~

好的，等我上岸了之后看看，好像是个大工程

PR #6745, opened 2026-09-10
XDRIVE transport: local storage

Risaro:
`infra/conf/transport_internet.go` uses `xdrive.Config`, but the file does not import `transport/internet/xdrive`, so the package does not build on this branch:

```
infra/conf/transport_internet.go:472:13: undefined: xdrive
```

Adding the import fixes it.

PR #6748, opened 2026-09-11
XDRIVE transport: Add the Google Drive and "template" backend

Risaro:
This stacks on #6745. Until that one is merged the diff here also contains its commit.

### What it adds

A `Storage` implementation for Google Drive, selected with `"service": "Google Drive"`, where `remoteFolder` is a Drive folder id and `secrets` holds ClientID, ClientSecret and RefreshToken.

### No SDK

Plain `net/http`, zero new entries in `go.mod`. It sits behind the same five method `Storage` interface as the local backend, so another provider is one more file and still no dependency.

### How it works

Paths are flattened into single names, so `streams/abc/c2s/000000000.seg` becomes `streams~abc~c2s~000000000.seg`, because Drive has no cheap path lookup and creating nested folders would cost a request per level.

A name to id cache is filled from every listing, so a download costs one request instead of two. The access token is cached and refreshed on expiry or on a 401.

Requests are retried on 429 and 5xx with a jittered backoff capped at 2 seconds. Jitter matters here: at concurrency 32 every request hits the rate limit at the same moment, and synchronised retries only make it worse. A request whose context is already cancelled is not retried at all.

### Measurements

Both ends on a server in the Netherlands, against a real Google Drive account.

Raw storage:

* 4 sequential puts of 256 KiB: 199 KiB/s
* 8 parallel puts of 256 KiB: 1637 KiB/s
* 8 parallel gets of 256 KiB: 2453 KiB/s
* listing a folder of 13 objects: 250 ms
* an uploaded object becomes listable after about 200 ms, so the listing is fresh enough to drive the log

Transport, 10 MB pushed through the WAL:

* segment 128 KiB, concurrency 16: 1187 KiB/s
* segment 256 KiB, concurrency 16: 1513 KiB/s
* segment 256 KiB, concurrency 32: 1849 KiB/s
* segment 512 KiB, concurrency 32: 2111 KiB/s

Real proxy, curl through SOCKS into VLESS over XDRIVE over Google Drive:

* 50 MB in 21.6 s, 2258 KiB/s, about 18.5 Mbit/s, byte identical
* the first byte takes about 2 s, so bulk transfer is comfortable and interactive use is not

Mux, 30 parallel requests, which is roughly one web page:

* without mux: 23675 ms and 30 sessions
* with mux.cool at concurrency 8: 8605 ms and 4 sessions

Stability: 12 of 12 live runs against Drive.

### Tests

A fake Drive server on `httptest` covers upload, listing, download, delete, token caching and retries, and the whole transport is run over it. Live tests against a real account live in the same file and skip unless `XRAY_XDRIVE_DRIVE_SECRETS` points at a credentials file, so nothing here needs secrets in CI.

### Known limits

Credentials sit on the client, so this is for self hosted use. One outbound should enable mux.cool, otherwise every TCP connection becomes its own session with its own files. The HTTP client does not go through Xray's own dialer yet, so `sockopt` and a dialer proxy are not applied to Drive requests.

### DIFF
--- a/infra/conf/transport_method.go
+++ b/infra/conf/transport_method.go
@@ -794,3 +794,9 @@ func readFileOrString(f string, s []string) ([]byte, error) {
 	}
 	return nil, errors.New("both file and bytes are empty.")
 }
+
+type XDriveConfig struct {
+	RemoteFolder string   `json:"remoteFolder"`
+	Service      string   `json:"service"`
+	Secrets      []string `json:"secrets"`
+}


Risaro:
brad-pitt

RPRX:
> The HTTP client does not go through Xray's own dialer yet

这一步直接写成 https://github.com/XTLS/Xray-core/pull/5645#issuecomment-3849778103 https://github.com/XTLS/Xray-core/pull/5645#issuecomment-3851839033

RPRX:
> Paths are flattened into single names, so `streams/abc/c2s/000000000.seg` becomes `streams~abc~c2s~000000000.seg`, because Drive has no cheap path lookup and creating nested folders would cost a request per level.

~~我感觉至少弄一层子文件夹吧别全放根目录了~~

Fangliding:
大量 vibe code incoming

RPRX:
人家虽然 vibe 但说了会人工 deVibe 不挺好的吗，还能在真实 Google Drive 环境中测试，~~反正是完全新的传输层，更无所谓了~~

~~不像我看了 https://github.com/XTLS/Xray-core/pull/5654 后发现自己甚至懒得 deVibe，更懒得弄真实 Google Drive 环境去测试~~

Risaro:
@RPRX  Update on latency. Small segments now ride in the file's description instead of the file content, so discovery and download collapse into a single files.list call. Google returns description in the listing at no extra cost, and it accepts up to 16384 characters, which covers TLS records and HTTP headers comfortably. Larger segments still go as media.

Measured on a server in the Netherlands against a real Drive account, before and after:

- HTTP first byte, session already established: 4.1-4.7 s became 1.86 s
- HTTPS to cloudflare.com: 9.39 s became 3.80 s, of which TLS 4.71 s became 2.05 s
- HTTPS to google.com: 15.85 s became 3.23 s
- Bulk transfer of 20 MB: about 19-24 Mbit/s, byte identical

For reference, the same requests without the tunnel answer in 0.03 s.

I think this is the floor on Google Drive. A single traversal is one write plus one read, and Drive's own per-call latency is what is left:

- create a file with description: 686 ms
- patch the description of an existing file: 541 ms
- files.get by a known id: 229 ms
- files.list by prefix: 260 ms

Network round trip from the Netherlands to Google is 10-30 ms, so these numbers are Drive's own processing, not the link. An HTTP request needs two traversals, which is the 1.86 s we measure.

I also tried the manifest shape you described. Writing through files.update on a pre-created file and reading it back by id gives 770 ms per traversal against 946 ms now, so about 19 percent. I did not take it: a single manifest per stream serialises writes and kills the parallel segment upload that the throughput depends on, and a pool of pre-created files makes session setup much more expensive.

Also fixed in this round: Drive reports rate limiting as 403 with a userRateLimitExceeded reason rather than 429, so those were treated as fatal; one Drive client is now shared by all connections instead of one per connection; announcement deletion no longer sits on the accept path; and a segment whose upload fails now leaves a marker so the peer fails fast instead of waiting forever. Under 100 concurrent connections without mux this took the success rate from 22 out of 100 to 99 out of 100.

One important caveat about the throughput numbers: they depend on the link of the machine running the node, not on the transport alone. All measurements above come from a server on a 300 Mbit link, where a single stream pulls about 95 Mbit/s from Google on its own. On a thinner uplink the picture changes completely. My first measurements were taken from home, where the uplink is the bottleneck, and they showed 60 KiB/s instead of the 1637 KiB/s the same code does from the server. So treat the numbers as an upper bound for a well connected node, not as what an arbitrary client will see.

With that, I am stopping work on the Google Drive backend here. The remaining ceiling is Google's own behaviour, not the design: their per-call latency is 229 ms at best and 686 ms for a write, and no protocol shape on top of Drive removes that. Everything that could be won without changing the medium has been won, and what is left is a 19 percent gain that costs the throughput.

I am moving my attention to another transport. The Storage interface stays as it is, so whoever wants a faster service can drop it in as one file, and it will immediately show whether the medium or the design was the limit.

Risaro:
@RPRX I’ll be rewriting all this crap by hand soon; I’ve seen enough of what the AI ​​can do to know its limits, so from here on out, I’ll be tinkering with the code manually. For now, I want to mess around and test out all the routes in this branch to understand their limits, and then comes the big, painstaking job of rewriting it all by hand.

I find it easier to rewrite things than to write standard methods from scratch.

Risaro:
@RPRX Understood on the dialer. Routing the Drive requests through Xray's own TLS stack rather than a plain http.Client, so that SNI, verifyPeerCertByName and the h3 choice come from the TLS section and only the inner HTTP host is filled in, makes sense to me. I read both of your February comments, and the domain fronting angle is a much bigger deal than the dialer detail itself: it means the transport stops depending on whether googleapis.com is allowed. I will do it that way.

I will also take the rest of what you asked, and I will pack more services into #6748 rather than splitting them out, since you would rather have them together.

That is for later though. I am done with my research for today, so I will pick this up tomorrow and answer your open questions then, including what I meant by the xhttp method.

RPRX:
@Risaro 修改 file desc 能批量/有速率限制吗，~~比如一次改一千个~~，另外 googleapis.com 是否支持 domain fronting 还需测试

Risaro:
@RPRX  I get it now - I’m going through the comments one by one; we’ll check everything out tomorrow, and I’ll rewrite some of the functions. As for the packages -yes, that’s possible.

Risaro:
Sorry guys, I got more tasks at work; I'll be back on Monday.

Risaro:
giphy

Risaro:
@RPRX Done, the Drive requests go through Xray's own dialer now.

internet.DialSystem does the dialing, so sockopt and a dialer proxy apply. The TLS section decides everything about the handshake: SNI, verifyPeerCertByName, the uTLS fingerprint, and REALITY. The connection target comes from address, and the service is identified only by the HTTP host of the request, which is what makes domain fronting work.

I tested it against the real thing rather than a mock. With

"address": "www.googleapis.com",
"port": 443,
"security": "tls",
"tlsSettings": { "serverName": "www.google.com" }

the connection is made with www.google.com as the SNI and Drive still serves the request, because only the inner HTTP host says where it should go. There is also a unit test that points the connection at a local server while the request URL keeps the API hostname, so the two cannot be confused.

HTTP/1.1 and h2 are chosen from ALPN. h3 is not wired yet: it needs quic-go and an http3.Transport, which I would rather do as its own change.

This also moved the numbers, measured on a server in the Netherlands against a real Drive account, before and after:

- HTTP first byte, session already established: 1.86 s became 1.57 s
- bulk transfer of 20 MB: 2359 KiB/s became 2915 KiB/s
- HTTPS to cloudflare.com: 3.80 s became 2.92 s
- HTTPS to google.com: 3.23 s became 2.99 s

The gain is h2 doing what it should: the thirty-two concurrent segment requests now share one connection instead of opening a TLS handshake each.

No new dependencies. golang.org/x/net was already required and splithttp and tls already use http2, so go.mod and go.sum are untouched.

Risaro:
I'll check the spike now; I'm going to look at the speed regarding collaborative editing of Google Docs with anonymous users.

RPRX:
@Risaro 振奋人心！需确保内层 host 始终固定而不由 address 来决定，请检查一下，~~另外 fmt 需要修一下~~

Risaro:
@RPRX A side note on the "write into a doc, read from the other end" idea, since you mentioned abusing any site, not just proper storage.

I looked at whether Google Docs could be driven anonymously, to get rid of the credentials-on-the-client problem. What I found so far, measuring against a real doc shared as "anyone with the link can edit":

- The plain Docs REST API is a dead end for speed. batchUpdate write is ~550 ms and documents.get read is ~500 ms even with fields narrowed to just the revision id, so a round trip is worse than the Drive backend (~920 ms vs the ~1 s I have now), and the response is ~1.75x bloat on top of base64. No reason to use it.
- The interesting part is the realtime collaboration channel, because it is client-initiated long-polling: push instead of my current polling, and no public endpoint needed, so the no-public-IP property survives. Loading the doc anonymously (no account, datacenter IP) is not blocked - Google serves the full page with a session id and the document model chunk right in it, so reading a doc anonymously is clearly feasible.
- Writing anonymously is the wall. The anonymous page does not expose the save/bind endpoint, an editor token, or the mutation machinery - a real browser mints those through follow-up requests and by running Google's JS. Replicating that means reversing the browserchannel handshake and emulating their client, which is fragile by nature and would break on any frontend change.

So the read direction is easy anonymously, the write direction is a real reverse-engineering project, and a transport needs both. It is not production material and I would not attach it to the current PRs. I am leaving it here as notes and may come back to it as a separate experiment later. The Drive backend stays the working path.

```
watafa man google translate is shit neyron
```

Risaro:
@RPRX Both done.

The inner host was already fixed and does not come from address -  the request URLs are the www.googleapis.com / oauth2.googleapis.com constants, and address only sets the outer TCP target and the SNI, never the HTTP Host header. But you were right to ask me to check, so I made it provable rather than a claim: the fronting test now points address at a local server while the API URLs keep their real hostnames, and the fake server records every Host it receives and asserts it is always www.googleapis.com. I confirmed the test actually catches a leak by temporarily forcing the host from address and watching it fail.

For the format, the check is infra/vformat, not plain gofmt, which is why it slipped past me locally - one test file needed a trailing comma and a closing brace on its own line. Fixed, and vformat -mode check is clean now.

RPRX:
@Risaro 其实 ~~abusing Docs~~ 的吞吐量可能不及 Drive，但如果能完全匿名则是它的优势，既然 XDRIVE over Google 已经能用 www.google.com 这个 SNI，那么 ~~abusing Docs~~ 可以暂放，若要可以进行一些预先研究可以优先~~滥用~~那些没有网盘的服务

另外我说那个 host 的意思是，address 仅解析/提供 IP 地址，但如果其它存储服务能自定义 host，则 host 应放进“参数”数组，还有一点和其它传输层不同：address 不应决定 SNI，SNI 的默认值应当与 host 相同，除非在 TLS 层手动指定了 SNI，请修改

fatyzzz:
guys, allow Risaro write to this PR
image

Fangliding:
最近有一些bot号搞破坏 那我先关掉吧。。

fatyzzz:
Risaro asked me to pass this along:

he’s done with the Google part and finished the PR fixes.

Next, he’ll finish the route with Mail Disk sometime this week.

It’s his birthday this week 😅, so everything should be ready around Saturday.

By then, he’ll also rewrite it manually.

Risaro:
Check

Risaro:
xdriveSettings:
  template:
    vars:          
      # {name} {id} {folder} {data} {token} {secret0..n} {cursor}
    auth:
      type: oauth2-refresh        
      tokenUrl: "https://oauth2.googleapis.com/token"
      form: { grant_type: refresh_token, refresh_token: "{secret2}", ... }
      extract: { token: "$.access_token", expiresIn: "$.expires_in" }
    put:
      method: POST
      url: "https://www.googleapis.com/upload/...?uploadType=media"
      headers: { Authorization: "Bearer {token}" }
      body: { place: description|multipart|raw, encode: base64 }
      extract: { id: "$.id" }
    get:
      method: GET
      url: "https://www.googleapis.com/drive/v3/files/{id}?alt=media"
      extract: { data: "$body" }
    list:
      method: GET
      url: ".../files?q='{folder}' in parents and name contains '{prefix}'"
      extract: { names: "$.files[*].name", inline: "$.files[*].description",
                 nextCursor: "$.nextPageToken" }
    delete:
      method: DELETE
      url: ".../files/{id}"
    errors:
      retryOn: { status: [429, 500, 502, 503], jsonReason: "$.error.errors[*].reason" }
      
Remember I said I wanted to implement it like xhttp? Does it make sense now?
@RPRX

Risaro:
@RPRX Added a template backend so a store can be described in config instead of compiled in Go. Any WebDAV/S3/plain-HTTP store that addresses objects by path works with a JSON template - no code. Google Drive stays a compiled backend behind the same Storage interface (its OAuth + opaque-id + JSON is beyond a sane template).

The template declares the auth (none / static / basic / oauth2), the four operations with URL and header templates, and a regex to pull names out of a listing (works for WebDAV XML, S3 XML, JSON alike). Variables {folder} {name} {prefix} {secret0..n} {token} {data} are substituted into any field. concurrency is set in the template too. The existing dialer applies, so sockopt, dialer proxy, SNI and fronting work for templated stores as well.

Example - Yandex.Disk over WebDAV:

"xdriveSettings": {
  "remoteFolder": "xray",
  "service": "template",
  "secrets": ["you@yandex.ru", "app-password"],
  "concurrency": 32,
  "template": {
    "flatten": true,
    "concurrency": 32,
    "auth":   { "type": "basic", "username": "{secret0}", "password": "{secret1}" },
    "put":    { "method": "PUT",      "url": "https://webdav.yandex.ru/{folder}/{name}" },
    "get":    { "method": "GET",      "url": "https://webdav.yandex.ru/{folder}/{name}" },
    "delete": { "method": "DELETE",   "url": "https://webdav.yandex.ru/{folder}/{name}" },
    "list":   { "method": "PROPFIND", "url": "https://webdav.yandex.ru/{folder}/",
                "headers": { "Depth": "1" },
                "namesRegex": "<d:href>[^<]*/([^/<]+)</d:href>" },
    "retry":  { "status": [429, 500, 502, 503] }
  }
}

oauth2 auth for token-based stores looks like:

"auth": {
  "type": "oauth2",
  "tokenUrl": "https://oauth2.googleapis.com/token",
  "form": { "grant_type": "refresh_token", "client_id": "{secret0}", "client_secret": "{secret1}", "refresh_token": "{secret2}" },
  "tokenPath": "access_token", "expiryPath": "expires_in",
  "header": { "Authorization": "Bearer {token}" }
}

I ran it against real Yandex.Disk WebDAV, from both a Netherlands and a Russian server. It works end to end (full transport over it). Two honest notes: Yandex WebDAV needs a paid tariff (a free account returns 402 on writes), and Yandex throttles WebDAV upload to ~15 KiB/s per connection, so throughput only scales with concurrency (~440 KiB/s at 32) and is the same from RU and NL - it's a Yandex policy, not latency. Download is fast (~2 MiB/s). So the engine is proven on a real store; that particular store is just a weak uplink.

No new dependencies, go.mod untouched. Tests: a fake path store (put/get/list/delete, basic auth, oauth2 token cache, full transport) plus an env-gated live Yandex test that skips in CI.

Risaro:
image

Risaro:
Shout-out to all the vultures who messaged me on Telegram asking how xDrive works and which cloud methods are faster or better -some even offered money. In short: go fuck yourselves 🖕.

You shouldn't have messaged me on Telegram; now this transport method is going public - go ahead and configure it yourselves.

I did this so my disk-based methods wouldn't get patched/exposed; you can thank the vultures for that - this transport method could have lasted forever.

You'd be better off just having a beer with me.

RPRX:
更通用当然更好，可能还需开放更多可配置的 HTTP 相关参数，话说回来，应该加入 XMUX 以控制 XDRIVE 的 H2/H3 复用

RPRX:
为了凑到第 2000 个 commit ~~以及给 Risaro 的生日补个礼物~~，我打算现在就分两步合并 XDRIVE 相关的 PR 进 main

代码仍需完善，比如上面我提到的 XMUX，以及 config 的“Google Drive”改为“google-drive”，以及检测配置：需开启 mux.cool

~~可能还需要一些 deVibe~~，好在代码一旦进入 main 就会开始接受更广泛的 PR，会有很多人来完善，就像 XHTTP 那样

还需配套的文档、Google Drive API 配置教程等，@Risaro 请发出你的 ETH 地址，我会给你转一个 Project X NFT 作为纪念

**XDRIVE 将会是 Project X 2026 最具代表性的新协议，其次是 Finalmask、TUN，以及 FinalRules、pcs 等各种默认的安全改进**

Risaro:
0xac5bbFaF2621523131018F90E7FB74672714Caf3

ETH
@RPRX

Risaro:
Thanks a lot for the birthday wishes! I’ll definitely be back - I’m celebrating on Monday and Tuesday -and I’ll submit new pull requests for `mux`. As for final masks, I’m interested in that too; I’ll likely contribute to that effort as well.

@RPRX

RPRX:
@Risaro 0xe5554cceca6db722bebc4028b466d20e33d63029534ecb5875218943742cca90

我觉得最可贵的是能实时测试 Google Drive 等 backend 并改进 XDRIVE 的代码，~~是否 vibe 的倒不重要，可能明年全靠 AI 都行~~

目前比较需要一份 Google Drive API 配置教程，半年前我和 @iambabyninja 通过邮箱讨论过，但不确定是否适用于这个 PR

Risaro:
@RPRX hey again! I was thinking -  what about XMessage?

Most messengers don’t rate-limit notes, and they usually allow enough characters to store the segments. Plus, there are common APIs for these services and stuff like that :) 

I knowed it s  a sharp :) what about for this idea ? 

And it's a mega fast :0

RememberOurPromise:
~XCloudflare~

api.cloudflare.com

https://developers.cloudflare.com/api/resources/d1/
https://developers.cloudflare.com/api/resources/r2/

RPRX:
@Risaro 之前讨论过，“利用聊天/邮箱等服务”这类直接并入 XDRIVE 就行，区别是这些服务可能有主动推送而不用自己去扫 list

~~不过对于聊天软件，审查者本来就严管，比如中国就把境外的全封了，即使用境内的，比如微信语音会检查语音数据是否合理~~

XDRIVE 目前需改为默认 h2+uTLS（若 ALPN 仅 h1 那么只在内层用），以及基于魔改版 quic-go 库引入 h3 https://github.com/XTLS/Xray-core/pull/6754#issuecomment-5751585906

另外需加上配置检测：强制 VLESS Mux 以及 VLESS Encryption，后者是因为瓶颈不在终端，树莓派跑都没压力，默认更安全吧

由于有 mux 所以也就一条 enc，后者自带可配置 padding 以淡化初次握手的长度特征，以及改成“google-drive”方便分享链接

我们这边也会按计划增强一下 mux，加些选项、改改默认值，不止是 XDRIVE 必选，用处还挺多的，今年内也会加进分享链接

关于上文提到的 XMUX，大概需要把 XHTTP 里的相关代码抽出来做成独立模块，以及跨传输层联动“上下行分离”，晚点会实现

PR #6750, opened 2026-09-11
Clean-up GPL components as a new base

KobeArthurScofield:
This pr aims to clean up remaining GPL-licensed components and the remnants, providing a clean base, or as a reference.

**DO NOT merge before the substitute solution is on the horizon.** Merging it before the time will cause function loss accidentally.


Fangliding:
~~无非就是替换掉ss2022的实现而已 不用专门做一个删掉它的~~
或者把这个分支放在主仓库作为参考就行了（如果暂时有需要不违反许可证的需求）

PR #6755, opened 2026-09-13
XHTTP: Add opt-in downlink resume

Medium1992:
# XHTTP: Add opt-in downlink resume

Closes the failure mode measured in #4846. Off by default, `packet-up` /
`stream-up` / `stream-one` unaffected when disabled.

## Problem

In every XHTTP mode the downlink is one stream-down GET that lives as long as
the tunneled connection. A middlebox that resets that stream takes the
connection with it.

#4846 measures what the middlebox does: an **idle timer on the response
stream**, reset by any body byte.

None of them tells the origin. Neither `request.Context().Done()` nor a failing
write fires, so the server cannot detect or pre-empt the cut.

Worst case is an upload: in `packet-up` the uplink never idles, but the
downlink carries proxied payload only. During an upload it carries nothing,
idles for the whole transfer, and is cut. The upload dies with it.

#6422 solved this with a keepalive byte and was closed: *"introducing a frame
format just to keep alive an overly idle downlink connection is
inappropriate"*. #6554 closed for the same reason. **This PR adds no framing
and does not touch the response body.**

## Mechanism

1. Server keeps the session and a replay buffer of bytes the client has not
   acknowledged, instead of ending the session when the GET ends.
2. Client reopens stream-down carrying the offset it actually received.
3. Server replays from that offset and continues writing into the new stream.
4. Every uplink request carries the client's offset; the server releases
   retention up to it.
5. `scMaxDownlinkSecs` (optional) closes the stream from the client side before
   the edge does, at a randomised interval.

The reader on the client hides the boundary, so the tunneled connection sees an
uninterrupted byte stream.

## Wire format

| Direction | Carries | Notes |
|---|---|---|
| stream-down request | offset | Same placement as `sessionID` (header / query / cookie), key `downlinkResumeKey` |
| stream-down response | echoed offset | Sent **only** when the offset arrived |
| uplink request | offset | Acknowledgement; releases retention |
| `410 Gone` | — | Offset no longer servable; client fails instead of retrying |
| `204 No Content` | — | Session finished; client ends as on a complete response |

`Range: bytes=N-` was the obvious alternative and was tested against a
production CDN: the edge answered `206`/`416` from its own copy and **removed
the header before the origin**, while an ordinary header passed through. A
stripped offset would replay a live stream from byte 0, so the offset keeps its
own header, and two rules make a hostile path safe:

- the response echoes the offset only when it arrived, so such a path is
  indistinguishable from a server without the feature and the client stops
  resuming;
- a stream-down request for a session that has already sent bytes but carries
  no offset is refused, never served from 0.

## Changes

| File | Change | Why |
|---|---|---|
| `downlink.go` *(new, 206)* | `downlinkBuffer`: retains unacknowledged bytes in pooled `buf.MultiBuffer`, replays from an offset, survives a failing writer | Server side of resume; pooled so retention costs what it holds |
| `resumable.go` *(new, 207)* | `resumableDownlink`: `io.ReadCloser` that reopens the request on error and continues at its offset | Hides stream boundaries from the tunneled connection |
| `downlink_test.go` *(new, 250)* | 9 tests: replay, trimming, retention bound, finished buffer, back-pressure release, concurrent attach/detach/write | — |
| `hub.go` | Session outlives the GET; grace timer; resume/finished/refused branches; offset acknowledgement on uplink | Server lifecycle |
| `client.go` | `OpenDownlink`, offset on the request, `Rejected` / `Supported` / `Finished` on the response | Client transport |
| `dialer.go` | Wires `resumableDownlink` in when both ends enable it; falls back otherwise | Opt-in wiring |
| `config.go` | Normalisation for the new fields, placement/key helpers reusing the `sessionID` ones | No new config convention |
| `config.proto`, `config.pb.go` | 4 fields | — |
| `infra/conf/transport_method.go` | JSON plumbing (12 lines added, none modified) | — |
| `browser_client.go` | Interface signature | Browser dialer cannot resume; warns and falls back |

## Configuration

```json
"xhttpSettings": {
  "extra": {
    "scDownlinkResume": true,
    "scMaxReplayBytes": 262144,
    "scMaxDownlinkSecs": "18-25",
    "downlinkResumeKey": "X-Offset"
  }
}
```

| Field | Default | Meaning |
|---|---|---|
| `scDownlinkResume` | `false` | Enable. Required on both ends, like `mode` |
| `scMaxReplayBytes` | 256 KiB | Retention per session |
| `scMaxDownlinkSecs` | off | Client-side rotation range, seconds |
| `downlinkResumeKey` | `X-Offset` / `x_offset` | Key for the offset, placement follows `sessionID` |

Grace period after a stream ends: 30 s, matching the existing session reaper.

## Memory

Retention is bounded per session and released on acknowledgement:

**ceiling = concurrent sessions × `scMaxReplayBytes`**

256 KiB is 1/120 of what XHTTP already allows per session on the uplink
(`scMaxBufferedPosts` × `scMaxEachPostBytes`, 30 MiB by default).

Measured, 200 sessions downloading at 50 KB/s against a fast server for 100 s —
the worst case, because a pure download sends no uplink and therefore never
acknowledges:

| | server RSS |
|---|---|
| feature off | 74 MB |
| feature on | 150 MB |

Sessions with any uplink traffic acknowledge continuously and hold kilobytes.
Repeated bursts of ~630 short sessions leave RSS flat (43.6 → 43.2 → 42.5 MB).

## Degradation

| Condition | Behaviour |
|---|---|
| One side without the feature | Response carries no offset, client warns once and runs as it does today |
| Path strips the offset | Same as above; verified against an edge that deletes the header: 80/80 transfers byte-exact, 0 refusals |
| Client falls outside the retained window | `410`, connection fails cleanly after one attempt, no retry storm |
| Browser dialer | Warns and falls back |

Truncation is possible in the degradation path; corruption is not.

## Testing

Bench: patched client and server with a reverse proxy modelling the edge —
resetting a response stream idle beyond a randomised threshold, and keeping the
origin request running afterwards, as a real edge does.

| Scenario | Feature off | Feature on |
|---|---|---|
| 600 MB upload, downlink idle, cuts 8–15 s | dies at 9.2 s, `curl 56` | completes in 70.9 s |
| Long-poll, 40 s delayed response | never arrives, `curl 52` | `200` at 40.06 s |
| Trickle, 45 s gaps vs cuts 8–31 s | — | 3/3 ticks, 0 gaps |
| 30 × 20 MB concurrent, cuts 8–31 s | — | every hash exact |
| 160 connections closed by the remote first | one wasted round trip each | 160/160 byte-exact |
| Edge holding the origin request open | — | 3 rounds, 0 refusals, 0 gaps |
| Retention overrun | — | `410`, clean failure, logs stable |

Production, 6.5 minutes of browsing and video through a CDN:

- 556 resumes, **all on the first attempt**
- 1 refusal (a client that reattached 44 s after its stream was cut)
- 0 unacknowledged, 0 errors
- interval between reconnections: 30.0 s in 160 of them — that edge's idle timer

## Known limitations

1. **Traffic that is purely downstream.** With no uplink there are no
   acknowledgements, so retention is released by cap alone and a transfer at
   full rate outruns the window. It then fails exactly as it does today. A
   periodic acknowledgement-only request would fix it but needs a request shape
   the server does not read as uplink data.
2. **Both ends must opt in**, like `mode`.
3. **No auto-detection of the edge's threshold.** Resume already covers the
   unpredictable case; rotation is the latency optimisation on top.
4. **`packet-up` focused.** `stream-up` carries its uplink as one long request,
   so the acknowledgement has no natural place there yet.


Jolymmiles:
closed as not planed

Fangliding:
这个重连复杂度太高了

Medium1992:
> 这个重连复杂度太高了 没法保证不丢字节

I tested it on a live server via a CDN and on the client side, and fixed all the bugs I could find.

I'm not a programmer, but I'd like to see this feature added to the core. I created a PR to immediately demonstrate the idea and its implementation in code.

Version compatibility has also been taken into account to ensure that the implementation of this feature will not affect the operation of previous kernel versions.
Therefore, I would appreciate a more detailed review.
Please let me know if there are any suggestions for what to correct or remove.

In the current release of the kernel, the logs show the following:
```
2026/09/12 21:36:20.608760 [Info] [1689181628] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > proxy/vless/encoding: failed to read packet length > stream error: stream ID 1; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.608838 [Info] [4173552224] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > stream error: stream ID 7; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.608861 [Info] [1511575834] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > stream error: stream ID 13; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.608881 [Info] [1556905573] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > stream error: stream ID 5; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.857399 [Info] [683396471] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > proxy/vless/encoding: failed to read packet length > stream error: stream ID 3; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.857504 [Info] [3556331530] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > proxy/vless/encoding: failed to read packet length > stream error: stream ID 5; INTERNAL_ERROR; received from peer
2026/09/12 21:36:20.857526 [Info] [2844643601] app/proxyman/outbound: app/proxyman/outbound: failed to process outbound traffic > proxy/vless/outbound: connection ends > proxy/vless/outbound: failed to transfer response payload > proxy/vless/encoding: failed to read packet length > stream error: stream ID 7; INTERNAL_ERROR; received from peer
```

Every 30 or 10 seconds, depending on which CDN edge the connection passes through
Common problems in this regard include:
- Game connections drop, and the game reconnects
- The video stream freezes and buffering occurs; sometimes the player even crashes.

Fangliding:
一般情况下我们不接受纯ai的代码 你的pr message甚至看不出你真的读了里面列出的几个其他pr

Medium1992:
> 一般情况下我们不接受纯ai的代码 你的pr message甚至看不出你真的读了里面列出的几个其他pr

That was a trial implementation, and I should have started from what you suggested in #6554.

I've now tried Mux.Cool over XHTTP. Un-banning alone doesn't help, I checked it through the real CDN as well: a quiet Mux connection still gets cut, and then every session on it drops at once. SessionStatusKeepAlive is handled on both sides but never sent. With the server sending it every few seconds, the cuts are gone: 13 minutes through the CDN with no stream errors, where stock gave 10 in the first minute, and no more disconnects in mobile games. The client stays stock, with concurrency: 1 (the default 8 makes parallel downloads about 3× slower).

Commit: https://github.com/Medium1992/Xray-core-fork/commit/6aa4da56

Could Mux.Cool be allowed for XHTTP together with keepalive? The keepalive is sent only on XHTTP packet-up and stream-up, so raw, REALITY and stream-one get no extra traffic that could become a pattern.

RPRX:
~~某位俄罗斯友人~~在私信中提到了该问题，我的看法是，目前来看“恢复连接”过于复杂且有局限性，比如需解决延迟、ACK 等问题、以及没有覆盖 https://github.com/XTLS/Xray-core/issues/4846 的需求，此外如果通过 mux 或 vision 的 padding 格式来发 keep-alive 包固然可以，**但 XHTTP 作为一个独立的传输层，在设计哲学上不应过度依赖特定的上层协议来解决自身存在的问题**，此外对 CDN 来说上行可以不活跃但下行得活跃，所以是时候开始设计 packet-down 的格式了，但不是 meek 那种，而是在 stream-down 中跑 packet-down，这也并不意味着直接实现了 https://github.com/XTLS/Xray-core/issues/4846 所需的复杂策略、multi-path-down 甚至多个相同包竞速，就像 packet-up 早有这些潜力但还没设计那样的配置，packet-down 的帧格式也需要保留这些潜力，大概五个要素：可省略的 session id 节选、可省略的包序号、不可省略的类型即 padding/data、不可省略的 length、实际数据，现在主要是利用该框架来发送下行的 keep-alive 包，obfs 是后话

~~但可以先开放周期、长度 range、padding 内容，至于服务端备份、客户端重排序、ACK 等，等拓展了“高级用法”再实现吧~~

Issue #6732, opened 2026-09-07
Proposal: user-configurable traffic shaping profiles for XMC / FinalMask (follow-up to #6487)


BolTun-prog:
The #6487 summary ends with:

> This intentionally does not emit version-specific Play or Chunk Data packets. **User-configurable traffic shaping is deferred to a follow-up PR.**

In the merge discussion @RPRX also noted that custom traffic-shape configuration would be added later as an incremental feature (流量形状自定义配置下次再补吧，毕竟是增量功能).

This issue proposes a concrete design for that follow-up and asks for design agreement before any implementation (a draft PR would follow if the direction is acceptable). If the original XMC author already has shaping work in flight, happy to stand down and review/help instead.

## Problem

The current XMC shaping is a built-in model of one fixed target (uncompressed Minecraft 26.1.2 login/join). Two practical limits:

- Built-in models decay when the camouflage target's traffic shape evolves; updating them requires core releases.
- Different deployments may need different camouflage targets, but today the XMC shape is a compiled-in preset and is not configurable (other masks, e.g. fragment #6334 or header-custom #5920, already take declarative config).

The knobs that exist today are scattered across features and do not compose with XMC's directional shaping: freedom fragment `maxSplit` (#4998), XHTTP upload pacing (`minUploadInterval`, #3592) and its padding mechanism (#3643), fragment `lengths`/`delays` (#6334), and the programmable `header-custom` templates (#5920).

## Proposal

Make the XMC shaping data-driven: a declarative profile loaded from config, interpreted by a small engine in `finalmask` — no policy logic in core.

Sketch:

```jsonc
{
  "schema": "xray.shape@1",
  "name": "xmc-default",
  "version": 1,
  "phases": {
    "login": {
      "c2s": { "lenDist": [ /* sampled length histogram */ ] },
      "s2c": { "lenDist": [ /* ... */ ] },
      "keepAlive": { "iatMs": 1000, "jitterMs": 100 }
    }
    // phase transitions follow the existing protocol state machine
  }
}
```

Runtime invariants (deliberately conservative):

- Padding only ever increases record length; shortening is fragmentation, which stays a separate mask.
- Zero added latency for real data: the profile only rate-limits sends (`iatMs` as "not faster than") and fills pauses with dummies. Delay-based scheduling is explicitly out of scope here.
- The turn schedule stays alternating-direction (client first), exactly like the merged XMC padding schedule; a profile only controls lengths and cadence, never the message flow.
[...]

PR #6769, opened 2026-09-14
XMC: make startup padding lengths configurable

spr334-a:
Follow-up to #6487, covering the startup-padding part of #6732.

Adds an optional `padding` array to XMC settings:

```json
{"padding": ["44", "64-128", "25", "32768-49152"]}
```

Entries are fixed lengths or inclusive ranges. Turns alternate direction, client first, and use the existing encrypted padding records. Both endpoints need the same array. The first length includes Login Acknowledged's two bytes; invalid ranges are rejected during config loading and connection setup.

An omitted or empty array preserves the existing 26.1.2 preset. Custom turns have no intentional delays. Data framing, keep-alives, and mask ordering are unchanged. No new dependency or wire-format change.

This deliberately leaves profile registries, version negotiation, and ongoing traffic scheduling for separate discussion. In particular, existing length checks are not a configuration-fingerprint handshake and do not detect every endpoint mismatch; the documentation calls this out.

Tests cover range validation, prefix and VarInt boundaries, default-preset preservation, and encrypted bidirectional round trips with single, odd, and even turn counts.

Checked on Windows with Go 1.27.0:

```text
go test -count=20 ./transport/internet/finalmask/xmc
go test -race -count=1 ./transport/internet/finalmask/xmc
go test -count=1 ./infra/conf
go test -count=1 ./transport/internet/finalmask/...
go vet ./transport/internet/finalmask/...
go build ./main
go run ./infra/vformat/main.go -mode check -pwd ./
```

`go vet ./infra/conf` reports existing unreachable code at `infra/conf/xray.go:606`, also reproduced on the base commit.


RPRX:
@Fangliding 你先看一下吧

BolTun-prog:
Sounds interesting

transport/internet/finalmask/xmc/config.go
```diff
@@ -23,10 +23,53 @@ func (c *Config) WrapConnServer(conn net.Conn) (net.Conn, error) {
...
+func paddingScheduleFromConfig(padding []*Padding) ([]paddingTurn, error) {
+	if len(padding) == 0 {
+		return nil, nil
+	}
+	schedule := make([]paddingTurn, len(padding))
+	for i, turn := range padding {
+		if turn == nil || turn.LengthMin < 1 || turn.LengthMax < turn.LengthMin || turn.LengthMax > maxPaddingTurnLength {
+			return nil, fmt.Errorf("invalid padding length range at turn %d", i)
+		}
+		direction := paddingClientToServer
```
BolTun-prog, line 49: I think, that switch case better use for this business logic

transport/internet/finalmask/xmc/packet_stream_test.go
```diff
@@ -45,7 +58,15 @@ func TestPacketStreamRoundTrip(t *testing.T) {
 
 	const password = "packet-stream-shared-key"
 	privateKey, publicKey := deriveTestRSAKey(t, password)
-	profiles := []loginProfile{testLoginProfile("packet_user")}
+	profile := testLoginProfile("packet_user")
+	config := &Config{
+		Password: password, RsaPrivateKey: privateKey, RsaPublicKey: publicKey,
+		Hostname: "localhost", Padding: padding,
+		Profiles: []*Profile{{
+			Username: profile.Username, Uuid: profile.UUID[:],
```
BolTun-prog, line 66:
```suggestion
			Uuid: profile.UUID[:],
```
Why are you using slice splice, I think, you shoud use `copy` for `profile.UUID`

spr334-a: I don't think a copy is needed here ,uuid isnt modified later in this test

RPRX:
@spr334-a ready 了吗，给文档开个 PR

PR #6772, opened 2026-09-16
XHTTP transport: CDN buffering bypass & KeepAlive

cerg2010cerg2010:
Some CDNs don't respect the 'X-Accel-Buffering: no' header, which makes data transfers from server to client unstable - GET method hangs because data sits in nginx's buffer and doesn't get sent out - it flushes only when connection to a CDN is closed on an upstream server.

Another problem is idle connections - CDNs don't respect HTTP/2 keepalive pings, so since we have a way to send data that can be ignored by client, we can send empty padding frames to kick server and prevent the connection from closing.

To configure:
- set scDownlinkKeepAliveHeader on server and on client. This HTTP header is used as an indication that client and server support this feature. Older clients will continue working without this feature.
- scDownlinkKeepAliveSecs is how often, in seconds, client wants a ping when downlink is idle. Client-only. Accepts integer range for obfuscation, however, delay value is selected on client and is kept as-is.
- scDownlinkFlushBytes is the size of a padding frame. Should be not less than a buffer on a CDN. Default buffer size on nginx is 4k, so try 4096-6144.
- scDownlinkFlushDelayMs sets how long the downlink must be idle before the flush fires. If any request comes in that time, the padding is delayed. Also accepts integer range.

The code is AI generated but I manually reviewed it and it looks OK to me. I just want this upstreamed so we can use CDNs on mobile clients - iOS requires special apple account, for example.


M1khail909:
closed as not planned)))))

michaelH4nter:
Thanks for the fix! I’ve been struggling with the same problem where CDNs don't respect X-Accel-Buffering: no, causing GET requests to hang.

Fangliding:
https://github.com/XTLS/Xray-core/pulls?q=is%3Apr+state%3Aclosed+XHTTP

cerg2010cerg2010:
> https://github.com/XTLS/Xray-core/pulls?q=is%3Apr+state%3Aclosed+XHTTP

Ok, I will rewrite the code. Thanks for the explanation!

PR #6786, opened 2026-09-19
Log: Skip disabled messages before preparing entries

JkLondon:
Skip disabled general log messages before caller lookup, Error allocation and session-prefix preparation. Use an optional `Enabled(Severity) bool` query with atomically published handler and logger-level snapshots; preserve nested severity, Handle-only handlers, startup/restart behavior and the final locked check in `Handle`. Access/DNS logging is unchanged.

**Draft:** enabled parallel logging still has a small cost. A focused 20-sample run found 249.6 → 252.1 ns/op (+1.02%, p<0.001); the final 10-sample run below found no significant difference (p=0.072). This does not establish zero overhead.

Go 1.27.0 (go.mod), darwin/arm64, Apple M3 Max, GOMAXPROCS=8. Identical benchmarks on upstream/main `dcdfc57c` and this patch; 10 runs per version, 1s per case, alternating order for each serial/parallel pair. Accepted messages are formatted without writer I/O. Benchstat medians:

| Benchmark | Before ns/op | After ns/op | B/op | allocs/op |
|---|---:|---:|---:|---:|
| Disabled Debug / serial | 320.85 | 11.06 | 408 → 16 | 7 → 1 |
| Disabled Debug / parallel | 193.90 | 5.072 | 408 → 16 | 7 → 1 |
| Disabled Info / serial | 320.15 | 10.84 | 408 → 16 | 7 → 1 |
| Disabled Info / parallel | 194.95 | 5.028 | 408 → 16 | 7 → 1 |
| None / serial | 316.90 | 10.85 | 408 → 16 | 7 → 1 |
| None / parallel | 194.65 | 5.011 | 408 → 16 | 7 → 1 |
| Enabled Debug / serial | 507.6 | 510.9 | 536 → 536 | 14 → 14 |
| Enabled Debug / parallel | 250.8 | 253.6 | 536 → 536 | 14 → 14 |

Disabled time improves 96.55–97.43% (p<0.001). Enabled serial time has no significant change (p=0.404). These are local microbenchmarks, not production throughput measurements.

Validation: focused tests and race ×10 for common/errors, common/log, app/log and app/log/command, focused vet, repository formatting, and compilation of all packages pass.

Reason: on xray server have noticed that even if logging was disabled, `common/errors.doLog` was accounted for ~500Mb or ~3% of allocated bytes. So, I was like: maybe we can check is the message enabled before?


Fangliding:
这样有点太麻烦了 我回去想想 而且这点不在热路径上也提升不了很多吞吐量

Meo597:
https://github.com/XTLS/Xray-core/pull/5947#issuecomment-4310367859
https://github.com/XTLS/Xray-core/pull/5947#issuecomment-4322139107

巧了我昨天趁着还有重置卡正在用 Astra ultra 猛蹬

PR #6789, opened 2026-09-20
xdrive: make storage object names configurable

kolya13245:
### Motivation

The `xdrive` transport lays out a fixed, self-describing directory tree in the
shared storage:

```
sessions/<unixnano>-<id>
streams/<id>/c2s/000000000.seg
streams/<id>/s2c/000000000.end
```

The directory and suffix strings (`sessions`, `streams`, `c2s`, `s2c`, `.seg`,
`.end`, `.err`) are hardcoded. For a transport whose whole point is to route
traffic through a third-party storage the client doesn't control, the owner of
that storage can read the layout at a glance and see exactly what it is. Being
able to rename these objects to something neutral is a cheap way to make the
layout less obviously a proxy.

### What this changes

Adds an optional `naming` message to the xdrive config. Every field defaults to
the current literal, so:

- an unconfigured peer is byte-for-byte compatible with the existing behavior;
- setting the fields lets both ends agree on their own object names.

Both peers must configure identical values.

```jsonc
"xdriveSettings": {
  "service": "Google Drive",
  "remoteFolder": "<folder id>",
  "secrets": ["...", "...", "..."],
  "naming": {
    "sessionsDir": "a", 
    "streamsDir": "b",
    "uplinkDir": "u", 
    "downlinkDir": "d",
    "segSuffix": ".1", 
    "endSuffix": ".2", 
    "errSuffix": ".3"
  }
}
```

### Implementation

- `config.proto`: new `Naming` message, `Config.naming` (field 13). Regenerated `config.pb.go`.
- The layout strings move into a small `names` struct, built by `namesFromConfig`
  with the upstream literals as defaults. The functions that build object paths
  (`announceName`, `*Prefix`, `parseEntry`) become methods on it.
- `infra/conf`: parse the `naming` object in `xdriveSettings`.

### Compatibility

Defaults are unchanged, so this is fully backward compatible: a peer without a
`naming` block interoperates with any existing xdrive peer. A peer *with*
overrides only interoperates with a peer configured the same way — this is
by design and documented as a requirement.

### Testing

- `go test ./transport/internet/xdrive/` — pass. New `TestNamesFromConfig`
  covers defaults, partial override (empty fields keep defaults), and that
  `parseEntry` only accepts the configured suffixes.
- `go test ./infra/conf/ -run XDrive` — pass.
- Manual: two local xray processes (`service: "local"`) with a custom `naming`
  block — the custom directory/suffix names appear in the storage and traffic
  flows (HTTP 200 through the SOCKS inbound).


Fangliding:
xhttp那堆混淆参数又来了
但是这里的固定特征更甚 服务商只要管理极其高频的文件写入读出删除操作就行了

RPRX:
目前的重点是先把核心功能做好，至于这类 obfs 参数，等有相关的封锁出现后再有针对性地慢慢加

